PR #192, opened 2014-08-07
All Cards in category page should be centered

ffarfan:
Changed the alignment so that all the cards in the category page appear centered, no matter the resolution.

[endlessm/eos-sdk#1838]


ffarfan:
Doing some cleanup.

PR #195, opened 2014-08-12
Add coordinates to ContentObjectModel

davidswinegar:
Allow marshalling of coordinates from JSON-LD on the ContentObjectModel.
[endlessm/eos-sdk#1949]


ptomato:
Well, I haven't looked at this one in a while!

We could add it to endlessm/libdmodel which is where this code lives now. But since the code is now in C there's nothing really salvageable from this PR :disappointed: 

Thanks, David and Will, sorry this never got merged!

PR #303, opened 2014-12-18
Removed debug_back_button binding to bookmark_issue.

goizueta:
The binding wasn't correctly handling rapid updates to
bookmark_issue.

Now, whenever bookmark_issue gets changed, the value of
the debug back button's sensitivity will update.

[endlessm/eos-sdk#2438]


ptomato:
This code doesn't exist anymore, unfortunately we never got around to fixing this up.

PR #438, opened 2015-04-17
Hide scrollbars during page transition

rmacqueen:
The method for hiding the scrollbars
on the current page seems to only work about 30%
of the time. Hiding the scrollbars for
the next page (when going backwards) works
all the time.

[endlessm/eos-sdk#3028]


ptomato:
I haven't looked at this one in a while. The code that it modifies is now defunct... sorry it never got merged!

PR #962, opened 2016-04-26
Knowledge.Class defines getters for style properties

mattdangerw:
This is a small bit of syntactic sugar that adds getters for each style
property. The getter wraps EosKnowledgePrivate.widget_style_get_X(),
where X is int or float.

https://phabricator.endlessm.com/T11442


ptomato:
We closed the associated issue as wontfix because we're hoping to move to GTK 4 at some point, where style properties are deprecated.

PR #1153, opened 2016-08-18
T12846

mattdangerw:
https://phabricator.endlessm.com/T12846

### DIFF
--- a/js/app/historyStore.js
+++ b/js/app/historyStore.js
@@ -42,6 +42,20 @@ const HistoryStore = new Lang.Class({
 
     Properties: {
         /**
+         * Property: current-set
+         *
+         * The model for the current set in the history store.
+         */
+        'current-set': GObject.ParamSpec.object('current-set', 'current-set', 'current-set',
+            GObject.ParamFlags.READABLE, SetObjectModel.SetObjectModel),
+        /**
+         * Property: current-query
+         *
+         * The model for the current query in the history store.
+         */
+        'current-query': GObject.ParamSpec.string('current-query', 'current-query', 'current-query',
+            GObject.ParamFlags.READABLE, ''),
+        /**
          * Property: animating
          * Whether a page in this app is animating.
          */
@@ -153,22 +167,45 @@ const HistoryStore = new Lang.Class({
         return item || null;
     },
 
+    get current_query () {
+        let item = this.search_backwards(0, (item) => item.query.length > 0);
+        if (item)
+            return item.query;
+        return '';
+    },
+
+    get current_set () {
+        let item = this.search_backwards(0, (item) => item.page_type === Pages.SET);
+        if (item)
+            return item.model;
+        return null;
+    },
+
+    _update_index: function (delta) {
+        let old_set = this.current_set;
+        let old_query = this.current_query;
+        this._index += delta;
+        this.emit('changed');
+        if (old_query !== this.current_query)
+            this.notify('current-query');
+        if (old_set !== this.current_set)
+            this.notify('current-set');
+    },
+
     // Common helper functions for history stores, not for use from other
     // modules...
     go_back: function () {
         if (!this.can_go_back())
             return;
-        this._index--;
         this._direction = Direction.BACKWARDS;
-        this.emit('changed');
+        this._update_index(-1);
     },
 
     go_forward: function () {
         if (!this.can_go_forward())
             return;
-        this._index++;
         this._direction = Direction.FORWARDS;
-        this.emit('changed');
+        this._update_index(1);
     },
 
     set_current_item: function (item) {
--- a/js/app/modules/selection/articleContext.js
+++ b/js/app/modules/selection/articleContext.js
@@ -3,6 +3,7 @@
 // Copyright 2016 Endless Mobile, Inc.
 
 const Module = imports.app.interfaces.module;
+const HistoryStore = imports.app.historyStore;
 const Pages = imports.app.pages;
 const QueryObject = imports.search.queryObject;
 const Xapian = imports.app.modules.selection.xapian;
@@ -13,42 +14,42 @@ const ArticleContext = new Module.Class({
 
     _init: function (props={}) {
         this.parent(props);
-        this._item = {};
+        this._item = null;
+        this._update_item();
+
+        HistoryStore.get_default().connect('changed', this._update_item.bind(this));
     },
 
-    // Selection.Xapian implementation
-    on_history_changed: function (history) {
-        let item = history.get_current_item();
-        if (!item)
+    _update_item: function () {
+        let item = HistoryStore.get_default().search_backwards(0, (item) => {
+            return item.page_type === Pages.SET || item.query.length > 0;
+        });
+        if (item === this._item)
             return;
-
-        if ((item.query && item.query !== this._item.query) ||
-            (item.page_type === Pages.SET && this._item.model != item.model)) {
-            this._item = item;
-            this._set_needs_refresh(true);
-        }
+        this._item = item;
+        this._set_needs_refresh(true);
     },
 
     construct_query_object: function (limit, query_index) {
         if (query_index > 0)
             return null;
 
-        if (this._item.query) {
+        if (this._item === null)
+            return null;
+
+        if (this._item.query.length > 0) {
             return new QueryObject.QueryObject({
                 limit: limit,
                 query: this._item.query,
                 tags_match_all: ['EknArticleObject'],
             });
-        } else if (this._item.page_type === Pages.SET) {
-            return new QueryObject.QueryObject({
-                limit: limit,
-                tags_match_all: ['EknArticleObject'],
-                tags_match_any: this._item.model.child_tags,
-                sort: QueryObject.QueryObjectSort.SEQUENCE_NUMBER,
-            });
-        } else {
-            return null;
         }
-        return null;
+
+        return new QueryObject.QueryObject({
+            limit: limit,
+            tags_match_all: ['EknArticleObject'],
+            tags_match_any: this._item.model.child_tags,
+            sort: QueryObject.QueryObjectSort.SEQUENCE_NUMBER,
+        });
     },
 });
--- a/js/app/modules/selection/articlesForSet.js
+++ b/js/app/modules/selection/articlesForSet.js
@@ -4,9 +4,9 @@
 
 const Actions = imports.app.actions;
 const Dispatcher = imports.app.dispatcher;
+const HistoryStore = imports.app.historyStore;
 const Module = imports.app.interfaces.module;
 const QueryObject = imports.search.queryObject;
-const SetObjectModel = imports.search.setObjectModel;
 const Xapian = imports.app.modules.selection.xapian;
 
 const ArticlesForSet = new Module.Class({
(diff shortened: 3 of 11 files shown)


mattdangerw:
I should probably test all the selections a bit more thoroughly, but curious what people think of the approach.

js/app/modules/selection/articleContext.js
```diff
@@ -13,37 +13,47 @@ const ArticleContext = new Module.Class({
...
+            if (history.current_query) {
+                this._use_query = true;
+                this._use_set = false;
+            } else {
+                this._use_query = false;
+            }
             this._set_needs_refresh(true);
-        }
+        });
+        history.connect('notify::current-set', () => {
```
rmacqueen: can you not combine both these callbacks into one? That is called both for `notify::current-set` and for `'notify::current-query`?

js/app/modules/selection/articlesForSet.js
```diff
@@ -15,31 +15,28 @@ const ArticlesForSet = new Module.Class({
...
 
     construct_query_object: function (limit, query_index) {
         if (query_index > 0)
             return null;
-        if (!this.model)
-            throw new Error('You should not be loading this selection unless on the set page');
+
+        let model = this.model;
+        if (this.global)
+            model = HistoryStore.get_default().current_set;
```
rmacqueen: I think this needs to be `this.model` so that it gets picked up by DynamicTitle

js/app/modules/selection/search.js
```diff
@@ -10,20 +11,19 @@ const Search = new Module.Class({
     Name: 'Selection.Search',
     Extends: Xapian.Xapian,
 
-    // Selection.Xapian implementation
-    on_history_changed: function (history) {
-        let item = history.get_current_item();
-        if (item.query) {
-            this._item = item;
+    _init: function (props) {
```
rmacqueen: `props={}`

js/app/modules/selection/search.js
```diff
@@ -10,20 +11,19 @@ const Search = new Module.Class({
...
+        HistoryStore.get_default().connect('notify::current-query', () => {
             this._set_needs_refresh(true);
-        }
+        });
     },
 
     construct_query_object: function (limit, query_index) {
-        if (query_index > 0)
+        let query = HistoryStore.get_default().current_query;
+        if (query_index > 0 || query.length === 0)
```
rmacqueen: do we want to be returning `null` if query is empty string? Does `engine` handle `null` queries nicely?

js/app/modules/selection/subsetsForSet.js
```diff
@@ -2,34 +2,37 @@
...
+                this._set_needs_refresh(true);
+            });
         }
     },
 
     construct_query_object: function (limit, query_index) {
         if (query_index > 0)
             return null;
 
+        let model = this.model;
```
rmacqueen: ditto re `this.model`

js/app/historyStore.js
```diff
@@ -153,22 +167,45 @@ const HistoryStore = new Lang.Class({
...
+    },
+
+    get current_set () {
+        let item = this.search_backwards(0, (item) => item.page_type === Pages.SET);
+        if (item)
+            return item.model;
+        return null;
+    },
+
+    update_index: function (delta) {
```
ptomato: Nitpick: I'd make this private

js/app/modules/selection/articleContext.js
ptomato: I would guess no, because they do slightly different things.

js/app/modules/selection/search.js
ptomato: It's handled in Selection.Xapian, although I believe the code currently requires at least the first call to `construct_query_object` to return a non-null query object; subsequent ones may be null.

js/app/modules/selection/search.js
```diff
@@ -11,7 +11,7 @@ const Search = new Module.Class({
     Name: 'Selection.Search',
     Extends: Xapian.Xapian,
 
-    _init: function (props) {
+    _init: function (props={}) {
```
ptomato: Nitpick: Seems like it should have been squashed into the previous commit

ptomato:
@rmacqueen I agree with your assessment re. `this.model`, though I think in that case we need to make sure that a `notify::model` is emitted when it changes via the global state. It seems that previously in the code, `this.global` and `this.model` were mutually exclusive, but as you correctly point out, DynamicTitle relies on `this.model`.

ptomato:
Never mind about `notify::model`, after all; `notify` has a setter which does that already.

tests/js/app/testHistoryStore.js
```diff
@@ -32,6 +33,40 @@ describe('History Store', function () {
         expect(history_store.get_items().length).toBe(1);
     });
 
+    it('tracks the current query', function () {
+        history_store.set_current_item_from_props({
+            page_type: 'search',
```
tchx84: nit: we could use Pages.SEARCH instead?

js/app/modules/selection/supplementaryArticles.js
```diff
@@ -12,22 +12,20 @@ const SupplementaryArticles = new Module.Class({
     Name: 'Selection.SupplementaryArticles',
     Extends: Xapian.Xapian,
 
-    // Selection.Xapian implementation
-    on_history_changed: function (history) {
-        if (!this.global)
-            return;
```
tchx84: Why is this check no longer necessary?

js/app/modules/selection/articleContext.js
```diff
@@ -13,37 +13,47 @@ const ArticleContext = new Module.Class({
...
-        } else if (this._item.page_type === Pages.SET) {
+        } else if (this._use_set) {
             return new QueryObject.QueryObject({
                 limit: limit,
                 tags_match_all: ['EknArticleObject'],
-                tags_match_any: this._item.model.child_tags,
+                tags_match_any: history.current_set.child_tags,
             });
         } else {
             return null;
```
tchx84: I know this is not part of the changes, but isn't this "return null" redundant?

tchx84:
It works! But I don't understand what is _the_ exact change that fixes the bug mentioned in T12846. I like how the API was polished so we don't have to repeat the same over and over, but in general the changes look kind of equivalent (?). @ptomato @rmacqueen @mattdangerw can you explain a bit more?

I know must be something so obvious, that I don't see it, but I think we should add at least an explanation in the corresponding commits (only to those that are related to the issue, not all the commits that just make use of the new API).

js/app/modules/selection/supplementaryArticles.js
ptomato: `this.global` was already always true in the case of Selection.SupplementaryArticles.

ptomato:
Good question, actually. I feel silly for not asking it myself. To my surprise, I can't easily answer it.

I've made the other changes you requested.

js/app/modules/selection/articlesForSetCrossSection.js
```diff
@@ -16,24 +16,18 @@ const ArticlesForSetCrossSection = new Module.Class({
...
         this.parent(props);
         this._set_needs_refresh(true);
-    },
-
-    // Selection.Xapian implementation
-    on_history_changed: function (history) {
-        let item = history.get_current_item();
-        if (item.model instanceof SetObjectModel.SetObjectModel &&
-            item.model !== this._current_set) {
-            this._current_set = item.model;
```
mattdangerw: @tchx84 the lines of code here are what were actually causing the breakage with the travel app filtering.

The problem was that the old `articlesForSetCrossSection` would only have the correct state after a history change, where it would pick up on the new `_current_set` and track it. But in the case where these selections are created dynamically (which happens in the travel app), there could be a "current set" in our history that predates the creation of this selection.

So basically
 - history changes to reflect current set--Asia
 - `articlesForSetCrossSection` is created, `this._current_set` is null, `this.model` is "Beautiful Beaches" 
 - query is run with null `_current_set`, so we get all beautiful beaches not just "beautiful beaches in asia"


 The fix for this is a bit subtle, but the one I liked the most. We remove any attempt to duplicate and track state from the history store in this module, and just write a getter directly on the history store that retrieves the state we need. That was no matter when this selection is created, `HistoryStore.get_default().current_set` will be the correct set.

This was a change I'd been eyeing for a while, so just did a big refactor of selections while doing that, but other selection changes are not specifically needed for this bug. This clears up a number of theoretical bugs that don't really show up in practice with out yaml files, such as a `search` selection being dynamically created after a search query was entered and not knowing the correct query to use.

mattdangerw:
I think the whole `this.model` for dynamic title module needs a little more thought. Moving back to dev.

tchx84:
@mattdangerw thanks for explaining, makes sense! btw, I recently added a filter which could also be theoretically affected for the same issue, think we should cover it here as well?

https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/modules/filter/currentSet.js#L21

mattdangerw:
@tchx84 yeah we should, I'll add that now.

mattdangerw:
@tchx84 actually that currentSet filter seems broken a little more broadly.

This filter seems to need some way of signalling that it needs to be rerun when the current_set changes. The current impl relies on the fact that the selection will call `set_needs_refresh` when the set changes, but that won't always be true.

If that sounds right to you, maybe we just fix that up on another issue?

tchx84:
@mattdangerw I was giving a final round of tests this morning and noticed something odd:

Using com.endlessm.animals.en:

1. launch the app.
2. click on Reptiles.
3. click on "Yacare Caiman" article.
4. click on "home" (top left button).
5. search for "bird".
6. click on "back" (top left button), all way to the "Yacare Caiman" article.

You will see a "Sorry, no results" error message instead of the list of articles. It behaves different on master (but it doesn't seems correct either).

mattdangerw:
> Using com.endlessm.astronomy.en:
> 
> launch the app.
> search of "as".
> click on "back" (top left button).

Actually this is happening on master too, pretty sure its related to the markup capitalization issue. Pretty close to tracking it down so I'll try to clean it up.

mattdangerw:
Opened https://phabricator.endlessm.com/T13356 for the second issue. First bug should be all cleared up

js/app/modules/selection/articleContext.js
```diff
@@ -13,41 +14,43 @@ const ArticleContext = new Module.Class({
...
+        this._set_needs_refresh(true);
     },
 
     construct_query_object: function (limit, query_index) {
         if (query_index > 0)
             return null;
 
-        if (this._item.query) {
+        if (this._item === null)
+            return;
```
tchx84: shouldn't this be "return null"?

tchx84: shouldn't this be "return null"?

tchx84:
Looks good, I no longer see the first regression I mentioned (and the second one is being worked in a separate issue). There is only one small comment.

tchx84:
@mattdangerw I noticed something else, but can't really do much science here. Can you try this on T12846 branch and master to compare:

1. switch to e T12846 branch (or master) and make install;
2. launch travel app.
3. on home- page scrolling down to bottom and from "Browse by Topic" select "Central America" or another topic with many articles.
4. once in the all-sets-page just search for "de".
5. once in the search-page, use the back navigation button to go all the way back to the home-page and then forward button to go all way to the search page again.

Now, repeat this back and forward thing (step 5) around 5-10 times. I do notice a big difference on the e T12846 branch:

1. on each repetition of the step 5, it seems to take longer to switch pages.
2. around the second or third time I the application freezes. You get the "force quit" message.

While on Master I don't see this happening.

mattdangerw:
@tchx84 good catch! This appears to me to be a separate bug though, all old articlesForSetCrossSection selection are just hanging around permanently and as you navigate to more and more set pages you will have more and more selections trying to update themselves when the current set changes.

This happens for me on master too just by continually clicking around new set categories. E.g. from home -> Castles and Places -> Asia -> Tourist Attractions -> Europe -> Beautiful Beaches. Things will just slowly grind to a halt. I'll open up a separate issue.

tchx84:
Works great ! **CR PASS**

PR #1154, opened 2016-08-18
Use the actual variable!

ffarfan:
We were using the wrong variable when producing the chunk data in server-
templated news articles.

This now rightfully shows the backlink to the featured set in article pages.

https://phabricator.endlessm.com/T12631


magcius:
Ugh, this is my fault. Looks good, CR pass from me.

I don't know if we need to produce a new runtime for this -- it's a relatively minor bug.

PR #1155, opened 2016-08-24
T12959

ebassi:
https://phabricator.endlessm.com/T12959


xjuan:
The only issue I found is that horizontal thumbs wont fit as much text as it could

virtualbox_endless dev_29_08_2016_19_29_59

ptomato:
I was wondering this morning if we could keep the two lines and instead get rid of the smaller tags label if there was not enough room; perhaps using SpaceContainer for that purpose.

ptomato:
Discussed during standup that we won't do so until we have constraint layouts :smile: **CR pass**

PR #1156, opened 2016-08-24
T12825

rmacqueen:
https://phabricator.endlessm.com/T12825

### DIFF
--- a/js/app/modules/layout/articleStack.js
+++ b/js/app/modules/layout/articleStack.js
@@ -81,6 +81,10 @@ const ArticleStack = new Module.Class({
         },
     },
 
+    References: {
+        'selection': {},  // type: Selection
+    },
+
     CONTENT_TRANSITION_DURATION: 500,
 
     _init: function (props={}) {
@@ -94,6 +98,17 @@ const ArticleStack = new Module.Class({
 
         HistoryStore.get_default().connect('changed',
             this._on_history_changed.bind(this));
+
+        this.reference_module('selection', selection => {
+            this._selection = selection;
+            if (this._selection) {
+                this._selection.connect('models-changed', () => {
+                    let models = this._selection.get_models();
+                    if (models.length > 0)
+                        this._load_article_model(models[0]);
+                });
+            }
+        });
 
         // Clear old views from the stack when its not animating.
         this.connect('notify::transition-running', () => {
--- a/js/app/modules/order/alphabetical.js
+++ b/js/app/modules/order/alphabetical.js
@@ -16,7 +16,25 @@ const Alphabetical = new Module.Class({
     Extends: GObject.Object,
     Implements: [Order.Order],
 
+    Properties: {
+        'case-sensitive': GObject.ParamSpec.boolean('case-sensitive', 'Case sensitive',
+            'Whether the sorting should be case sensitive',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT, false),
+    },
+
     compare_impl: function (left, right) {
-        return left.title.localeCompare(right.title);
+        let a = left.title;
+        let b = right.title;
+
+        // We always want uppercase letters to come after lowercase ones
+        // which won't happen if we just leave it to localeCompare
+        let first_char_a = a.charAt(0);
+        let first_char_b = b.charAt(0);
+        let is_upper_a = first_char_a == first_char_a.toLocaleUpperCase();
+        let is_upper_b = first_char_b == first_char_b.toLocaleUpperCase();
+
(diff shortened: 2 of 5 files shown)


ptomato:
re. Search box optional: it's already supposedly optional in the constructor, but badly implemented. (`git blame` blames me.) It would be nice to refactor it a bit so that the search stack and invisible frame are also not added if there is no module in the search box slot.

js/app/modules/layout/articleStack.js
```diff
@@ -79,6 +79,11 @@ const ArticleStack = new Module.Class({
         'nav-card': {
             multi: true,
         },
+        'selection': {},
```
ptomato: Seems like the slot is never used, only the reference.

ptomato:
re. TiledGrid: I'm OK with that, and consider it an improvement, even though it's probably not something that we would merge into an stable version in the future.

js/app/modules/arrangement/tiledGrid.js
```diff
@@ -28,8 +28,8 @@ const TiledGrid = new Module.Class({
         this.get_children().forEach(this.remove, this);
         // The card to be packed is already in this array:
         let cards = this.get_cards();
-
-        let columns = Math.ceil(this.get_card_count() / 2);
+        let count = this.get_card_count();
+        let columns = count == 2 ? 2 : Math.ceil(this.get_card_count() / 2);
```
ptomato: While you're at it, I would probably refactor this so that 3 cards make 3 columns too.

ptomato:
I cherry picked the article loading and the dollar sign commits to master, so you can rebase on top of those for a smaller review next round.

rmacqueen:
Added a commit to make alphabetical sort put uppercase words last, so that `Anexo` shows up at the end

js/app/modules/layout/articleStack.js
```diff
@@ -81,6 +81,10 @@ const ArticleStack = new Module.Class({
         },
     },
 
+    References: {
+        'selection': {},  // type: Selection
```
ptomato: On further consideration, I think this reference should be made optional, since none of the other presets use it (yet).

js/app/modules/layout/articleStack.js
```diff
@@ -95,6 +99,15 @@ const ArticleStack = new Module.Class({
         HistoryStore.get_default().connect('changed',
             this._on_history_changed.bind(this));
 
+        this.reference_module('selection', selection => {
+            this._selection = selection;
+            this._selection.connect('models-changed', () => {
```
ptomato: ...which means you'd have to have a null check before this line.

tests/js/app/modules/layout/testArticleStack.js
```diff
@@ -31,6 +31,9 @@ describe('Layout.ArticleStack', function () {
         let factory;
         [module, factory] = new MockFactory.setup_tree({
             type: ArticleStack.ArticleStack,
+            references: {
+                'selection': 'selection',
+            },
```
ptomato: ...and it would be good to add a test that ensures the module still works without the reference.

ptomato:
I'm OK with not doing the search box refactor I suggested above, for now, but if you choose not to then please open an issue for it in the next milestone.

js/app/modules/order/alphabetical.js
```diff
@@ -17,6 +17,20 @@ const Alphabetical = new Module.Class({
     Implements: [Order.Order],
 
     compare_impl: function (left, right) {
-        return left.title.localeCompare(right.title);
+        let a = left.title;
+        let b = right.title;
+        if (!a || !b)
```
ptomato: This should be unnecessary since titles are mandatory.

js/app/modules/order/alphabetical.js
```diff
@@ -17,6 +17,20 @@ const Alphabetical = new Module.Class({
...
+            return a > b ? -1 : 1;
+
+        // We always want uppercase letters to come after lowercase ones
+        // which wont get capture by localeCompare
```
ptomato: typos: won't get captured
(and it's not clear what you mean by "captured")

ptomato:
Now having seen the changes to Order.Alphabetical, I wonder if it's not better to make a separate module Order.Collation that just simply does `a > b ? -1 : 1` all the time.

ptomato:
Picked the TiledGrid commit to master, so you can rebase again for a smaller review.

rmacqueen:
ready again

ptomato:
**CR pass**, don't forget to open the issue to refactor the optional search box in 3.1

PR #1157, opened 2016-08-25
T12831

rmacqueen:
https://phabricator.endlessm.com/T12831

### DIFF
--- a/js/app/modules/contentGroup/contentGroup.js
+++ b/js/app/modules/contentGroup/contentGroup.js
@@ -166,9 +166,14 @@ const ContentGroup = new Module.Class({
 
     _on_models_changed: function () {
         let models = this._selection.get_models();
-        if (models.length === 0 && this._no_results) {
+        this.visible = true;
+        if (models.length === 0) {
             this._arrangement.set_models([]);
-            this._stack.visible_child_name = NO_RESULTS_PAGE_NAME;
+            if (this._no_results) {
+                this._stack.visible_child_name = NO_RESULTS_PAGE_NAME;
+            } else {
+                this.visible = false;
+            }
         } else {
             let max_cards = this._arrangement.get_max_cards();
             if (max_cards > -1)
--- a/tests/js/app/modules/contentGroup/testContentGroup.js
+++ b/tests/js/app/modules/contentGroup/testContentGroup.js
@@ -35,7 +35,6 @@ describe('ContentGroup.ContentGroup', function () {
                 'selection': {
                     type: Minimal.MinimalSelection,
                 },
-                'no-results': { type: null },
             },
         });
         arrangement = factory.get_last_created('arrangement');
@@ -60,6 +59,7 @@ describe('ContentGroup.ContentGroup', function () {
     it('adds cards created by selection to the arrangement', function () {
         selection.queue_load_more(5);
         expect(arrangement.get_count()).toBe(5);
+        expect(content_group.visible).toBe(true);
         expect(factory.get_created('arrangement.card').length).toBe(5);
     });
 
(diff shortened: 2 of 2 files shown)


ptomato:
Are the Arrangement.Windshield modifications necessary? The app doesn't seem to use Windshield.

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -166,9 +166,14 @@ const ContentGroup = new Module.Class({
 
     _on_models_changed: function () {
         let models = this._selection.get_models();
-        if (models.length === 0 && this._no_results) {
+        this.visible = true;
+        if (models.length === 0) {
             this._arrangement.set_models([]);
-            this._stack.visible_child_name = NO_RESULTS_PAGE_NAME;
+            if (this._no_results) {
```
ptomato: Not sure I agree with this; shouldn't app writers take care that content groups have a no-results submodule if there is a chance that they can be empty?

js/app/modules/selection/allSets.js
```diff
@@ -23,6 +23,7 @@ const AllSets = new Module.Class({
         return new QueryObject.QueryObject({
             limit: limit,
             tags_match_all: ['EknSetObject'],
+            sort: QueryObject.QueryObjectSort.SEQUENCE_NUMBER,
```
ptomato: Oh nice, we're missing this in a couple other apps too.

ptomato:
The app banner orientation commit should be unnecessary with the re-exported logos.

js/app/utils.js
```diff
@@ -365,10 +365,10 @@ function unsquash_all_window_content_updates_heavy_handedly (window) {
 
 function alignment_to_justification (align) {
     switch (align) {
-        case Gtk.Align.FILL:
```
ptomato: This will break other uses of `alignment_to_justification`, such as aligning Banner.App's subtitle. That definitely should be centered when Banner.App's halign is set to FILL.

ptomato:
Picked some commits to master, so you can rebase on that and have a smaller review for next round.

rmacqueen:
> Are the Arrangement.Windshield modifications necessary? The app doesn't seem to use Windshield.

Yea, not anymore it doesn't. I still think those commits are a net positive improvement, but since we're on a tight schedule, I'd say it's safer to leave them out for now.

js/app/modules/contentGroup/contentGroup.js
rmacqueen: I don't think so. Sometimes you just want to completely hide the content group when it has nothing to show, as in the case for this app.

ptomato: Maybe I'll understand with more context; which content group is this for, and when is it supposed to show nothing? I'm having trouble imagining a use case for this that isn't bad UX.

rmacqueen:
ready again

ptomato:
I still need a good use case for the no-results commit. Picked the other one to master.

js/app/modules/contentGroup/contentGroup.js
rmacqueen: The set page shows two content groups, one for subsets of the current set, and one (below it) for articles of the current set. Some sets _only_ have subsets (and no articles) or _only_ have articles (and no subsets). In those cases you want the content group with no content in it to just hide

ptomato: OK, I got confused with the mention of search results. I understand now.

We were already doing this in news.yaml with the set page; how come it worked properly there?

ptomato:
Could you add a test to make sure the content group hides itself when there are no results? Otherwise looks good.

rmacqueen:
Added test

ptomato:
**CR pass**

PR #1158, opened 2016-08-26
Use custom uri scheme for license links

tchx84:
This is just a workaround for issue we discovered
with webkit and file uri schemes. So, we just use
a custom uri scheme for these links now.

https://phabricator.endlessm.com/T12960


tchx84:
bah, noticed it is used in other places as well... will update the commit, but makes this approach less attractive.

xjuan:
replaced by https://github.com/endlessm/eos-knowledge-lib/pull/1173

PR #1159, opened 2016-08-26
domain.js: downloader.fetch_update_finish only needs task

tchx84:
We were passing the wrong params to fetch_update_finish which
at the very least prevented us from seeing errors in requests.

https://phabricator.endlessm.com/T13140


ptomato:
**CR pass**

PR #1160, opened 2016-08-30
T13152

tchx84:
https://phabricator.endlessm.com/T13152

### DIFF
--- a/data/preset/news.yaml
+++ b/data/preset/news.yaml
@@ -155,10 +155,7 @@ defines:
                               slots:
                                 order: 'Order.PublishedDate(ascending: false)'
                                 filter:
-                                  type: Filter.OtherGroup
-                                  references:
-                                    other: set-page-subsets
-
+                                  type: Filter.CurrentSet
                     - type: Layout.ResponsiveMargins
                       slots:
                         content:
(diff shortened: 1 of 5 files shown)


data/preset/news.yaml
```diff
@@ -155,10 +155,7 @@ defines:
                               slots:
                                 order: 'Order.PublishedDate(ascending: false)'
                                 filter:
-                                  type: Filter.OtherGroup
```
rmacqueen: `Filter.OtherGroup` isn't used at all anymore, right? So we can just kill it

rmacqueen:
Just a comment that I think you can now remove some dead code. Other than that looks good

rmacqueen:
CR PASS

ptomato:
We definitely wouldn't remove Filter.OtherGroup on the eos3.0 branch, that would be an API break, and it would break apps if people hadn't updated recently. We probably should not remove any modules until we have a good story for releasing new eos-knowledge runtime versions.

Maybe we could add a comment to Filter.OtherGroup saying it's being considered for removal, to remind ourselves to review it next time we have an API break? Deprecated would be a bit strong, since if we found a good use for it in the meantime we would still keep it.

PR #1164, opened 2016-09-05
T13191

tchx84:
(not for merging yet, just need some feedback)

https://phabricator.endlessm.com/T13191


ptomato:
**CR pass**

The "not for merging" referred to our discussion on Slack about the correct behaviour for Layout.DynamicBackground. Summary: this is exactly as I thought it should work, so I agree.

PR #1165, opened 2016-09-05
T12580

tchx84:
https://phabricator.endlessm.com/T12580


tchx84:
For reviewer, I double checked with @joanafilizola and is OK to let the search banner "go away" with the scrolling.

mattdangerw:
This looks like two separate changes, one for changes background size, and another for changing scroll behavior.

I'd make two separate commits.

Otherwise looks good

mattdangerw:
CR pass

PR #1167, opened 2016-09-06
Retrieve the return value from Pango.parse_markup

rmacqueen:
Previously we were calling Pango.parse_markup but
not actually looking at the return value to get
the parsed text.

https://phabricator.endlessm.com/T13220

### DIFF
--- a/js/app/widgets/formattableLabel.js
+++ b/js/app/widgets/formattableLabel.js
@@ -92,7 +92,9 @@ const FormattableLabel = new Knowledge.Class({
         if (!text)
             text = '';
         if (this.use_markup) {
-            Pango.parse_markup(text, -1, '0', null, text);
+            // Pango.parse_markup returns an array where the third
+            // element is the result of the parsing.
+            text = Pango.parse_markup(text, -1, '0', null, text)[2];
         }
 
         let formatted_label = this._format_capitals(text, transform);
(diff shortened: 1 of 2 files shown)


js/app/widgets/formattableLabel.js
```diff
@@ -92,7 +92,9 @@ const FormattableLabel = new Knowledge.Class({
         if (!text)
             text = '';
         if (this.use_markup) {
-            Pango.parse_markup(text, -1, '0', null, text);
+            // Pango.parse_markup returns an array where the third
+            // element is the result of the parsing.
+            text = Pango.parse_markup(text, -1, '0', null, text)[2];
```
mattdangerw: Should we handle broken markup somehow? first return value of false I think? This will just hard error.

mattdangerw:
Talked, think this will behave reasonably (readable error printed out) in the markup fail state.

CR pass

mattdangerw:
Talked, think this will behave reasonably (readable error printed out) in the markup fail state.

CR pass

mattdangerw:
Talked, think this will behave reasonably (readable error printed out) in the markup fail state.

CR pass

mattdangerw:
Talked, think this will behave reasonably (readable error printed out) in the markup fail state.

CR pass

PR #1168, opened 2016-09-06
Set page and search page banners should be consistent

rmacqueen:
As per design review. The base css file for Cards sets
a different pixel size than that of the window. As a
consequence, we cant use ems because while set banner
uses a Card.Title, search banner uses no card, so the
base pixel sizes would be different.

https://phabricator.endlessm.com/T13202


mattdangerw:
CR pass

PR #1169, opened 2016-09-06
Legacy polaroid should not change color when highlighted

rmacqueen:
Since apps with these cards were made before the notion
of highlights existed.

https://phabricator.endlessm.com/T13201

### DIFF
--- a/data/css/modules/card/_legacyPolaroid.scss
+++ b/data/css/modules/card/_legacyPolaroid.scss
@@ -15,6 +15,10 @@
         }
     }
 
+     &.highlighted &__title {
+        color: $accent-dark-color;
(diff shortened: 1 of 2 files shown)


mattdangerw:
Think this is a good change for now. In the future I think this needs a little more thought. Don't we want all cards to support highlight if any of the cards support highlight? So we can mix and match them on any preset with the same user experience.

A more "modular" way to do this might be to continue to have some sort of highlight in the theme for legacy polaroid cards, and somehow not activate the highlight on any cards on the template A home page, be that explicitly through the yaml or implicitly somehow.

Anyway this PR improves things, so CR pass

PR #1172, opened 2016-09-07
Fix search string highlighting

ptomato:
This regressed on a previous commit because we were using
Pango.parse_markup() to get text to capitalize, which stripped all the
<span>s from the string.

This gets the Pango attribute list from Pango.parse_markup() and
reapplies it, as well as correcting a bug with the accel character. It
also replaces the existing test with a more robust one that looks at the
Pango.Layout directly instead of relying on formatting to be set using
Pango markup.

Original fix by @mattdangerw and @rmacqueen.

https://phabricator.endlessm.com/T13220

### DIFF
--- a/js/app/widgets/formattableLabel.js
+++ b/js/app/widgets/formattableLabel.js
@@ -91,13 +91,15 @@ const FormattableLabel = new Knowledge.Class({
     _format_label: function (text, transform) {
         if (!text)
             text = '';
+        let attrs = null;
         if (this.use_markup) {
             // Pango.parse_markup returns an array where the third
             // element is the result of the parsing.
-            text = Pango.parse_markup(text, -1, '0', null, text)[2];
+            [, attrs, text] = Pango.parse_markup(text, -1, '\0', null);
         }
 
         let formatted_label = this._format_capitals(text, transform);
         this.set_label(formatted_label);
+        this.set_attributes(attrs);
     },
 });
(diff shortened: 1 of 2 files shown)


tchx84:
I noticed something else when testing this with com.endlessm.travel.en, right after we launch the app, I see the following errors:

```
(com.endlessm.travel.en:6406): Gtk-WARNING **: Failed to set text 'Castles & Palaces' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;

(com.endlessm.travel.en:6406): Gtk-WARNING **: Failed to set text 'Castles & Palaces' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;

(com.endlessm.travel.en:6406): Gtk-WARNING **: Failed to set text 'Religious Sites & Places of Worship' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;

(com.endlessm.travel.en:6406): Gtk-WARNING **: Failed to set text 'Religious Sites & Places of Worship' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity - escape ampersand as &amp;
```

And then, these cards are missing their titles in the all sets page:

screenshot from 2016-09-07 15-28-31

tchx84:
**CR PASS**, I will move this to Dev or create a new issue for that report, since it also happens in Master.

PR #1174, opened 2016-09-07
T13271

tchx84:
https://phabricator.endlessm.com/T13271

### DIFF
--- a/tests/js/app/widgets/testFormattableLabel.js
+++ b/tests/js/app/widgets/testFormattableLabel.js
@@ -53,6 +53,6 @@ describe('Formattable Label', function () {
         Utils.update_gui();
         label.use_markup = true;
         label.label = "Custer&apos;s Last Stand";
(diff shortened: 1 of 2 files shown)


tests/js/app/widgets/testFormattableLabel.js
```diff
@@ -53,6 +53,6 @@ describe('Formattable Label', function () {
         Utils.update_gui();
         label.use_markup = true;
         label.label = "Custer&apos;s Last Stand";
-        expect(label.label).toBe("CUSTER'S LAST STAND");
```
tchx84: The only question I have is how this worked before? When were we converting these?

ptomato:
**CR pass**

ptomato:
We discussed on Slack that it might be simpler to turn off `use_markup`, but that would be confusing for an API consumer, to have that property adjust itself depending on CSS. Instead, we really need to do this formatting upstream in Pango.

PR #1175, opened 2016-09-08
T13271 (fixup)

tchx84:
https://phabricator.endlessm.com/T13271

### DIFF
--- a/tests/js/app/widgets/testFormattableLabel.js
+++ b/tests/js/app/widgets/testFormattableLabel.js
@@ -1,4 +1,5 @@
 const Gtk = imports.gi.Gtk;
+const GLib = imports.gi.GLib;
 
 const FormattableLabel = imports.app.widgets.formattableLabel;
 const Utils = imports.tests.utils;
@@ -54,5 +55,30 @@ describe('Formattable Label', function () {
         label.use_markup = true;
         label.label = "Custer&apos;s Last Stand";
         expect(label.label).toBe("CUSTER&apos;S LAST STAND");
+    });
+
+    it('handles ampersand symbols correctly', function () {
+        let cssProvider = new Gtk.CssProvider();
+
+        cssProvider.load_from_data ("* { \
+            -EknFormattableLabel-text-transform: 'uppercase';\
+        }");
+
+        let win = new Gtk.OffscreenWindow();
+        win.add(label);
+        win.show_all();
+        label.get_style_context().add_provider(cssProvider,
+                                               Gtk.StyleProvider.PRIORITY_APPLICATION + 10);
+
+        Utils.update_gui();
+        let text = "a & b";
+
+        label.use_markup = true;
+        label.label = GLib.markup_escape_text(text, -1);
+        expect(label.label).toBe("A &amp; B");
+
+        label.use_markup = false;
+        label.label = text;
+        expect(label.label).toBe("A & B");
     });
 });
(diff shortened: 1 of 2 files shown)


tchx84:
The previous commit fixed the issue with the missing titles, but, introduced yet another regressions:

a) non-markup text gets escaped and shows incorrectly. (this is fixed by the previous commit)
b) markup text which does get escaped, keep changing so the label it set kind-of-recursively (I am still trying to figure out how to fix this).

tchx84:
I am adding an extra test to make sure this widget is robust enough against these confusing ampersand symbols...

tests/js/app/widgets/testFormattableLabel.js
```diff
@@ -55,4 +56,29 @@ describe('Formattable Label', function () {
...
+        win.add(label);
+        win.show_all();
+        label.get_style_context().add_provider(cssProvider,
+                                                Gtk.StyleProvider.PRIORITY_APPLICATION + 10);
+
+        Utils.update_gui();
+        let text = "a & b";
+
+        label.use_markup = true;
+        label.label = GLib.markup_escape_text(text, -1);
```
mattdangerw: Why do you need to call `GLib.markup_escape_text(text, -1);` here? Won't the fact that you have `use_markup = true` plus that custom label setter that calls  `_format_label` mean this gets called already with your changes above?

tests/js/app/widgets/testFormattableLabel.js
```diff
@@ -55,4 +56,29 @@ describe('Formattable Label', function () {
...
+
+        cssProvider.load_from_data ("* { \
+            -EknFormattableLabel-text-transform: 'uppercase';\
+        }");
+
+        let win = new Gtk.OffscreenWindow();
+        win.add(label);
+        win.show_all();
+        label.get_style_context().add_provider(cssProvider,
+                                                Gtk.StyleProvider.PRIORITY_APPLICATION + 10);
```
mattdangerw: nit: remove extra space

tests/js/app/widgets/testFormattableLabel.js
tchx84: The main reason is that ```label.label = "a & b"```; is not valid input when use_markup is true (this is also how the base GtkLabel behaves), "&" is treated a special character for escaping symbols e.g. ```"&amp;"```, so text must be escaped. Please keep this in mind for what I explain next.

I know is confusing let me try explain:

1) Inside "_format_label", we use "parse_markup" and this "reverts" the escaped symbols, so

e.g. ```"a &amp; b"``` turns back into ```"a & b"```;
If we pass, a plain "&" the parser will reject it (as I explained, ```&`` is not valid input with use_markup true).

2) then, we do the transformations.

e.g. ```"a & b"``` turns into ```"A & B"```

3) then we set the text back to widget, using this.set_label, BUT we need to escape these symbols back or it won't work:

e.g. if use_markup is true, then we will try to ```set_label("A & B")```, which (as I explained before) is not valid, must be escaped back so we do ```set_label("A &amp; B")```.

Let me know If this explanation is not clear. But the point is, we do need to escape text before setting it to the label (when use_markup is true).

The logic is completely different if use_markup is false, that is why need all these ```if (this.use_markup)``` checks.

tchx84:
@mattdangerw fixed the nit!

ptomato:
Found some nitpicks, but due to the deadline I am fixing them myself. **CR pass**

PR #1176, opened 2016-09-08
Pause dispatcher while query is loading

ptomato:
This is a quick fix for the problem where the history store changes, then
changes again while a selection is still making its request.

https://phabricator.endlessm.com/T13268


ptomato:
This is not the right fix for the problem.

PR #1178, opened 2016-09-09
Align the Card.Title modules to start

rmacqueen:
So that the space to the right of them
is not clickable.

https://phabricator.endlessm.com/T13198

### DIFF
--- a/data/preset/encyclopedia.yaml
+++ b/data/preset/encyclopedia.yaml
@@ -46,9 +46,10 @@ defines:
                   - shortdef: ContentGroup.ContentGroup
                     slots:
                       arrangement:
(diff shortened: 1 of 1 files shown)


data/preset/encyclopedia.yaml
```diff
@@ -48,7 +48,8 @@ defines:
                       arrangement:
                         shortdef: 'Arrangement.List(hexpand: true)'
```
xjuan: I think we should also set the arrangement list homogeneous property to false, otherwise all the results will have the  width of the widest one

rmacqueen: Nice catch!

xjuan:
CR pass

PR #1180, opened 2016-09-13
T12815

tchx84:
https://phabricator.endlessm.com/T12815

### DIFF
--- a/data/css/modules/navigation/_searchBox.scss
+++ b/data/css/modules/navigation/_searchBox.scss
@@ -1,4 +1,11 @@
 .NavigationSearchBox {
+    .WindowSimple & {
+        selection {
+            color: black;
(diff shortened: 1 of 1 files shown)


data/css/modules/navigation/_searchBox.scss
```diff
@@ -1,4 +1,11 @@
 .NavigationSearchBox {
+    .WindowSimple & {
```
mattdangerw: The theme for top bar search search bar actually lives in eos-sdk IIRC. See the `.endless-search-box` theming there.

We want any app using a top bar search box to get this fix, so I think this should land in eos-sdk instead.

tchx84:
You are absolutely right, and it's even better, because it won't require to rebuild apps.

PR #1181, opened 2016-09-14
T13171

tchx84:
https://phabricator.endlessm.com/T13171


js/app/modules/contentGroup/contentGroup.js
```diff
@@ -13,6 +13,7 @@ const Overflow = imports.app.modules.arrangement.overflow;
 const Utils = imports.app.utils;
 
 const BATCH_SIZE = 10;
+const EXTRA_CLIP = 40;
```
ebassi: This is a bit of a magic number; it would be best to leave a comment saying why it's 40px.

tchx84:
@ebassi I modified the value and added a comment to explain where the value comes from, BUT to be honest, since this clip thing is new to me, I am not entirely sure what kind of side effects this approach could have.

For example,  not all the GtkStack children will actually need that extra clip space. What could happen in that case? Is there a better way of calculating that extra space we need?

tchx84:
@ebassi no more magic numbers in this version.

PR #1182, opened 2016-09-14
T13356

mattdangerw:
https://phabricator.endlessm.com/T13356

### DIFF
--- a/js/app/interfaces/card.js
+++ b/js/app/interfaces/card.js
@@ -199,7 +199,6 @@ const Card = new Lang.Interface({
         if (this._highlight_string === value)
             return;
         this._highlight_string = value;
-        this.update_highlight_string();
         this.notify('highlight-string');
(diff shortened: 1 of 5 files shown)


tchx84:
@mattdangerw I noticed that 794ccc2 fixes the issue I reported with home-page cards's getting highlighted, but I don't fully understand how to reproduce what is being fixed in 3d4d13c, can you clarify?

mattdangerw:
@tchx84 see above. I was seeing the highlight string problem with only https://github.com/endlessm/eos-knowledge-lib/pull/1182/commits/794ccc2a91d03151c4ec305cd7ef382a62a85d68 applied. Both necessary.

tchx84:
As you mentioned on slack, animations makes the difference, once I enabled them the second patch become necessary.

**CR PASS**

PR #1183, opened 2016-09-14
lib/eosknowledgeprivate: Added new EknWebViewDialog widget. 

xjuan:
This widget is mainly intended to display file:// uris

https://phabricator.endlessm.com/T13273

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,8 +44,6 @@ distdir:
 	@false
 endif
 
-# # # GRESOURCES # # #
-
 scss_verbose = $(scss_verbose_@AM_V@)
 scss_verbose_ = $(scss_verbose_@AM_DEFAULT_V@)
 scss_verbose_0 = @echo "  SCSS     $@";
@@ -169,6 +167,7 @@ CLEANFILES += $(scss_targets)
 clean-local::
 	rm -rf .sass-cache
 
+# # # GRESOURCES # # #
 resource_deps_verbose = $(resource_deps_verbose_@AM_V@)
 resource_deps_verbose_ = $(resource_deps_verbose_@AM_DEFAULT_V@)
 resource_deps_verbose_0 = @echo "  RESDEPS  $@";
@@ -190,6 +189,11 @@ resource_verbose_0 = @echo "  RES      $@";
 	$(GLIB_COMPILE_RESOURCES) --target=$@ \
 		--sourcedir=$(srcdir)/$(*D) --sourcedir=$(builddir)/$(*D) $<
 
+%.c: %.gresource.xml
+	$(resource_verbose) $(MKDIR_P) $(dir $@) && \
+	$(GLIB_COMPILE_RESOURCES) --target=$@ --generate-source \
+		--sourcedir=$(srcdir)/$(*D) --sourcedir=$(builddir)/$(*D) $<
+
 gresources =
 gresource_depedencies =
 -include $(gresources:.gresource=.gresource.deps)
@@ -207,6 +211,18 @@ gresources += data/moltres/moltres.gresource
 pkgdata_DATA += $(gresources)
 
 # # # MAIN LIBRARY # # #
+BUILT_SOURCES = \
+	lib/eosknowledgeprivate/ekn-resources.c \
+	$(NULL)
+
+lib/eosknowledgeprivate/ekn-resources.c: $(shell $(GLIB_COMPILE_RESOURCES) --generate-dependencies --sourcedir=$(srcdir)/lib/eosknowledgeprivate/ $(srcdir)/lib/eosknowledgeprivate/ekn-resources.gresource.xml)
+
+CLEANFILES += $(BUILT_SOURCES)
+
+EXTRA_DIST += \
+	lib/eosknowledgeprivate/ekn-resources.gresource.xml \
+	lib/eosknowledgeprivate/EknRuntimeDocumentViewer.ui \
+	$(NULL)
 
 eosknowledgeprivate_headers = \
 	lib/eosknowledgeprivate/ekn-util.h \
@@ -215,6 +231,9 @@ eosknowledgeprivate_headers = \
 eosknowledgeprivate_sources = \
 	lib/eosknowledgeprivate/ekn-init.c lib/eosknowledgeprivate/ekn-init-private.h \
 	lib/eosknowledgeprivate/ekn-util.c \
+	lib/eosknowledgeprivate/ekn-resources.c \
+	lib/eosknowledgeprivate/ekn-runtime-document-viewer.h \
+	lib/eosknowledgeprivate/ekn-runtime-document-viewer.c \
 	$(NULL)
 
 # Endless Knowledge Apps GUI library
--- a/js/app/widgets/eknWebview.js
+++ b/js/app/widgets/eknWebview.js
@@ -12,6 +12,7 @@ const ArticleObjectModel = imports.search.articleObjectModel;
 const AsyncTask = imports.search.asyncTask;
 const Config = imports.app.config;
 const Engine = imports.search.engine;
+const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
 const Knowledge = imports.app.knowledge;
 const SearchUtils = imports.search.utils;
 const Utils = imports.app.utils;
@@ -95,6 +96,13 @@ const EknWebview = new Knowledge.Class({
             GObject.signal_stop_emission_by_name(this, 'query-tooltip');
             return false;
         });
+        this.connect('unmap', () => {
+            // Destroy License viewer when the view gets hidden
+            if (this._license_view) {
+                this._license_view.destroy();
+                delete this._license_view;
+            }
+        });
         gtk_settings.connect('notify::gtk-xft-dpi', this._updateFontSizeFromGtkSettings.bind(this));
     },
 
@@ -174,11 +182,19 @@ const EknWebview = new Knowledge.Class({
             let scheme = GLib.uri_parse_scheme(uri);
             if (scheme !== null && this.EXTERNALLY_HANDLED_SCHEMES.indexOf(scheme) !== -1) {
                 if (scheme === 'license') {
-                    let license = GLib.uri_unescape_string(uri.replace('license://', ''), null);
-                    uri = Endless.get_license_file(license).get_uri();
-                }
+                    // Create a license viewer
+                    if (!this._license_view) {
+                        this._license_view = new EosKnowledgePrivate.RuntimeDocumentViewer({
+                            transient_for: this.get_toplevel(),
+                        });
+                        this._license_view.maximize();
+                    }
 
-                Gtk.show_uri(null, uri, Gdk.CURRENT_TIME);
+                    let license = GLib.uri_unescape_string(uri.replace('license://', ''), null);
+                    this._license_view.index_uri = Endless.get_license_file(license).get_uri(); 
+                } else {
+                    Gtk.show_uri(null, uri, Gdk.CURRENT_TIME);
+                }
                 decision.ignore();
                 return true; // handled
             }
--- a/lib/eosknowledgeprivate/ekn-runtime-document-viewer.c
+++ b/lib/eosknowledgeprivate/ekn-runtime-document-viewer.c
@@ -1,0 +1,348 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*- */
+/*
+ * ekn-runtime-document-viewer.c
+ *
+ * Copyright (C) 2016 Endless Mobile, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Author: Juan Pablo Ugarte <ugarte@endlessm.com>
+ *
+ */
+
+#include "ekn-runtime-document-viewer.h"
+
+typedef struct
+{
+  gchar *index_uri;
+  gchar *index_scheme;
+  gboolean show_on_load;
+
+  WebKitWebView *webview;
+  GtkHeaderBar  *headerbar;
+  GtkWidget     *back;
+  GtkWidget     *forward;
+} EknRuntimeDocumentViewerPrivate;
+
+struct _EknRuntimeDocumentViewer
+{
+  GtkWindow parent_instance;
+};
+
+enum
+{
+  PROP_0,
+
+  PROP_INDEX_URI,
+  PROP_SHOW_ON_LOAD,
+  N_PROPERTIES
+};
+
+static GParamSpec *properties[N_PROPERTIES];
+
+G_DEFINE_TYPE_WITH_PRIVATE (EknRuntimeDocumentViewer,
+                            ekn_runtime_document_viewer,
+                            GTK_TYPE_WINDOW);
+
+#define ERDV_PRIVATE(d) ((EknRuntimeDocumentViewerPrivate *) ekn_runtime_document_viewer_get_instance_private(d))
+
+static void
+on_back_forward_list_changed (WebKitBackForwardList     *list,
+                              WebKitBackForwardListItem *items_added,
+                              GList                     *items_removed,
+                              EknRuntimeDocumentViewer  *dialog)
+{
+  EknRuntimeDocumentViewerPrivate *priv = ERDV_PRIVATE (dialog);
+
+  gtk_widget_set_sensitive (priv->back, webkit_web_view_can_go_back (priv->webview));
+  gtk_widget_set_sensitive (priv->forward, webkit_web_view_can_go_forward (priv->webview));
+}
+
+static void
+ekn_runtime_document_viewer_init (EknRuntimeDocumentViewer *dialog)
+{
+  EknRuntimeDocumentViewerPrivate *priv = ERDV_PRIVATE (dialog);
+
+  priv->show_on_load = TRUE;
+
+  gtk_widget_init_template (GTK_WIDGET (dialog));
+
+  g_object_bind_property (priv->webview, "title", priv->headerbar, "title", 0);
+
+  g_signal_connect (webkit_web_view_get_back_forward_list (priv->webview),
+                    "changed",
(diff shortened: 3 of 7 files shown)


Makefile.am
```diff
@@ -207,14 +216,30 @@ gresources += data/moltres/moltres.gresource
...
 	lib/eosknowledgeprivate/ekn-util.h \
+	lib/eosknowledgeprivate/ekn-resources.h \
+	lib/eosknowledgeprivate/ekn-web-view-dialog.h \
 	$(NULL)
 
 eosknowledgeprivate_sources = \
 	lib/eosknowledgeprivate/ekn-init.c lib/eosknowledgeprivate/ekn-init-private.h \
 	lib/eosknowledgeprivate/ekn-util.c \
+	lib/eosknowledgeprivate/ekn-resources.c \
+	lib/eosknowledgeprivate/ekn-web-view-dialog.c \
```
xjuan: Perhaps we should put this in a new shared object eosknowledgeui ?

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,31 @@ const EknWebview = new Knowledge.Class({
                 if (scheme === 'license') {
                     let license = GLib.uri_unescape_string(uri.replace('license://', ''), null);
                     uri = Endless.get_license_file(license).get_uri();
+
+                    let license_view = new EosKnowledgePrivate.WebViewDialog({
+                      'transient-for' : this.get_toplevel(),
```
xjuan: Do we want the license window to be transient?

Makefile.am
```diff
@@ -207,14 +216,30 @@ gresources += data/moltres/moltres.gresource
...
 
 # # # MAIN LIBRARY # # #
+BUILT_SOURCES = \
+	lib/eosknowledgeprivate/ekn-resources.h \
+	lib/eosknowledgeprivate/ekn-resources.c \
+	$(NULL)
+
+ekn_resources_deps = $(shell $(GLIB_COMPILE_RESOURCES) --generate-dependencies --sourcedir=lib/eosknowledgeprivate/ lib/eosknowledgeprivate/ekn-resources.gresource.xml)
+lib/eosknowledgeprivate/ekn-resources.h: $(ekn_resources_deps)
+lib/eosknowledgeprivate/ekn-resources.c: $(ekn_resources_deps)
```
cosimoc: Why not just make the generic `%.c` and `%.h` rules depend on `%.gresource.deps` instead?

Makefile.am
```diff
@@ -207,14 +216,30 @@ gresources += data/moltres/moltres.gresource
...
+
+CLEANFILES += $(BUILT_SOURCES)
+
+EXTRA_DIST += lib/eosknowledgeprivate/ekn-resources.gresource.xml
 
 eosknowledgeprivate_headers = \
 	lib/eosknowledgeprivate/ekn-util.h \
+	lib/eosknowledgeprivate/ekn-resources.h \
```
cosimoc: I am not sure this needs to go in here (as opposed to the sources list below). I don't think we'll be directly calling into functions of this file from introspection.

Makefile.am
cosimoc: Nah, I think this is OK

Makefile.am
```diff
@@ -231,6 +256,7 @@ libeosknowledgeprivate_la_CPPFLAGS = \
 	@EOS_C_COVERAGE_CFLAGS@ \
 	-DG_LOG_DOMAIN=\"EosKnowledge\" \
 	-DCOMPILING_EOS_KNOWLEDGE \
+	-DPKGDATADIR=\""$(pkgdatadir)"\" \
```
cosimoc: This does not look like it's used anywhere

js/app/widgets/eknWebview.js
```diff
@@ -15,6 +15,7 @@ const Engine = imports.search.engine;
 const Knowledge = imports.app.knowledge;
 const SearchUtils = imports.search.utils;
 const Utils = imports.app.utils;
+const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
```
cosimoc: Alphabetical order

js/app/widgets/eknWebview.js
cosimoc: Yes

cosimoc: Please remove quotes and use underscores for these parameters (e.g. `transient_for`)

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,31 @@ const EknWebview = new Knowledge.Class({
                 if (scheme === 'license') {
                     let license = GLib.uri_unescape_string(uri.replace('license://', ''), null);
                     uri = Endless.get_license_file(license).get_uri();
+
+                    let license_view = new EosKnowledgePrivate.WebViewDialog({
+                      'transient-for' : this.get_toplevel(),
+                      'width-request' : 640,
+                      'height-request': 720,
```
cosimoc: Not sure these make a lot of sense, since we want to be compatible with composite screens. You can set the default width and height though.

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,31 @@ const EknWebview = new Knowledge.Class({
...
+                      'home-uri'      : uri,
+                    });
+
+                    license_view.webview.connect('decide-policy', (view, decision, type) => {
+                      let uri = decision.request.uri;
+
+                      if (!GLib.str_has_prefix (uri, 'file')) {
```
cosimoc: Prefer an early return

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,31 @@ const EknWebview = new Knowledge.Class({
...
+
+                    let license_view = new EosKnowledgePrivate.WebViewDialog({
+                      'transient-for' : this.get_toplevel(),
+                      'width-request' : 640,
+                      'height-request': 720,
+                      'home-uri'      : uri,
+                    });
+
+                    license_view.webview.connect('decide-policy', (view, decision, type) => {
+                      let uri = decision.request.uri;
```
cosimoc: You could use `GLib.uri_parse_scheme` to decide instead of simply looking at the prefix.

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,31 @@ const EknWebview = new Knowledge.Class({
...
+
+                      if (!GLib.str_has_prefix (uri, 'file')) {
+                        Gtk.show_uri(null, uri, Gdk.CURRENT_TIME);
+                        decision.ignore();
+                        return true;
+                      }
+                      return false;
+                    });
+
+                    license_view.present ();
```
cosimoc: Remove space before paren

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,31 @@ const EknWebview = new Knowledge.Class({
...
+                }
+                else {
```
cosimoc: No newline here

lib/eosknowledgeprivate/ekn-web-view-dialog.c
```diff
@@ -0,0 +1,256 @@
...
+static void
+ekn_web_view_dialog_init (EknWebViewDialog *dialog)
+{
+  EknWebViewDialogPrivate *priv;
+
+  dialog->priv = priv = ekn_web_view_dialog_get_instance_private (dialog);
+
+  gtk_widget_init_template (GTK_WIDGET (dialog));
+  
+  g_signal_connect (priv->webview, "notify::title",
```
cosimoc: You could use a GBinding here instead

lib/eosknowledgeprivate/ekn-web-view-dialog.c
```diff
@@ -0,0 +1,256 @@
...
+
+  priv = dialog->priv;
+
+  if (g_strcmp0 (priv->home_uri, uri))
+    {
+      g_free (priv->home_uri);
+      g_free (priv->home_uri_scheme);
+
+      priv->home_uri = g_strdup (uri);
+      priv->home_uri_scheme = g_uri_parse_scheme (uri);
```
cosimoc: This is never used and could be removed. (It is also leaked.)

lib/eosknowledgeprivate/ekn-web-view-dialog.h
```diff
@@ -0,0 +1,70 @@
...
+  GtkWindow parent_instance;
+
+  EknWebViewDialogPrivate *priv;
+};
+
+struct EknWebViewDialogClass
+{
+  GtkWindowClass parent_class;
+
+  void   (* reserved1)   (void);
```
cosimoc: We typically don't use padding, especially since this is private.

Makefile.am
xjuan: Oops, leftovers from when I needed it to load the resources manually

lib/eosknowledgeprivate/ekn-web-view-dialog.h
xjuan: Old habit :)

lib/eosknowledgeprivate/ekn-web-view-dialog.c
xjuan: ohh good catch, thats some leftover from when I called gtk_show_uri() from this object.
Then I decided that behavior was to specific to be in the widget itself and moved that logic to JS

lib/eosknowledgeprivate/ekn-web-view-dialog.c
xjuan: Right, that reminds me I need to add support for specifying bindings in glade! :S

Makefile.am
xjuan: Because %.gresource.deps generates dependencies for .gresource target and I need the deps for .c and .h files

js/app/widgets/eknWebview.js
xjuan: I tried default-width/height but its not respected, I guess the webview is requesting for more.

For example CC is responsive and it fits well in less than 640, but it does not fit so well in between  640 and 840

So I am not sure if there is an easy way to know which size the window should be.
Fo now I think the best think would be to open the view fullscreen!

xjuan:
Philips and I decided to do it in C because he mention that eventually he wants to create a new UI lib and migrate some JS widget to C so that other apps can use it.

@ptomato was there any other reason?

Makefile.am
ptomato: Are the dependencies not the same for both targets?

Makefile.am
```diff
@@ -207,14 +216,30 @@ gresources += data/moltres/moltres.gresource
 pkgdata_DATA += $(gresources)
 
 # # # MAIN LIBRARY # # #
+BUILT_SOURCES = \
+	lib/eosknowledgeprivate/ekn-resources.h \
+	lib/eosknowledgeprivate/ekn-resources.c \
+	$(NULL)
+
+ekn_resources_deps = $(shell $(GLIB_COMPILE_RESOURCES) --generate-dependencies --sourcedir=lib/eosknowledgeprivate/ lib/eosknowledgeprivate/ekn-resources.gresource.xml)
```
ptomato: Use `:=` otherwise the command will be re-run every time the variable is referenced.

Makefile.am
ptomato: Agreed, this should probably go into `sources` below

lib/eosknowledgeprivate/ekn-web-view-dialog.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+#ifndef _EKN_WEB_VIEW_DIALOG_H_
+#define _EKN_WEB_VIEW_DIALOG_H_
+
+#include <webkit2/webkit2.h>
+
+G_BEGIN_DECLS
+
+#define EKN_TYPE_WEB_VIEW_DIALOG             (ekn_web_view_dialog_get_type ())
```
ptomato: Use one of the `G_DECLARE_` macros instead.

lib/eosknowledgeprivate/ekn-web-view-dialog.c
```diff
@@ -0,0 +1,239 @@
...
+  gtk_widget_set_sensitive (priv->back, webkit_web_view_can_go_back (priv->webview));
+  gtk_widget_set_sensitive (priv->forward, webkit_web_view_can_go_forward (priv->webview));
+}
+
+static void
+ekn_web_view_dialog_init (EknWebViewDialog *dialog)
+{
+  EknWebViewDialogPrivate *priv;
+
+  dialog->priv = priv = ekn_web_view_dialog_get_instance_private (dialog);
```
ptomato: We tend not to do this in other code, just use `...get_instance_private()` everywhere since that is not really more than a pointer access either.

lib/eosknowledgeprivate/ekn-web-view-dialog.c
```diff
@@ -0,0 +1,239 @@
...
+
+  g_return_if_fail (EKN_IS_WEB_VIEW_DIALOG (dialog));
+
+  priv = dialog->priv;
+
+  if (g_strcmp0 (priv->home_uri, uri))
+    {
+      g_free (priv->home_uri);
+      priv->home_uri = g_strdup (uri);
+      webkit_web_view_load_uri (priv->webview, uri);
```
ptomato: For completeness, needs a `g_object_notify()` here

lib/eosknowledgeprivate/ekn-web-view-dialog.h
ptomato: Still should be removed.

js/app/widgets/eknWebview.js
```diff
@@ -12,6 +12,7 @@ const ArticleObjectModel = imports.search.articleObjectModel;
 const AsyncTask = imports.search.asyncTask;
```
ptomato: Commit message: lincese -> license typo

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,34 @@ const EknWebview = new Knowledge.Class({
                 if (scheme === 'license') {
                     let license = GLib.uri_unescape_string(uri.replace('license://', ''), null);
                     uri = Endless.get_license_file(license).get_uri();
-                }
 
-                Gtk.show_uri(null, uri, Gdk.CURRENT_TIME);
+                    let license_view = new EosKnowledgePrivate.WebViewDialog({
+                        transient_for : this.get_toplevel(),
```
ptomato: Nitpick: we generally don't do any vertical alignment like this in JS.

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,34 @@ const EknWebview = new Knowledge.Class({
...
+                        home_uri      : uri,
+                    });
+
+                    // Show local files only in the license window.
+                    // Open non file:// schemes in the system browser.
+                    license_view.webview.connect('decide-policy', (view, decision, type) => {
+                        if (type !== WebKit2.PolicyDecisionType.NAVIGATION_ACTION)
+                            return false;
```
ptomato: Nitpick: I always find the false/true return values from this handler confusing, so I like to put comments such as `// No decision made, continue with default action` or `// Decision was made`. Up to you, certainly optional.

js/app/widgets/eknWebview.js
```diff
@@ -176,9 +177,34 @@ const EknWebview = new Knowledge.Class({
...
 
-                Gtk.show_uri(null, uri, Gdk.CURRENT_TIME);
+                    let license_view = new EosKnowledgePrivate.WebViewDialog({
+                        transient_for : this.get_toplevel(),
+                        home_uri      : uri,
+                    });
+
+                    // Show local files only in the license window.
+                    // Open non file:// schemes in the system browser.
+                    license_view.webview.connect('decide-policy', (view, decision, type) => {
```
ptomato: Perhaps it would be better to include this behaviour in the license window widget itself? That way you wouldn't even have to expose the `webview` property.

data/widgets/tooltips.ui
```diff
@@ -47,7 +47,7 @@
       <object class="GtkLabel" id="license-label">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="label" translatable="yes">View the license in your browser</property>
+        <property name="label" translatable="yes">View the license in a new window</property>
```
ptomato: I'd say simply "View the license"

In any case, this one I think we'll have to skip cherry picking to eos3.0 unless we order another round of translations for 3.0.5.

xjuan: OK I will make a different PR for this, so its easier to cherry-pick

xjuan: https://github.com/endlessm/eos-knowledge-lib/pull/1191

js/app/widgets/eknWebview.js
ptomato: Still need to change the commit message on this commit

js/app/widgets/eknWebview.js
```diff
@@ -174,11 +182,20 @@ const EknWebview = new Knowledge.Class({
             let scheme = GLib.uri_parse_scheme(uri);
             if (scheme !== null && this.EXTERNALLY_HANDLED_SCHEMES.indexOf(scheme) !== -1) {
                 if (scheme === 'license') {
+                    // Create a license viewer
+                    if (this._license_view === undefined) {
```
ptomato: I would say `if (!this._license_view)`, comparing things to `undefined` can be problematic in JS

js/app/widgets/eknWebview.js
```diff
@@ -174,11 +182,20 @@ const EknWebview = new Knowledge.Class({
...
+                            transient_for: this.get_toplevel(),
+                        });
+                        this._license_view.maximize();
+                    }
+
                     let license = GLib.uri_unescape_string(uri.replace('license://', ''), null);
-                    uri = Endless.get_license_file(license).get_uri();
+                    this._license_view.index_uri = Endless.get_license_file(license).get_uri(); 
+                }
+                else {
```
ptomato: Nitpick: `} else {` on one line

lib/eosknowledgeprivate/ekn-runtime-document-viewer.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#ifndef _EKN_RUNTIME_DOCUMENT_VIEWER_H_
+#define _EKN_RUNTIME_DOCUMENT_VIEWER_H_
+
+#include <webkit2/webkit2.h>
+
+G_BEGIN_DECLS
+
+#define EKN_TYPE_WEB_VIEW_DIALOG (ekn_runtime_document_viewer_get_type ())
```
ptomato: Should be `EKN_TYPE_RUNTIME_DOCUMENT_VIEWER`

lib/eosknowledgeprivate/ekn-runtime-document-viewer.c
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * ekn_runtime_document_viewer_get_webview:
+ * @dialog: a #EknRuntimeDocumentViewer
+ *
+ * Returns the Webkit web view used to display the URI
+ *
+ * Returns: (transfer none): the webview is owned by the dialog.
+ */
+WebKitWebView *
+ekn_runtime_document_viewer_get_webview (EknRuntimeDocumentViewer *dialog)
```
ptomato: Sorry for forgetting to mention this before. Can you remove this property, I think it's not needed anymore?

ptomato:
**CR pass**

PR #1184, opened 2016-09-14
T13261

rmacqueen:
https://phabricator.endlessm.com/T13261


rmacqueen:
We are. In the soy502 overrides.scss I need to add a `-EknFormattableLabel-text-transform: 'lowercase'`

mattdangerw:
CR pass

PR #1187, opened 2016-09-20
Knowledge lib: Added a small sanity check for getting the app datadir

sgnn7:
Old error message was not intuitive and did not describe the actual
fault so now we include a better one.

### DIFF
--- a/js/search/domain.js
+++ b/js/search/domain.js
@@ -515,6 +515,12 @@ const DomainV3 = new Lang.Class({
    no EKN_VERSION file is found. This function does synchronous file I/O. */
 function get_ekn_version (app_id) {
     let dir = datadir.get_data_dir(app_id);
+
+    // Sanity check
+    if (!dir) {
+        throw new Error(Format.vprintf('Could not find data dir for app ID %s', [app_id]));
+    }
+
     let ekn_version_file = dir.get_child('EKN_VERSION');
     try {
         let [success, contents, _] = ekn_version_file.load_contents(null);
--- a/js/search/domain.js
+++ b/js/search/domain.js
@@ -522,13 +522,10 @@ function get_ekn_version (app_id) {
     }
 
     let ekn_version_file = dir.get_child('EKN_VERSION');
-    try {
-        let [success, contents, _] = ekn_version_file.load_contents(null);
-        let version_string = contents.toString();
-        return parseInt(version_string);
-    } catch (e) {
-        return 1;
-    }
+    let [success, contents, _] = ekn_version_file.load_contents(null);
+    let version_string = contents.toString();
+
+    return parseInt(version_string);
 }
 
 function get_domain_impl (app_id, xapian_bridge) {
--- a/tests/js/search/testDomain.js
+++ b/tests/js/search/testDomain.js
@@ -2,6 +2,7 @@ const Gio = imports.gi.Gio;
 
 const ArticleObjectModel = imports.search.articleObjectModel;
 const ContentObjectModel = imports.search.contentObjectModel;
+const Datadir = imports.search.datadir;
 const Domain = imports.search.domain;
 const Engine = imports.search.engine;
 const QueryObject = imports.search.queryObject;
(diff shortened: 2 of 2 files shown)


js/search/domain.js
```diff
@@ -515,6 +515,12 @@ const DomainV3 = new Lang.Class({
    no EKN_VERSION file is found. This function does synchronous file I/O. */
 function get_ekn_version (app_id) {
     let dir = datadir.get_data_dir(app_id);
+
+    // Sanity check
+    if (!dir) {
+        throw new Error(Format.vprintf('Could not find data dir for app ID %s', [app_id]));
```
magcius: I would remove the `return 1;` below since that's also bogus and shouldn't happen.

tests/js/search/testDomain.js
```diff
@@ -118,6 +119,20 @@ function create_mock_shard_with_link_table (link_table_hash) {
     return mock_shard_file;
 }
 
+describe('Domain', function () {
+    describe('get_ekn_version', function () {
+        it('should throw an exception when datadir can\'t be found', function (done) {
```
ptomato: This doesn't need to be async, so you can remove the `done`

tests/js/search/testDomain.js
```diff
@@ -118,6 +119,20 @@ function create_mock_shard_with_link_table (link_table_hash) {
...
+            let engine = new Engine.Engine();
+            spyOn(Datadir, 'get_data_dir').and.callFake(() => undefined);
```
ptomato: Use .and.returnValue(undefined) (or I think that's even what happens by default if you just spyOn the function)

js/search/domain.js
sgnn7: Hmm... I don't want to add unforseen side-effects. @ptomato - opinion on this?

magcius: The `return 1;` below will actually just cause an error. It will get looked up in an `EKN_VERSIONS` map which doesn't have a key for 1, and try to construct `undefined`, which will also throw an error.

sgnn7: k - i'll fix that up

sgnn7:
CR ready again

@ptomato: PS: It seems like I can't use sync version of `expect` with `toThrow` so I left the `done()` in there.

tests/js/search/testDomain.js
```diff
@@ -118,6 +119,20 @@ function create_mock_shard_with_link_table (link_table_hash) {
     return mock_shard_file;
 }
 
+describe('Domain', function () {
+    describe('get_ekn_version', function () {
+        it('should throw an exception when datadir can\'t be found', function (done) {
+            let engine = new Engine.Engine();
```
ptomato: Sorry I didn't notice this the first time; `engine` isn't used, so can be removed.

tests/js/search/testDomain.js
ptomato: I don't quite understand the comment you made about this; I was talking about the sync version of `it()`. As far as I know there is no async version of `expect()`; even if you do `expect(func).toThrow(err)` then `func` is executed immediately. I removed the `done` parameter from `it` and removed line 131 and it worked fine for me.

sgnn7:
@ptomato For some reason I missed that the `done` cb was passed into the `it()` itself but should be fixed now. CR ready again (sorry for the bouncebacks but I've been living away from JS for a while now).

magcius:
Yeah, Jasmine changing behavior based on arity is certainly confusing. I've ran into that one before as well.

js/search/domain.js
```diff
@@ -522,13 +522,10 @@ function get_ekn_version (app_id) {
...
-        let [success, contents, _] = ekn_version_file.load_contents(null);
-        let version_string = contents.toString();
-        return parseInt(version_string);
-    } catch (e) {
-        return 1;
-    }
+    let [success, contents, _] = ekn_version_file.load_contents(null);
+    let version_string = contents.toString();
+
+    return parseInt(version_string);
```
magcius: Perhaps add a sanity check error for the load_contents as well? Up to you.

PR #1189, opened 2016-09-20
Layout.Overlay

ptomato:
Introduces a Layout module that works pretty much like Gtk.Overlay.

https://phabricator.endlessm.com/T12440


rmacqueen:
CR PASS

PR #1192, opened 2016-09-21
Various fixes

ptomato:
https://phabricator.endlessm.com/T13467

### DIFF
--- a/js/app/interfaces/card.js
+++ b/js/app/interfaces/card.js
@@ -117,6 +117,9 @@ const MaxSize = {
  *
  * Requires:
  *   Gtk.Widget
+ *
+ * CSS classes:
+ *   Card--pdf, <class>--pdf - Added when the card is displaying a PDF record
  */
 const Card = new Lang.Interface({
     Name: 'Card',
@@ -171,6 +174,16 @@ const Card = new Lang.Interface({
         'sequence': GObject.ParamSpec.uint('sequence', 'Sequence', 'Sequence',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             0, Object.keys(Sequence).length, Sequence.NONE),
+    },
+
+    _interface_init: function () {
+        // null check on this.model: we don't really support null but it is
+        // often convenient for testing.
+        if (this.model && this.model.content_type === 'application/pdf') {
+            let context = this.get_style_context();
+            context.add_class(Utils.get_modifier_style_class('Card', 'pdf'));
+            context.add_class(Utils.get_modifier_style_class(this.constructor, 'pdf'));
+        }
     },
 
     set css (v) {
--- a/data/css/modules/banner/_set.scss
+++ b/data/css/modules/banner/_set.scss
@@ -5,7 +5,7 @@
         color: $primary-medium-color;
         font-weight: bold;
         font-size: 54px;
-        font-family: $display-font;
+        font-family: $title-font;
         padding-top: 50px;
     }
 }
--- a/data/css/modules/card/_thumb.scss
+++ b/data/css/modules/card/_thumb.scss
@@ -27,4 +27,11 @@
     .composite &__context {
         color: gray;
     }
+
+    &--pdf &__thumbnail {
+        background-image: url('resource:///com/endlessm/knowledge/data/images/pdf_icon.png');
+        background-repeat: no-repeat;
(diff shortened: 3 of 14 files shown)


js/app/modules/card/thumb.js
```diff
@@ -113,6 +113,9 @@ const Thumb = new Module.Class({
     _init: function (props={}) {
         this.parent(props);
 
+        if (this.model.content_type === 'application/pdf')
```
mattdangerw: Might be nicer to do this generally, some way of adding style classes based on content type for all cards not just thumb.

js/app/modules/card/thumb.js
```diff
@@ -113,6 +113,9 @@ const Thumb = new Module.Class({
...
+            this.get_style_context().add_class('pdf');
```
mattdangerw: But if we leaving it specific to thumb, I think the BEMier form would be `.CardThumb--pdf` or something like that

js/app/interfaces/card.js
```diff
@@ -302,18 +299,18 @@ const Card = new Lang.Interface({
...
-        frame.add(coveredFrame);
-
-        let cancellable = null;
-        let file = Gio.File.new_for_uri(this.model.thumbnail_uri);
-        let stream = file.read(cancellable);
-        coveredFrame.set_content(stream);
-        frame.visible = true;
+        let bg_provider = new Gtk.CssProvider();
+        bg_provider.load_from_data('frame {\
+            background-size: cover;\
```
mattdangerw: This might not actually be more performant, remember holding off on this change a while ago for that reason. Gtk won't attempt to cache anything, it will redraw (and scale) the background every time. Which for a big background image getting sized down could be slow. Template B is a good example.

However if this is going to land in 3.1 along with some gsk goodies. Suspect this is a non issue.

ptomato: Good point, I should probably leave this commit off and open a separate issue for it, to be merged after we get GSK.

js/app/modules/card/thumb.js
ptomato: OK, I expanded the changeset a bit.

js/app/interfaces/card.js
```diff
@@ -236,6 +239,24 @@ const Card = new Lang.Interface({
...
+     * Add style classes to the card depending on model properties
+     *
+     * This adds some CSS classes to cards based on model properties such as
+     * content type, that card implementations may want to use to customize the
+     * way they style themselves.
+     *
+     * Every card implementation's constructor should call this method after
+     * setting the construct properties.
+     */
+    add_style_classes_from_model: function () {
```
mattdangerw: We have interface inits now right? Any reason not to put this there? Less to mess up when writing a new card.

mattdangerw:
Sorry meant merge ready not CR ready, not sure how to edit that comment

js/app/interfaces/card.js
ptomato: The interface init runs before the construct properties are set, so there is no model at that point. i'd be open to changing the order around but I suspect there are just as many cases where we'd want the existing behaviour.

mattdangerw: I think you got the order flipped.

This comes after the  chain to parent.
https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/knowledge.js#L90

and this in card.js
```
    _interface_init: function () {
        printerr(this.model);
    },
```
seems to print out models

js/app/interfaces/card.js
```diff
@@ -236,6 +239,24 @@ const Card = new Lang.Interface({
...
+        if (this.model.content_type === 'application/pdf') {
+            let clsname = Utils.get_modifier_style_class('Card', 'pdf');
```
mattdangerw: Ah you could also add another.
```Utils.get_modifier_style_class(this.constructor, 'pdf');```
This would give you a version of the modifier class specific to the card (its what Fernando did for the title and similar classes).

Then above in the themeing you could just `&--pdf` instead of `&.Card--pdf`

js/app/interfaces/card.js
ptomato: Huh, maybe I got something else wrong when I tried it — I initially tried to do this with `_interface_init`. I'll take another look

ptomato:
All good suggestions; ready again.

mattdangerw:
CR pass

PR #1193, opened 2016-09-22
Adjust Card.List font sizes

rmacqueen:
To match designs. Previously, the synopsis
text was too large relative to the title
text.

https://phabricator.endlessm.com/T13216

### DIFF
--- a/data/css/modules/card/_list.scss
+++ b/data/css/modules/card/_list.scss
@@ -38,29 +38,17 @@
         color: $primary-medium-color;
     }
 
-    &.width-d &__title,
-    &.width-d &__synopsis {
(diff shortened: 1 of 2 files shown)


tchx84:
Hey @rmacqueen, I think there is something else causing this, please try this:

1. build prensa libre on top of the master.
2. launch it with gtk inspector.
3. go to "internacional" section.
4. and inspect a EknCard_List card.
5. go the nodes view and check the classes assigned to the card.

You will see that the card has 2 different width classes at the same time... that might explain what is reported in the issue. My screen solution is 1366x768.

tchx84:
** CR PASS **

PR #1197, opened 2016-09-26
T13437

tchx84:
https://phabricator.endlessm.com/T13437

### DIFF
--- a/js/app/widgets/dynamicLogo.js
+++ b/js/app/widgets/dynamicLogo.js
@@ -1,0 +1,308 @@
+// Copyright 2016 Endless Mobile, Inc.
+
+/* exported DynamicLogo */
+
+const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
+const Gio = imports.gi.Gio;
+const Gtk = imports.gi.Gtk;
+const GObject = imports.gi.GObject;
+const Rsvg = imports.gi.Rsvg;
+
+const Knowledge = imports.app.knowledge;
+
+const HEIGHT_WIDTH_RATIO = 1.5;
+
+const Width = {
+    MIN: 50,
+    MAX: 400,
+};
+
+const Height = {
+    MIN: 50,
+    MAX: 400,
+};
+
+/**
+ * Class: DynamicLogo
+ *
+ * This widget implements our dynamic logo system. A logo can be composed
+ * of an image, text or both. This widget is considered dynamic because it
+ * can be resized dynamically depending on its allocation size and it can
+ * switch between different layouts depending on the geometry of its
+ * allocation.
+ */
+const DynamicLogo = new Knowledge.Class({
+    Name: 'DynamicLogo',
+    Extends: Gtk.Widget,
+
+    Properties: {
+         /**
+         * Property: image-uri
+         *
+         * A URI to the image displayed in the logo.
+         */
+        'image-uri': GObject.ParamSpec.string('image-uri', 'image-uri', 'image-uri',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY, ''),
+         /**
+         * Property: text
+         *
+         * The text displayed in the logo.
+         */
+        'text': GObject.ParamSpec.string('text', 'text', 'text',
+            GObject.ParamFlags.READWRITE, ''),
+         /**
+         * Property: mode
+         *
+         * The mode of the logo, it can be text, image or full.
+         */
+        'mode': GObject.ParamSpec.string('mode', 'mode', 'mode',
+            GObject.ParamFlags.READWRITE, ''),
+    },
+
+    StyleProperties: {
+        /**
+         * Property: sizing
+         *
+         * -EknDynamicLogo-sizing:
+         *  * 'auto'     - size determined by allocated size. Default.
+         *  * 'size-min' - size determined by min-width and min-height css properties.
+         */
+        'sizing': GObject.ParamSpec.string('sizing', 'sizing', 'sizing',
+            GObject.ParamFlags.READWRITE, 'auto'),
+        /**
+         * Property: text-transform
+         *
+         * -EknDynamicLogo-text-transform:
+         *  * 'none'      - no transformation is performed. Default.
+         *  * 'uppercase' - the text is transformed to uppercase.
+         *  * 'lowercase' - the text is transformed to lowercase.
+         */
+        'text-transform': GObject.ParamSpec.string('text-transform', 'text-transform', 'text-transform',
+            GObject.ParamFlags.READWRITE, 'none'),
+    },
+
+    _init: function (props={}) {
+        this.parent(props);
+        this.set_has_window(false);
+        this._mode = 'text';
+        this._text = '';
+        this._sizing = 'auto';
+        this._text_transform = 'none';
+        this._layout = null;
+        this._image = null;
+
+        try {
+            let file = Gio.File.new_for_uri(this.image_uri);
+            let stream = file.read(null);
+            this._image = Rsvg.Handle.new_from_stream_sync(stream, file, 0, null);
+        } catch (e) {
+            logError(e, 'Could not read image data');
+        }
+        this._layout = this.create_pango_layout(this._text);
+
+        this.connect('style-set', () => this._update_custom_style());
+        this.connect('style-updated', () => this._update_custom_style());
+    },
+
+    set mode(value) {
+        if (this._mode === value)
+            return;
+        this._mode = value;
+        this.queue_draw();
+    },
+
+    get mode() {
+        return this._mode;
+    },
+
+    set text(value) {
+        if (this._text === value)
+            return;
+        this._text = value;
+        this._update_text();
+        this.queue_draw();
+    },
+
+    get text() {
+        return this._text;
+    },
+
+    _update_custom_style: function () {
+        this._sizing = EosKnowledgePrivate.style_context_get_custom_string(this.get_style_context(), 'sizing');
+        if (['size-min', 'auto'].indexOf(this._sizing) == -1) {
+            let error = new Error('Unrecognized style property value for EknDynamicLogo-sizing ' + this._sizing);
+            logError(error);
+        }
+
+        this._text_transform = EosKnowledgePrivate.style_context_get_custom_string(this.get_style_context(), 'text-transform');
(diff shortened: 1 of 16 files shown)


tchx84:
**NOTE:** For reviewer, not intended for MERGE yet.

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,107 @@
...
+ *  * 'image' - only the image is visible.
+ *  * 'text' - only the text is visible.
+ *  * 'auto' - visibility depends on constraints (not implemented).
+ */
+const DynamicLogo = new Knowledge.Class({
+    Name: 'DynamicLogo',
+    Extends: Gtk.Box,
+
+    StyleProperties: {
+        'layout': GObject.ParamSpec.string('layout', 'layout', 'layout',
```
ptomato: Hard to tell if these are more like style properties or regular properties. Do you expect them to be style-able in response to a widget state change?

js/app/modules/banner/app.js
```diff
@@ -42,6 +45,20 @@ const App = new Module.Class({
         props.expand = props.expand || false;
         this.parent(props);
 
+        let file = Gio.File.new_for_uri(IMAGE_URI);
+        if (file.query_exists(null)) {
```
ptomato: I would suggest making it a new module instead of branching on this

data/css/modules/banner/_app.scss
```diff
@@ -1,21 +1,47 @@
...
+        -EknDynamicLogo-layout: "vertical";
+        -EknDynamicLogo-mode: "full";
+
+        &__image {
+            color: white;
+            min-height: 100px;
+            min-width: 100px;
+            -EknThemeableImage-sizing: "size-min";
+            -gtk-icon-source: -gtk-recolor(url('resource:///app/assets/logo-symbolic.svg'));
+        }
```
xjuan: I wonder if there is a better way than the obscure symbolic logo feature.
I think if we add a custom text style property to specify the logo uri we could preppend the css to the svg data itself before rendering/loading the image with librsvg.

Long term solution would be to properly integrate svg in GTK

js/app/widgets/dynamicLogo.js
xjuan: Right, this looks more like a regular property.
The problem would be how to override this property in the yaml file without duplicating the templates

Ideally templates would be proper widget with all the internal children exposed for users to override properties values without having to duplicate the entire hierarchy (my eventual goal once we make gjs not crash when instantiating a JS widget)

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,339 @@
...
+
+    _draw_text_mode: function (cr) {
+        let context = this._get_fresh_state_context();
+        let total_width =  this.get_allocated_width();
+        let total_height = this.get_allocated_height();
+
+        let text_width = total_width > Width.MAX ? Width.MAX : total_width ;
+        let text_height = total_height > Height.MAX ? Height.MAX : total_height;
+
+        let font_desc = context.get_font(context.get_state());
```
xjuan: get_font() is deprecated you can you get() directly with the 'font' property according to docs

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,339 @@
...
+        let pango_layout = PangoCairo.create_layout(cr);
+        pango_layout.set_font_description(font_desc);
+        pango_layout.set_text(this.text, -1);
+
```
xjuan: it would be better if we could cache the layout to improve performance

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,339 @@
...
+    },
+
+    _draw_text_mode: function (cr) {
+        let context = this._get_fresh_state_context();
+        let total_width =  this.get_allocated_width();
+        let total_height = this.get_allocated_height();
+
+        let text_width = total_width > Width.MAX ? Width.MAX : total_width ;
+        let text_height = total_height > Height.MAX ? Height.MAX : total_height;
+
```
xjuan: most of these variables could be moved to _draw_mode() and passed to _draw_*_mode() functions as parameters instead of copy paste in each function

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,339 @@
...
+        this._mode = 'full';
+        this._sizing = 'auto';
+        this._text_transform = 'none';
+
+        this._image = null;
+        let file = Gio.File.new_for_uri(this.image_uri);
+        if (file.query_exists(null)) {
+            let stream = file.read(null);
+            this._image = Rsvg.Handle.new_from_stream_sync(stream, file, 0, null);
+        }
```
tchx84: +1 for handling cases where the image is not a svg, but do we really need the setter considering this property is construct only?

js/app/widgets/dynamicLogo.js
tchx84: ouch, I tried with .get() but does not seem to be introspect-able. Will take a look for other alternatives.

js/app/widgets/dynamicLogo.js
xjuan: I did not realized it was construct only, so no in that case we do not need the setter.
BTW text property will also need to be construct only or have a setter since you need to queue a redraw when it changes.

js/app/widgets/dynamicLogo.js
tchx84: Thanks for the tip, gtk_widget_create_pango_layout did all the trick.

ebassi: `Gtk.Widget.create_pango_layout()` does not cache the `Pango.Layout`, though; you'll have to keep it around yourself in your class.

tchx84:
For the reviewer, I just added an extra fix to the way we set the text-transformed'd text.

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,301 @@
...
+    vfunc_get_preferred_width: function () {
+        let min_width = EosKnowledgePrivate.style_context_get_int(this.get_style_context(),
+                                                                  'min-width',
+                                                                  this.get_state_flags());
+        min_width = min_width ? min_width : this.width_request;
+        min_width = min_width > 0 ? min_width : Width.MIN;
+        let nat_width = this._sizing === 'size-min' ? min_width : Width.MAX;
+
+        let margin = this._get_margin();
+        let extra = margin.left+ margin.right;
```
ebassi: Coding style: missing space between '.left' and '+'.

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,301 @@
...
+        return [min_height + extra, nat_height + extra];
+    },
+
+    vfunc_size_allocate: function (alloc) {
+        let margin = this._get_margin();
+        alloc.x += margin.left;
+        alloc.y += margin.top;
+        alloc.width -= margin.left + margin.right;
+        alloc.height -= margin.top + margin.bottom;
+        this.set_allocation(alloc);
```
ebassi: Why are you tweaking the allocation, here?

I assume because you want to do `get_allocation()` inside `vfunc_draw()` and have the margins already gone. This is not really correct, though.

You should not override the `size_allocate()` virtual function…

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,301 @@
...
+    },
+
+    vfunc_draw: function (cr) {
+        let alloc = this.get_allocation();
```
ebassi: … But you should remove the margin from the allocation inside the `draw()` virtual function.

The reason why you should do it this way is to allow caching paths inside GTK+ to take into account the fact that the margin exists, and avoid a relayout if none is needed.

If the CSS margins change without changing the size of the allocation, GTK+ will just queue a redraw.

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,301 @@
...
+        let max_width = alloc.width > Width.MAX ? Width.MAX : alloc.width ;
+        let max_height = alloc.height > Height.MAX ? Height.MAX : alloc.height;
+
```
ebassi: At this point, you remove the margins from `max_width` and `max_height`, and you apply a translation by the top-left margins to the `cr`:

    max_width -= (margin.left + margin.right);
    max_height -= (margin.top + margin.bottom);
    cr.translate(margin.left, margin.top);

    switch (this._mode) {
        …
    }

tchx84:
@ebassi thanks, that helped to clean it lot more !

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,308 @@
...
+    StyleProperties: {
+        /**
+         * Property: sizing
+         *
+         * -EknDynamicLogo-sizing:
+         *  * 'auto'     - size determined by allocated size. Default.
+         *  * 'size-min' - size determined by min-width and min-height css properties.
+         */
+        'sizing': GObject.ParamSpec.string('sizing', 'sizing', 'sizing',
```
ebassi: This is not a blocker, since it's common in various modules; just a note for the future.

Custom style properties will go away in GTK+ 3.90, which means we'll have to find a replacement for this kind of policy change — likely using an actual GObject property, controlled by the YAML/JSON instead of using CSS.

Alternatively, since GTK+ allows using `min-width` and `min-height`, we could implement `max-width` and `max-height`; e.g.:

  - `auto` would be `max-width: fill-content; max-height: fill-content;`
  - `min-size` would be `max-width: min-content; max-height: min-content;`

js/app/widgets/dynamicLogo.js
```diff
@@ -0,0 +1,308 @@
...
+            GObject.ParamFlags.READWRITE, 'auto'),
+        /**
+         * Property: text-transform
+         *
+         * -EknDynamicLogo-text-transform:
+         *  * 'none'      - no transformation is performed. Default.
+         *  * 'uppercase' - the text is transformed to uppercase.
+         *  * 'lowercase' - the text is transformed to lowercase.
+         */
+        'text-transform': GObject.ParamSpec.string('text-transform', 'text-transform', 'text-transform',
```
ebassi: Same as above, not a blocker. We should really implement `text-transform` upstream.

xjuan: I agree and volunteer I guess :)

PR #1198, opened 2016-09-27
Card.LegacyPost only one line in composite mode

ptomato:
In composite TV mode, we should not let the title of the cards on the
Preset B home page take up more than one line.

https://phabricator.endlessm.com/T13527


tchx84:
**CR PASS**

PR #1199, opened 2016-09-27
T13538

mattdangerw:


mattdangerw:
Let's not merge quite yet, but open for comments

ptomato:
For option parsing, you're welcome to adapt https://github.com/ptomato/jasmine-gjs/blob/master/src/options.js — complete with embedded implementation of leftpad so your internet doesn't break :stuck_out_tongue_closed_eyes:

mattdangerw:
@ptomato actually gapplication comes with its own option parsing features complete with help dialog and more. Just using that.

js/app/application.js
```diff
@@ -53,17 +66,62 @@ const Application = new Knowledge.Class({
...
+
+        if (has_option('resource-path'))
+            this.resource_path = get_option_string('resource-path');
+        let app_resource = Gio.Resource.load(this.resource_path);
+        app_resource._register();
+
+        if (has_option('default-theme'))
+            this._theme = 'default';
+        if (has_option('theme-name'))
+            this._theme = get_option_string('theme-name');
```
ebassi: If you try to use `--default-theme` and `--theme-name` at the same time, we should probably warn, instead of having the last one win.

PR #1200, opened 2016-09-27
Reduce logo size and padding on Preset A composite

ptomato:
For composite TVs, we reduce the padding under the search entry, and
reduce the minimum size of the logo to quite small.

Unfortunately we make things quite ugly at the expense of fitting
everything on the screen at 720x480.

https://phabricator.endlessm.com/T13449

### DIFF
--- a/data/css/mesh.scss
+++ b/data/css/mesh.scss
@@ -74,10 +74,6 @@ $display-font-composite: 'Roboto' !default;
     }
 }
 
-.home-page .ArrangementOverflow {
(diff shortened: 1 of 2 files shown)


tchx84:
@ptomato I added an extra fix so the set and search pages fit in screen properly. I tested your commits and is a **CR PASS** from me, but I will let you take a look at that extra fix I added, if you think is OK, just merge it.

ptomato:
**CR pass** from me as well.

PR #1201, opened 2016-09-27
T11321

rmacqueen:
https://phabricator.endlessm.com/T11321


data/css/modules/layout/_paginatedScrolling.scss
```diff
@@ -0,0 +1,14 @@
+scrollbar.vertical slider {
+    min-height: 100px;
+}
+
+.LayoutPaginatedScrolling {
+    background-color: $background-light-color;
+    .text-button {
```
ptomato: Where's this class coming from? I don't see it in the Glade file

data/css/mesh.scss
```diff
@@ -41,6 +41,7 @@ $display-font-composite: 'Roboto' !default;
 
 @import 'modules/layout/hamburgerBasement';
 @import 'modules/layout/navigation';
+@import 'modules/layout/paginatedScrolling';
```
tchx84: should we import this in news.scss as well?

data/css/modules/layout/_paginatedScrolling.scss
```diff
@@ -0,0 +1,14 @@
...
+      background-color: $primary-medium-color;
+      color: white;
+      padding: 5px;
+      margin: 5px;
+      border-radius: 10px;
+    }
+}
```
tchx84: Do we have some sort of "spec" for what this button should look like? If not, my only suggestion would be to make it look "similar" to other buttons we use, e.g. "See more categories". Just the general look and the go-down-symbolic icon.

philip is also adding a debug button as part of T12422 [here](https://github.com/endlessm/eos-knowledge-lib/pull/1190/commits/c0a4b916d89d2a32619947e3fcc519ce3066fe33#diff-5c14d6ac60c52df7ee198a1bc42cbca2R68).

tests/js/app/modules/layout/testPaginatedScrolling.js
```diff
@@ -0,0 +1,72 @@
+// Copyright 2016 Endless Mobile, Inc.
+
+const Gtk = imports.gi.Gtk;
+
+const Utils = imports.tests.utils;
+Utils.register_gresource();
```
tchx84: is this really needed?

tchx84:
This works really well! Just left some small comments.

data/css/modules/layout/_paginatedScrolling.scss
rmacqueen: It's the class for the GtkButton

data/css/mesh.scss
rmacqueen: yes!

tests/js/app/modules/layout/testPaginatedScrolling.js
rmacqueen: Yea it is because the paginatedScrolling layout uses a `.ui` template file, which has to get compiled into the gresource

data/css/modules/layout/_paginatedScrolling.scss
rmacqueen: OK I just gave it the exact same styling as the TabButton, except without the `go-down-symbolic` because in this case you're not going down anywhere, just loading more content. If we get more designs for this button in the future I can open another issue to add those.

js/app/modules/selection/xapian.js
```diff
@@ -109,6 +109,11 @@ const Xapian = new Module.Class({
     clear: function () {
         this._get_more = null;
         this._query_index = 0;
+        let reset_can_load_more = true;
+        if (reset_can_load_more !== this._can_load_more) {
+            this._can_load_more = reset_can_load_more;
+            this.notify('can-load-more');
+        }
         this._can_load_more = true;
```
tchx84: This statement is a little bit confusing, what does "reset_can_load_more"? Does it determine whether we want to reset the value or not, or is it representing the original value? In any case, do we really need to assign the same value twice?

data/css/mesh.scss
```diff
@@ -49,6 +49,8 @@ $display-font-composite: 'Roboto' !default;
 
 @import 'modules/window/simple';
 
+@import 'widgets/tabButton';
```
tchx84: What about leaving the previous hamburgerBasement partial and only do the import there (like we do in the scrolling partial)? I think is easier to understand why we are importing the module rather than the widget directly. I think is also useful in the way it helps to keep track of where is the button being used.

tchx84:
Looks really good now :) just left two last comments.

data/css/mesh.scss
rmacqueen: Because that would mean that the tabButton styling is contigent on whether we are using the hamburgerBasement template somewhere in the app. If we decided to change the template on the front page to something other than hamburgerBasement, we would then lose the styling for the tab button on the section page. I'd rather they were separate concerns

tchx84: I wasn't suggesting to exclude the paginated scrolling import from the templates, neither to exclude the button widget styling import from paginated scrolling partial.

What I suggested was to import the button styling in hamburger basement partial, and then import the hamburger basement partial in the templates that use it (just like we do with the scrolling).

We would not lose the button styling in the section page because we are also importing it in the paginated scrolling partial, and if we change the front-page to some other module that also uses that button, then that other module partial should also import the button.

My point is, the button widget styling makes more sense inside the modules that use it, because technically the button is just an implementation detail of that module.

rmacqueen: ok got it. fixed

tchx84:
**CR PASS**

PR #1202, opened 2016-10-07
Moltres is dead

rmacqueen:
With the arrival of watson's new command line option,
moltres itself doesn't really offer much value add.
Because it is part of the runtime, you can't run moltres
unless you install one of the apps anyway, at which point
you might as well just run the app with a custom .yaml file.

MoltresEngine still survives of course, so that we can
have dummy content when testing apps.


js/app/application.js
```diff
@@ -88,6 +89,8 @@ const Application = new Knowledge.Class({
                              'Path to a overrides scss or css file to theme the application', null);
         this.add_main_option('app-json-path', 'J'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.FILENAME,
                              'Path to a yaml or json file to use as a preset', null);
+        this.add_main_option('moltres-content', 'M'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
```
mattdangerw: Not to argue against our word ladder name, but `dummy-content` would probably be a clearer option name here. Especially as we are opening up ekl and documenting all out things.

js/app/application.js
```diff
@@ -141,6 +146,9 @@ const Application = new Knowledge.Class({
     vfunc_startup: function () {
         this.parent();
         let engine = Engine.get_default();
+        if (engine instanceof MoltresEngine.MoltresEngine)
```
mattdangerw: Maybe a nit, but might be nice to split this whole thing into a separate function `_check_for_content` with your early return here.

Would add readability, and make sure we don't have early return under not that usual circumstances that was missing some other setup we need on application startup.

rmacqueen: I'm not sure I totally understood what you were requesting, but have refactored the code that was in `vfunc_startup()` into a separate `_check_for_content()` function.

mattdangerw:
CR pass

PR #1203, opened 2016-10-10
Remove moltres files from debian install

rmacqueen:
Since moltres got replaced by a more feature
rich command line.


mattdangerw:
CR pass

PR #1204, opened 2016-10-11
Add accordion arrangement

rmacqueen:
This is an arrangement that shows cards in
a vertical arrangement layout. Each card
is placed into an expander, which will expand
when the title is clicked.

https://phabricator.endlessm.com/T13573

### DIFF
--- a/js/app/modules/arrangement/accordion.js
+++ b/js/app/modules/arrangement/accordion.js
@@ -1,0 +1,66 @@
+// Copyright 2016 Endless Mobile, Inc.
+
+/* exported Accordion */
+
+const Gtk = imports.gi.Gtk;
+
+const Arrangement = imports.app.interfaces.arrangement;
+const Module = imports.app.interfaces.module;
+
+const TRANSITION_DURATION = 500;
+
+/**
+ * Class: Accordion
+ * Arrangement which puts its cards into an accordion-like widget.
+ * Each segment of the accordion has a title, which is always visible,
+ * and a revealer, in which lives the actual card. Clicking on the title
+ * expands the corresponding revealer.
+ */
+const Accordion = new Module.Class({
+    Name: 'Arrangement.Accordion',
+    Extends: Gtk.Grid,
+    Implements: [Arrangement.Arrangement],
+
+    _init: function (props={}) {
+        props.orientation = Gtk.Orientation.VERTICAL;
+        props.valign = Gtk.Align.START;
+        this.parent(props);
+    },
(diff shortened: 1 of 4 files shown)


js/app/modules/arrangement/accordion.js
```diff
@@ -0,0 +1,65 @@
...
+
+    _init: function (props={}) {
+        props.orientation = Gtk.Orientation.VERTICAL;
+        props.valign = Gtk.Align.START;
+        this.parent(props);
+    },
+
+    // Arrangement override
+    pack_card: function (card) {
+        let label = new Gtk.Button({
```
mattdangerw: Feels like this would be nicer if this too was a card. By default you'd want title card here, but you could also use a similar small card in the future. And you'd get capitalization, highlight and before after decoration support that just GtkButton wont buy you.

Pretty sure having two card types would break the arrangement interface, so probably not a change for this specific PR. More just thinking aloud here...

js/app/modules/arrangement/accordion.js
```diff
@@ -0,0 +1,65 @@
...
+        grid.show_all();
+        this.add(grid);
+    },
+
+    // Arrangement override
+    unpack_card: function (card) {
+        for (let i = 0; i < this.get_children().length; i++) {
+            let chunk = this.get_children()[i];
+            let card = chunk.get_children()[0].get_child();
+            if (card.model.ekn_id === card.model.ekn_id) {
```
tchx84: Also, not sure if removing widgets while iterating will work as expected.

tchx84:
I tried this arrangement with picard, and I get this error message, not sure it makes sense though:

```
(gjs:29581): Gtk-WARNING **: Attempting to add a widget with type EknColorBox to a container of type GtkRevealer, but the widget is already inside a container of type GtkRevealer, please remove the widget from its existing container first.
```

tchx84:
Just a few comments from my side! Back to Dev.

js/app/modules/arrangement/accordion.js
rmacqueen: lol good point. this is dumb

js/app/modules/arrangement/accordion.js
rmacqueen: yea that would be nice. Or even just being able to create a `TextCard` module directly here, without it being defined in a slot. That still doesn't really work well with current module factory setup

tchx84:
**CR PASS**

PR #1206, opened 2016-10-11
T13678

tchx84:
https://phabricator.endlessm.com/T13678

### DIFF
--- a/js/app/modules/layout/responsiveMargins.js
+++ b/js/app/modules/layout/responsiveMargins.js
@@ -124,6 +124,8 @@ const ResponsiveMargins = new Module.Class({
         let margins = this._get_responsive_margins();
         let [min_size, nat_size] = this.get_preferred_size();
         let base_min_width = min_size.width - margins.tiny.left - margins.tiny.right;
+        if (this.get_request_mode() === Gtk.SizeRequestMode.HEIGHT_FOR_WIDTH)
+            min_size.height = nat_size.height;
         let base_min_height = min_size.height - margins.tiny.top - margins.tiny.bottom;
         let margin = margins.tiny;
         ['small', 'medium', 'large', 'xlarge'].forEach((modifier) => {


js/app/modules/layout/responsiveMargins.js
```diff
@@ -124,6 +124,8 @@ const ResponsiveMargins = new Module.Class({
         let margins = this._get_responsive_margins();
         let [min_size, nat_size] = this.get_preferred_size();
         let base_min_width = min_size.width - margins.tiny.left - margins.tiny.right;
+        if (this.get_request_mode() === Gtk.SizeRequestMode.HEIGHT_FOR_WIDTH)
```
mattdangerw: This whole bug is nasty.

Not totally sure I'm understanding what this is trying to do. Could you walk me through the change and what you thinks happening with our allocation cycles? I can see that in height-for-width sizing the nat height will often be smaller than min height here, but it seems incorrect to just assume that here.

A lot of this widget seems to be playing fast and lose height for width sizing, and seems like we might want to just rewrite a bit of it. And long term, I'd really love to see this widget die and be replaced by much better responsive tools.

Another fun fact is that in the search page case you contain a `card/list` a widget which stomps on its own allocation without any attempt to update its preferred width of heights requests. It definitely breaks all sorts of assumptions in gtk and could be contributing to the problems here.

mattdangerw: > This whole bug is nasty.

Not your code by the way, just everything we are working with here :)

tchx84: No worries @mattdangerw, my bad I could not reply until now. You are right, it probably doesn't make sense just to take nat_size.height, but here is the thing:

At some point the ResponsiveMargins widget receives an allocation like (e.g.) 1000x8000, but then get_preferred_size returns something like 1000x8100, so, whatever calculations we make after that (for the margins) won't make any sense, because the allocation is already smaller than the preferred_size.

BUT, the thing is, we set that allocation anyway, regardless of what the preferred_size is. So, I don't quite understand why we use preferred_size for the following margins calculations.

tchx84:
This no longer makes sense so I am closing it.

PR #1208, opened 2016-10-12
Knowledge Lib: Remove code to handle git submodules

sgnn7:
Jasmine was the only module before and has been removed so this is dead
code that needed removing.

https://phabricator.endlessm.com/T13692


ptomato:
Correct, thanks! **CR pass**

PR #1209, opened 2016-10-12
Fix &-selector misuse

ptomato:
It's easy to forget that & repeats all outer selectors too.

https://phabricator.endlessm.com/T11455


tchx84:
**CR PASS**

PR #1213, opened 2016-10-14
EknMediaBin: added initial implementation of a media widget.

xjuan:
EknMediaBin is a frontend widget for GStreamer playbin.

Basic features implemented:
  o Audio/video playback support provided by GStreamer playbin plugin.
  o GTK video sink for output (will use gtkglsink if possible and fallback to gtksink)
  o UI controls implemented with regular Gtk+ widgets, themeable with CSS.
  o Play/Pause, Fullscreen/Unfullscreen and volume buttons.
  o Scale to control seeking and display playback position.
  o Various key bindings to control playback, seeking and entering/leaving fullscreen.
    (space, Right Left Up Down Page_Up Page_Down Home, f Escape)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -224,6 +224,8 @@ CLEANFILES += $(BUILT_SOURCES)
 EXTRA_DIST += \
 	lib/eosknowledgeprivate/ekn-resources.gresource.xml \
 	lib/eosknowledgeprivate/EknRuntimeDocumentViewer.ui \
+	lib/eosknowledgeprivate/EknMediaBin.ui \
+	lib/eosknowledgeprivate/ekn-media-bin.css \
 	$(NULL)
 
 eosknowledgeprivate_headers = \
@@ -236,6 +238,8 @@ eosknowledgeprivate_sources = \
 	lib/eosknowledgeprivate/ekn-resources.c \
 	lib/eosknowledgeprivate/ekn-runtime-document-viewer.h \
 	lib/eosknowledgeprivate/ekn-runtime-document-viewer.c \
+	lib/eosknowledgeprivate/ekn-media-bin.h \
+	lib/eosknowledgeprivate/ekn-media-bin.c \
 	$(NULL)
 
 # Endless Knowledge Apps GUI library
@@ -313,6 +317,20 @@ libscrollplugin_la_CPPFLAGS = \
 	$(NULL)
 libscrollplugin_la_LIBADD = $(SCROLL_PLUGIN_LIBS)
 libscrollplugin_la_LDFLAGS = -module -avoid-version -no-undefined
+
+# # # EXAMPLES # # #
+
+noinst_PROGRAMS = eos-player
+
+eos_player_SOURCES = examples/eos-player.c
+eos_player_CPPFLAGS = \
+	$(EOS_KNOWLEDGE_PRIVATE_CFLAGS) \
+	-I$(top_srcdir)/lib/eosknowledgeprivate/ \
+	$(NULL)
+eos_player_LDADD = \
+	$(EOS_KNOWLEDGE_PRIVATE_LIBS) \
+	$(top_builddir)/libeosknowledgeprivate.la \
+	$(NULL)
 
 # # # SUBSTITUTED FILES # # #
 # These files need to be filled in with make variables
--- a/examples/eos-player.c
+++ b/examples/eos-player.c
@@ -1,0 +1,51 @@
+#include <gst/gst.h>
+#include <ekn-media-bin.h>
+
+static void
+on_title_notify (GObject *gobject, GParamSpec *pspec, GtkWidget *window)
+{
+  gtk_window_set_title (GTK_WINDOW (window),
+                        ekn_media_bin_get_title (EKN_MEDIA_BIN (gobject)));
+}
+
+int
+main (int argc, char **argv)
+{
+  GtkWidget *window, *bin;
+
+  gst_init (&argc, &argv);
+  gtk_init (&argc, &argv);
+
+  if (argc < 2)
+    {
+      g_print ("Usage: %s file:///full/path/to/media/file\n", argv[0]);
+      return 0;
+    }
+
+  /* Make sure animations are enabled, just in case we are running inside a VM */
+  g_object_set (gtk_settings_get_default (), "gtk-enable-animations", TRUE, NULL);
+
+  /* Create main window */
+  window = gtk_window_new (GTK_WINDOW_TOPLEVEL);
+
+  /* Set some sane defaults */
+  gtk_window_set_default_size (GTK_WINDOW (window), 800, 450);
+  gtk_window_set_title (GTK_WINDOW (window), "Eos Player");
+  gtk_window_set_position (GTK_WINDOW (window), GTK_WIN_POS_CENTER);
+
+  /* Quit app on delete-event */
+  g_signal_connect (window, "delete-event", G_CALLBACK (gtk_main_quit), NULL);
+
+  bin = ekn_media_bin_new ();
+  ekn_media_bin_set_uri (EKN_MEDIA_BIN (bin), argv[1]);
+  gtk_container_add (GTK_CONTAINER (window), bin);
+
+  /* Update window title */
+  g_signal_connect (bin, "notify::title", G_CALLBACK (on_title_notify), window);
+
+  gtk_widget_show_all (window);
+
+  gtk_main ();
+
+  return 0;
+}
--- a/lib/eosknowledgeprivate/EknMediaBin.ui
+++ b/lib/eosknowledgeprivate/EknMediaBin.ui
@@ -1,0 +1,229 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.20.0 -->
+<interface>
+  <requires lib="gtk+" version="3.20"/>
+  <object class="GtkImage" id="fullscreen_image">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="icon_name">view-fullscreen</property>
+  </object>
+  <object class="GtkAdjustment" id="playback_adjustment">
+    <property name="upper">128</property>
+    <property name="step_increment">60</property>
+    <property name="page_increment">300</property>
+    <signal name="value-changed" handler="on_playback_adjustment_value_changed" swapped="no"/>
+  </object>
+  <object class="GtkImage" id="playback_image">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="icon_name">gtk-media-play</property>
+  </object>
+  <template class="EknMediaBin" parent="GtkBox">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="orientation">vertical</property>
+    <child>
+      <object class="GtkOverlay" id="overlay">
+        <property name="visible">True</property>
+        <property name="can_focus">True</property>
+        <property name="events">GDK_POINTER_MOTION_MASK | GDK_BUTTON_PRESS_MASK | GDK_BUTTON_RELEASE_MASK | GDK_KEY_PRESS_MASK | GDK_KEY_RELEASE_MASK | GDK_STRUCTURE_MASK</property>
+        <property name="no_show_all">True</property>
+        <signal name="button-press-event" handler="on_overlay_button_press_event" swapped="no"/>
+        <signal name="button-release-event" handler="on_overlay_button_release_event" swapped="no"/>
+        <signal name="motion-notify-event" handler="on_overlay_motion_notify_event" swapped="no"/>
+        <child>
+          <placeholder/>
+        </child>
+        <child type="overlay">
+          <object class="GtkRevealer" id="top_revealer">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="events">GDK_POINTER_MOTION_MASK | GDK_LEAVE_NOTIFY_MASK | GDK_STRUCTURE_MASK</property>
+            <property name="valign">start</property>
+            <signal name="leave-notify-event" handler="on_revealer_leave_notify_event" swapped="no"/>
+            <signal name="motion-notify-event" handler="on_revealer_motion_notify_event" swapped="no"/>
+            <child>
+              <object class="GtkBox">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="orientation">vertical</property>
+                <child>
+                  <object class="GtkLabel" id="title_label">
+                    <property name="can_focus">False</property>
+                    <property name="halign">start</property>
+                    <style>
+                      <class name="title"/>
+                    </style>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">True</property>
+                    <property name="position">0</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkBox" id="info_box">
+                    <property name="can_focus">False</property>
+                    <child>
+                      <placeholder/>
+                    </child>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">True</property>
+                    <property name="position">1</property>
+                  </packing>
+                </child>
+                <style>
+                  <class name="overlay-bar"/>
+                  <class name="top"/>
+                </style>
+              </object>
+            </child>
+          </object>
+          <packing>
+            <property name="index">1</property>
+          </packing>
+        </child>
+        <child type="overlay">
+          <object class="GtkRevealer" id="bottom_revealer">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="events">GDK_POINTER_MOTION_MASK | GDK_LEAVE_NOTIFY_MASK | GDK_STRUCTURE_MASK</property>
+            <property name="valign">end</property>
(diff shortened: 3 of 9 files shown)


lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+ */
+#define EMB_DEFINE_SETTER(type, prop, PROP, code) \
+void \
+ekn_media_bin_set_##prop (EknMediaBin *bin, type prop) \
+{ \
+  g_return_if_fail (EKN_IS_MEDIA_BIN (bin)); \
+  if (bin->prop != prop) \
+    { \
+      bin->prop = prop; \
+      code \
```
ptomato: `G_STMT_START` / `G_STMT_END` around `code`?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+                    "audio-tags-changed");
+  g_signal_connect (bin->play, "video-tags-changed",
+                    G_CALLBACK (on_playbin_video_tags_changed),
+                    "video-tags-changed");
+  g_signal_connect (bin->play, "text-tags-changed",
+                    G_CALLBACK (on_playbin_text_tags_changed),
+                    "text-tags-changed");
+
+  /* Setup volume */
+  /* NOTE: Bidirectional binding makes the app crash on X11 */
```
ptomato: Sounds like a bug that you should report upstream :smile:

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+static gboolean
+ekn_media_bin_bus_watch (GstBus *bus, GstMessage *msg, gpointer data)
+{
+  EknMediaBin *bin = data;
+
+  switch (GST_MESSAGE_TYPE (msg))
+    {
+      case GST_MESSAGE_ERROR:
+        return ekn_media_bin_handle_msg_error (bin, msg);
+      break;
```
ptomato: nitpick: no `break` needed after `return`

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+#define PROP_AUTOHIDE_DEFAULT    2    /* Controls autohide timeout in seconds */
+
+#define EMB_ICON_SIZE            GTK_ICON_SIZE_BUTTON
+
+#define EMB_ICON_NAME_PLAY       "gtk-media-play"
+#define EMB_ICON_NAME_PAUSE      "gtk-media-pause"
+#define EMB_ICON_NAME_FULLSCREEN "view-fullscreen"
+#define EMB_ICON_NAME_RESTORE    "view-restore"
+
+struct _EknMediaBin
```
xjuan: My understanding is that we need a private struct if we want the widget to be derivable

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+  else
+    {
+      gtk_image_set_from_icon_name (bin->playback_image, EMB_ICON_NAME_PLAY, EMB_ICON_SIZE);
+      bin->position = 0;
+      gtk_widget_remove_tick_callback (GTK_WIDGET (bin), bin->tick_id);
+      bin->tick_id = 0;
+    }
+}
+
+#if SHOW_METADATA
```
xjuan: Actually my plan is to make this information available somehow to the user with a keybinding or a right click option, so it wont be inside a #ifdef

lib/eosknowledgeprivate/ekn-media-bin.h
```diff
@@ -0,0 +1,70 @@
...
+ * 
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * Author: Juan Pablo Ugarte <ugarte@endlessm.com>
+ *
+ */
+
+#ifndef _EKN_MEDIA_BIN_H_
```
xjuan: I try to avoid using non standard features, but if we are already using it some other places then its fine with me

lib/eosknowledgeprivate/ekn-resources.gresource.xml
```diff
@@ -2,5 +2,7 @@
 <gresources>
   <gresource prefix="/com/endlessm/knowledge-private">
     <file preprocess="xml-stripblanks">EknRuntimeDocumentViewer.ui</file>
+    <file preprocess="xml-stripblanks">EknMediaBin.ui</file>
```
xjuan: I am linking statically with this resource, so if I set it to compressed then we would end up with the compressed data in the text area and the decompresed data in the heap.

So we either waste some disk space or some memory :S

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+ */
+GtkWidget *
+ekn_media_bin_new ()
+{
+  return (GtkWidget*) g_object_new (EKN_TYPE_MEDIA_BIN, NULL);
+}
+
+/* The following macros are used to define generic getters and setters */
+
+#define EMB_DEFINE_GETTER(type, prop, retval) \
```
xjuan: hehe yeah I really do not like doing copy/paste of code

lib/eosknowledgeprivate/ekn-media-bin.c
xjuan: yeah good catch!

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1462 @@
...
+ *
+ */
+void
+ekn_media_bin_play (EknMediaBin *bin)
+{
+  g_return_if_fail (EKN_IS_MEDIA_BIN (bin));
+
+  g_object_set (bin->play, "uri", bin->uri, NULL);
+
+  bin->state = GST_STATE_PLAYING;
```
xjuan: Its not for performance actually, ->state is meant to keep the user desired state, which is not always the same as the pipeline state.
I am probably going ahead of me here since we do not really need it right now, but I am pretty sure we will eventually.

Say we are streaming some video and the buffer runs out so we have to pause the pipeline, then once the buffer catch up, we wont know which stated the used wanted :S

In any case it deserves a comment :)

lib/eosknowledgeprivate/ekn-media-bin.c
xjuan: hmm now that I think more about it, we do not need them since this macro should not be used as a statement, its defining a function.

lib/eosknowledgeprivate/ekn-resources.gresource.xml
ptomato: Good point! I never thought of that.

lib/eosknowledgeprivate/ekn-media-bin.h
ptomato: A while ago we agreed that we'd consider `#pragma once` standard in Endless code

lib/eosknowledgeprivate/ekn-resources.gresource.xml
xjuan: Which reminds me I started writing a compact binary formar for GMarkupParseContext and I never finish it.
But the main idea is to have a compact generic binary format that would allow us to avoid all the heavy part of parsing the xml in a template and do something like

```<file preprocess="xml-convert-binary">EknMediaBin.ui</file>```

File: examples/eos-player.c [hunk 10]
```diff
@@ -0,0 +1,57 @@
...
+#include <ekn-media-bin.h>
+
+static void
+on_title_notify (GObject *gobject, GParamSpec *pspec, GtkWidget *window)
+{
+  gtk_window_set_title (GTK_WINDOW (window),
+                        ekn_media_bin_get_title (EKN_MEDIA_BIN (gobject)));
+}
+
+/*
```
ptomato: Since you have a makefile rule, you probably don't need this comment anymore

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+      gst_element_set_state (self->play, GST_STATE_PAUSED);
+    }
+}
+
+static void
+ekn_media_bin_fullscreen_apply (EknMediaBin *self, gboolean fullscreen)
+{
+  if (fullscreen)
+    {
+#if USE_TMP_IMAGE
```
ptomato: What information do you still need to decide whether to use this code path or not?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+{
+  ekn_media_bin_toggle_playback (self);
+}
+
+static gboolean
+on_overlay_button_press_event (GtkWidget   *widget,
+                               GdkEvent    *event,
+                               EknMediaBin *self)
+{
+  if (event->button.button != 1)
```
ptomato: (below as well)

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+      gint       width  = gtk_widget_get_allocated_width (self->overlay);
+      gint       height = gtk_widget_get_allocated_height (self->overlay);
+      gint       w      = gdk_pixbuf_get_width (pixbuf);
+      gint       h      = gdk_pixbuf_get_height (pixbuf);
+      gdouble    xscale = width / (gdouble)w;
+      gdouble    yscale = height / (gdouble)h;
+      gdouble    scale  = MIN (xscale, yscale);
+      GdkPixbuf *scaled;
+
+      scaled = gdk_pixbuf_scale_simple (pixbuf, w*scale, h*scale, GDK_INTERP_BILINEAR);
```
ptomato: I think this is leaked

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+
+  properties[PROP_VOLUME] =
+    g_param_spec_double ("volume",
+                         "Volume",
+                         "Stream volume",
+                         0.0, 1.0, 1.0,
+                         G_PARAM_READWRITE);
+
+  properties[PROP_AUTOHIDE] =
+    g_param_spec_int ("autohide",
```
ptomato: nitpick: Suggest calling this `autohide-timeout`, just `autohide` sounds like it should be a boolean

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+  object_class->set_property = ekn_media_bin_set_property;
+  object_class->get_property = ekn_media_bin_get_property;
+
+  widget_class->realize = ekn_media_bin_realize;
+
+  /* Properties */
+  properties[PROP_URI] =
+    g_param_spec_string ("uri",
+                         "URI",
+                         "The Media URI to playbacks",
```
ptomato: nitpick: playbacks → playback

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+      gchar *value = NULL;
+
+      if (!self->title_user_set)
+        {
+          if (self->video_tags)
+            gst_tag_list_get_string_index (self->video_tags, GST_TAG_TITLE, 0, &value)
+
+          ekn_media_bin_set_title (self, (self->video_tags && value) ? value : "");
+          self->title_user_set = FALSE;
+          g_free (value);
```
ptomato: nitpick: You could use `g_clear_pointer(&value, g_free)` here to save the extra assignment of `NULL` in the case where description is not set and title is

lib/eosknowledgeprivate/EknMediaBin.ui
```diff
@@ -0,0 +1,292 @@
...
+                    <property name="expand">False</property>
+                    <property name="fill">True</property>
+                    <property name="position">0</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkBox" id="info_box">
+                    <property name="can_focus">False</property>
+                    <child>
+                      <object class="GtkLabel" id="info_1_label">
```
ptomato: nitpick: Could be more descriptive than 1­ to 6

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+
+  /* TODO: handle audio and text tags */
+  if (g_str_equal (name, "video-tags-changed"))
+    {
+      gchar *value = NULL;
+
+      if (!self->title_user_set)
+        {
+          if (self->video_tags)
+            gst_tag_list_get_string_index (self->video_tags, GST_TAG_TITLE, 0, &value)
```
ptomato: Missing semicolon

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -0,0 +1,1577 @@
...
+
+      if (!self->description_user_set)
+        {
+          value = NULL;
+
+          if (self->video_tags &&
+              (gst_tag_list_get_string_index (self->video_tags, GST_TAG_COMMENT, 0, &value) ||
+               gst_tag_list_get_string_index (self->video_tags, GST_TAG_DESCRIPTION, 0, &value));
+
+          ekn_media_bin_set_description (self, (self->video_tags && value) ? value : ""));
```
ptomato: These statements should be in braces belonging to the above `if` statement?

Makefile.am
```diff
@@ -314,6 +318,21 @@ libscrollplugin_la_CPPFLAGS = \
 libscrollplugin_la_LIBADD = $(SCROLL_PLUGIN_LIBS)
 libscrollplugin_la_LDFLAGS = -module -avoid-version -no-undefined
 
+# # # EXAMPLES # # #
+
+noinst_PROGRAMS = eos_player
+
+eos_player_SOURCES = examples/eos-player.c
+eos_player_CFLAGS = $(EOS_KNOWLEDGE_PRIVATE_CFLAGS)
```
xjuan: yeah I always get confused about it!

xjuan:
I tried with this file and it works
```[jh] (T13359) ~/checkout/eos-knowledge-lib$ md5sum /home/xjuan/checkout/big_buck_bunny_480p_stereo.ogg
1430e16238705e3689debad80e288903  /home/xjuan/checkout/big_buck_bunny_480p_stereo.ogg```

you are probably missing some gstreamer plugins try installing all of them (good bad and ugly)

File: lib/eosknowledgeprivate/ekn-media-bin.c [hunk 21]
```diff
@@ -0,0 +1,1606 @@
...
+ * ekn_media_bin_set_fullscreen:
+ * @self: a #EknMediaBin
+ * @fullscreen:
+ *
+ * Sets whether to show the video in fullscreen mode or not
+ */
+EMB_DEFINE_SETTER (gboolean, fullscreen, FULLSCREEN,
+  /* If there is no video sink, delay fullscreen until realize event */
+  if (self->video_sink)
+    ekn_media_bin_fullscreen_apply (self, fullscreen);
```
ptomato: Obviously this is working, but just out of curiosity, how? Shouldn't everything after the comma in this line be treated like a separate argument to the macro?

lib/eosknowledgeprivate/ekn-media-bin.c
xjuan: Right, but this is not public since its defined in the .c file
Of course people could copy paste the struct in their code and access it but the same thing can be done with the private struct

lib/eosknowledgeprivate/ekn-media-bin.c
xjuan: it works because the comma is inside balanced parenthesis :)

xjuan:
Yeah I hope that it all depends on using a newer version of gstreamer, Tomorrow I will install debian sid in my netbook and try it out.

BTW I will add a private struct

ptomato:
I've been thinking about how to have a good experience on drivers where the gtkglsink is broken, so I wanted to post this before I forget... at least on my machine, it only errors out when the pipeline goes into the preroll state. Is there a way we could do some sort of test before playing the video and if it errors out switch back to the cairo sink?

xjuan:
I added a check to see if the gtkglsink works, if not it will fallback to gtksink
You can check which sink EknMediaBin decided to use with
```GST_DEBUG=EknMediaBin:4 ./eos-player file://```

File: Makefile.am [hunk 22]
```diff
@@ -314,6 +318,20 @@ libscrollplugin_la_CPPFLAGS = \
 libscrollplugin_la_LIBADD = $(SCROLL_PLUGIN_LIBS)
 libscrollplugin_la_LDFLAGS = -module -avoid-version -no-undefined
 
+# # # EXAMPLES # # #
+
+noinst_PROGRAMS = eos-player
+
+eos_player_SOURCES = examples/eos-player.c
+eos_player_CPPFLAGS = \
+        $(EOS_KNOWLEDGE_PRIVATE_CFLAGS) \
```
ptomato: Mixed tabs and spaces in these two assignments — careful, they can ruin your makefiles!

PR #1214, opened 2016-10-14
Knowledge lib: RFC Get allHighlights to show featured articles first

sgnn7:
https://phabricator.endlessm.com/T10252

### DIFF
--- a/js/app/modules/selection/allHighlights.js
+++ b/js/app/modules/selection/allHighlights.js
@@ -18,11 +18,21 @@ const AllHighlights = new Module.Class({
     },
 
     construct_query_object: function (limit, query_index) {
-        if (query_index > 0)
-            return null;
-        return new QueryObject.QueryObject({
-            limit: limit,
-            tags_match_any: ['EknArticleObject'], // FIXME: Should be getting featured articles only
-        });
+        if (query_index == 0) {
+            return new QueryObject.QueryObject({
+                limit: limit,
+                tags_match_all: ['EknArticleObject', 'EknFeaturedTag'],
+                excluded_tags: [],
+            });
+        } else if (query_index == 1) {
+            return new QueryObject.QueryObject({
+                limit: limit,
+                tags_match_all: ['EknArticleObject'],
+                excluded_tags: ['EknFeaturedTag'],
+            });
+        }
+
+        // All other indexes (>= 2) should return null
+        return null;
     },
 });
--- a/js/app/modules/selection/xapian.js
+++ b/js/app/modules/selection/xapian.js
@@ -30,7 +30,7 @@ const Xapian = new Module.Class({
     _init: function (props={}) {
         this._loading = false;
         this._can_load_more = true;
-        this._get_more = null;
+        this._next_query = null;
         this._query_index = 0;
         this._error_state = false;
         this._exception = null;
@@ -61,7 +61,7 @@ const Xapian = new Module.Class({
             return;
 
         let engine = Engine.get_default();
-        let query = this._get_more;
+        let query = this._next_query;
 
         if (!query) {
             let limit = num_desired;
@@ -140,7 +140,7 @@ const Xapian = new Module.Class({
                 more_results_query = this.construct_query_object(num_desired, this._query_index);
             }
 
-            this._get_more = more_results_query;
+            this._next_query = more_results_query;
             let can_load_more = !!more_results_query && (info.upper_bound > results.length);
             if (can_load_more !== this._can_load_more) {
                 this._can_load_more = can_load_more;
@@ -155,7 +155,7 @@ const Xapian = new Module.Class({
     },
 
     clear: function () {
-        this._get_more = null;
+        this._next_query = null;
         this._query_index = 0;
 
         let old_value = this._can_load_more;
--- a/js/app/modules/selection/xapian.js
+++ b/js/app/modules/selection/xapian.js
@@ -135,20 +135,23 @@ const Xapian = new Module.Class({
                 });
             }
 
+            let new_query = false;
             if (!more_results_query) {
                 this._query_index++;
+                new_query = true;
                 more_results_query = this.construct_query_object(num_desired, this._query_index);
             }
 
             this._next_query = more_results_query;
-            let can_load_more = !!more_results_query && (info.upper_bound > results.length);
+            let can_load_more = !!more_results_query && (info.upper_bound > results.length || new_query);
             if (can_load_more !== this._can_load_more) {
                 this._can_load_more = can_load_more;
                 this.notify('can-load-more');
             }
 
-            if (num_results_added < num_desired && this._can_load_more)
+            if (num_results_added < num_desired && this._can_load_more) {
                 this.queue_load_more(num_desired - num_results_added);
+            }
 
             this.emit_models_when_not_animating();
         });


js/app/modules/selection/xapian.js
```diff
@@ -85,24 +85,43 @@ const Xapian = new Module.Class({
                 return;
             }
 
-            let next_query = info.more_results;
-            if (!next_query) {
+            let results_added = results.filter(this.add_model, this);
+            let need_more_results = num_desired > results_added.length;
+
+            let more_results = info.more_results;
+            if (!!more_results) {
```
ptomato: Just use `if (more_results)`

js/app/modules/selection/xapian.js
```diff
@@ -85,24 +85,43 @@ const Xapian = new Module.Class({
...
                 this._query_index++;
-                next_query = this.construct_query_object(num_desired, this._query_index);
-            }
+                this._next_query = this.construct_query_object(num_desired,
+                                                               this._query_index);
+
+                // We have a new query. We only assume that we can get more results
+                // if we filled up the desired count. If the next query isn't defined,
+                // we know for sure that we won't be able to fill the results array.
+                let can_load_more = !need_more_results || !!this._next_query;
```
ptomato: I would take `!need_more_results` out of here; `can_load_more` should be independent of whether the caller actually wants more.

js/app/modules/selection/xapian.js
```diff
@@ -85,24 +85,43 @@ const Xapian = new Module.Class({
...
+                    this.notify('can-load-more');
+                }
 
-            this._next_query = next_query;
-            let can_load_more = !!next_query && (info.upper_bound > results.length ||
-                                                 num_desired > results.length);
-            if (can_load_more !== this._can_load_more) {
-                this._can_load_more = can_load_more;
-                this.notify('can-load-more');
+                if (need_more_results && !!this._next_query)
```
ptomato: Use `can_load_more` here

js/app/modules/selection/xapian.js
```diff
@@ -92,7 +92,8 @@ const Xapian = new Module.Class({
             }
 
             this._get_more = more;
-            let can_load_more = !!more && (info.upper_bound > results.length);
+            let can_load_more = !!more && (info.upper_bound > results.length ||
+                                           num_desired > results.length);
```
ptomato: I'm not sure I agree with this change, `can_load_more` should be independent of whether more results were requested.

js/app/modules/selection/xapian.js
```diff
@@ -85,24 +85,43 @@ const Xapian = new Module.Class({
...
-                next_query = this.construct_query_object(num_desired, this._query_index);
-            }
+                this._next_query = this.construct_query_object(num_desired,
+                                                               this._query_index);
+
+                // We have a new query. We only assume that we can get more results
+                // if we filled up the desired count. If the next query isn't defined,
+                // we know for sure that we won't be able to fill the results array.
+                let can_load_more = !need_more_results || !!this._next_query;
+                if (can_load_more !== this._can_load_more) {
```
ptomato: Can you deduplicate this setting of `this._can_load_more` below the if statement?

js/app/modules/selection/allHighlights.js
```diff
@@ -18,11 +18,23 @@ const AllHighlights = new Module.Class({
     },
 
     construct_query_object: function (limit, query_index) {
-        if (query_index > 0)
-            return null;
+        let tags_filter = ['EknArticleObject'];
+        let excluded_tags = ['EknFeaturedTag'];
+        switch (query_index) {
```
sgnn7: Yeah I can do that - I mostly took a look at another selection filter with multiple queries for stylistic reference to keep consistency and it used the switch statement: https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/modules/selection/supplementaryArticles.js#L31-L39

js/app/modules/selection/xapian.js
sgnn7: It somewhat of a tricky logic here and I'm not sure if the comment applies but I'll be happy to change it if needed.

In essence, if the query didn't get us all the results we wanted and we don't have a new query, there should be no case where re-running that same query will give us more new results so `can_load_more` in those cases should be `false`. If we don't do this check, if we got partial results we will have "can-load-more" buttons enabled when there's no more data to fetch which is imo an error.

js/app/modules/selection/xapian.js
sgnn7: Do you mind elaborating on this just a tiny bit (I'm not sure what change you'd like to see here)?

sgnn7: Do you mind elaborating on this just a tiny bit (I'm not sure what change you'd like to see here)?

js/app/modules/selection/xapian.js
sgnn7: It somewhat of a tricky logic here and I'm not sure if the comment applies but I'll be happy to change it if needed.

In essence, if the query didn't get us all the results we wanted and we don't have a new query, there should be no case where re-running that same query will give us more new results so can_load_more in those cases should be false. If we don't do this check, if we got partial results we will have "can-load-more" buttons enabled when there's no more data to fetch which is imo an error.

js/app/modules/selection/xapian.js
ptomato: I mean the stanza
```js
if (can_load_more !== this._can_load_more) {
    this._can_load_more = can_load_more;
    this.notify('can-load-more');
}
```
is the same in both branches and can probably be moved below the if statement.

sgnn7: oh, I see what you mean!

sgnn7: Hmm... now that I think of it, do we really want to update the button before recursing into the query function? Let me see if this is doable.

js/app/modules/selection/xapian.js
sgnn7: @ptomato Hmm, I see your point now but effectively `need_more_results` == 'is query exhausted'. Ideally the model only takes the exact number of articles it originally asked for with `this.add_model` so the same query _may_ have results that weren't added the first time around so we do a slight assumption that if we filled up the needed results, there might be more results and if we didn't fill up the results we know for sure that we can't load more.

js/app/modules/selection/xapian.js
sgnn7: :+1:

js/app/modules/selection/xapian.js
```diff
@@ -84,30 +84,50 @@ const Xapian = new Module.Class({
                 }
                 return;
             }
-            let more = info.more_results;
 
-            if (!more) {
+            let results_added = results.filter(this.add_model, this);
+            let need_more_results = num_desired > results_added.length;
```
rmacqueen: Yea so basically what I'll be changing is this line, so that we don't call `add_model` on _more_ results than was originally requested by `num_desired`

sgnn7:
Ouchie... this code snippet got heavily reworked in the meantime. I'll see if I can redo the changes on the new version.

sgnn7:
@ptomato / @rmacqueen: CR ready again. Much smaller implementation since the code got cleaned up a bit in the meantime so the refactoring wasn't needed.

ptomato:
**CR pass**

PR #1216, opened 2016-10-18
T13789

wgreenberg:


js/tools/kermit.js
```diff
@@ -444,16 +442,19 @@ function _pretty_print_table (elements) {
...
             let value = element[column].toString();
             let spaces = ' '.repeat(width - value.length);
             return str + value + spaces + ' ';
         }, '');
-        print(row);
+        return row;
     });
+
+    print(header_str);
+    rows.forEach(row => print(row));
```
wgreenberg: I thought so too, but it breaks since `forEach` passes a bunch of other args into its callback:

```gjs> x = ['foo', 'bar']
foo,bar
gjs> x.forEach(print)
foo 0 foo,bar
bar 1 foo,bar```

wgreenberg:
addressed the crash here: https://github.com/endlessm/eos-shard/pull/61

ptomato:
**CR pass**, but needs to be merged after eos-shard is built with that version number

wgreenberg:
eos-shard version commit landed, so merging this now

PR #1217, opened 2016-10-18
Never add more than the num_desired results to content group

rmacqueen:
When we have a filter present, we naively request three times
the original num_desired value, to account for the filter.
This results in us sometimes getting back more results than
we originally asked for. In that case we don't want to add those
extra results to the models map.

### DIFF
--- a/js/app/modules/selection/xapian.js
+++ b/js/app/modules/selection/xapian.js
@@ -4,8 +4,24 @@
 
 const Engine = imports.search.engine;
 const Module = imports.app.interfaces.module;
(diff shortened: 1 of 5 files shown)


rmacqueen:
> Document the new behaviour in selection.js

Do you mean document the new behaviour in Selection.Xapian? This logic is all particular to xapian queries, so I don't think it should be mentioned in the selection.js superclass

rmacqueen:
Ready again

ptomato:
**CR pass**

PR #1219, opened 2016-10-19
Make sure exception is not null before retrieving stack trace

rmacqueen:
Otherwise we can get a null reference error.


ptomato:
**CR pass**

PR #1220, opened 2016-10-19
control: add gir1.2-rsvg-2.0 dependency

mattdangerw:
We need this at build time for tests and on install for the new
dynamic logo module
https://phabricator.endlessm.com/T13437


ebassi:
CR pass.

ebassi:
Dang, @ptomato was faster, this time... ;-)

PR #1222, opened 2016-10-21
T13535all

rmacqueen:


js/app/modules/selection/subsetsForSet.js
```diff
@@ -10,13 +12,27 @@ const Xapian = imports.app.modules.selection.xapian;
...
         this.parent(props);
         if (this.global) {
-            this.model = HistoryStore.get_default().current_set;
-            HistoryStore.get_default().connect('notify::current-set', () => {
-                this.model = HistoryStore.get_default().current_set;
+            let property_name = 'current-set';
+            if (this.top_level_only) {
+                property_name = 'current-top-level-set';
+            }
+            this.model = HistoryStore.get_default()[property_name.replace('-', '_', 'g')];
```
mattdangerw: Did you ever try to that gjs patch? Would love to fix this up without carry workarounds for gjs quirks...

js/app/modules/banner/title.js
```diff
@@ -0,0 +1,61 @@
...
+
+        let subtitle = app_info.get_description();
+        if (this.show_subtitle && subtitle) {
+            this._subtitle_label.label = subtitle;
+            this._subtitle_label.justify = Utils.alignment_to_justification(this.halign);
+        }
+        this._subtitle_label.visible = this.show_subtitle;
+    },
+
+    set subtitle(value) {
```
mattdangerw: nit: space before (value). i think?

js/app/interfaces/card.js
```diff
@@ -365,18 +365,25 @@ const Card = new Lang.Interface({
      *
      * Sets up a frame to show the model's thumbnail uri.
      */
-    set_thumbnail_frame_from_model: function (frame) {
+    set_thumbnail_frame_from_model: function (frame, background_size='cover') {
         frame.visible = false;
         if (!this.model.thumbnail_uri)
             return;
+        if (background_size === 'center') {
+            let im = new Gtk.Image({
```
mattdangerw: You sure about this? If a set or article has a really big image it would not be able to size down here. Seems like we could be triggering resolution support bugs (to big for composite) with content updates, which sounds scary to me.

Also nit: s/im/image/

mattdangerw: Maybe we just pull the trigger on https://phabricator.endlessm.com/T12679 whether or not gsk finds it way into our gtk this milestone. Then you could try to do all this with css and keep things more responsive.

js/app/modules/banner/title.js
```diff
@@ -0,0 +1,61 @@
...
+    Template: 'resource:///com/endlessm/knowledge/data/widgets/banner/title.ui',
+    InternalChildren: [ 'title-label', 'subtitle-label' ],
+
+    _init: function (props={}) {
+        this.parent(props);
+
+        let app_info = Utils.get_desktop_app_info();
+
+        let title = app_info.get_display_name();
+        this._title_label.label = title;
```
mattdangerw: Maybe some sync up with dynamic logo in order here? Is there ever a reason to show something different in this text then with our logo system text? @tchx84 @rmacqueen ?

js/app/modules/arrangement/accordion.js
```diff
@@ -32,6 +32,8 @@ const Accordion = new Module.Class({
         let label = new Gtk.Button({
             label: card.model.title,
         });
+        // FIXME: Is there a less intrusive way to do this?
+        label.get_child().halign = Gtk.Align.START;
```
mattdangerw: I would just make you own label in this case and pack it in button directly (all gtk button is doing internally i think)

js/app/modules/card/thumb.js
```diff
@@ -166,7 +166,7 @@ const Thumb = new Module.Class({
             this.set_label_or_hide(this._synopsis_label, this.model.synopsis);
             this._title_label.valign = Gtk.Align.END;
         } else {
-            this._title_label.valign = Gtk.Align.CENTER;
+            this._title_label.valign = Gtk.Align.START;
```
mattdangerw: We are ok with this change everywhere for thumb cards?

data/css/course.scss
```diff
@@ -0,0 +1,187 @@
...
+
+$body-font: Roboto !default;
+$title-font: Roboto !default;
+$display-font: Roboto !default;
+$context-font: Roboto !default;
+$support-font: 'Montserrat' !default;
+
+/* === Colors and Fonts === */
+
+@import 'assets';
```
mattdangerw: If https://phabricator.endlessm.com/T13798 lands this will need to be reworked to build off our default theme.

js/app/modules/controller/course.js
```diff
@@ -0,0 +1,63 @@
...
+const QueryObject = imports.search.queryObject;
+
+/**
+ * Class: Course
+ * Controller that presents content as a course to be consumed.
+ *
+ * Implements:
+ *    <Module>, <Controller>
+ */
+const Course = new Module.Class({
```
mattdangerw: With Course and Buffet, I saw we unnaturally force all controller names to match a dining theme. We can start with Mesh -> Strainer.

js/app/modules/controller/course.js
```diff
@@ -0,0 +1,63 @@
...
+const ReadingHistoryModel = imports.app.readingHistoryModel;
+const SetMap = imports.app.setMap;
+const QueryObject = imports.search.queryObject;
+
+/**
+ * Class: Course
+ * Controller that presents content as a course to be consumed.
+ *
+ * Implements:
+ *    <Module>, <Controller>
```
mattdangerw: nit: but I'd leave this to code to avoid rot

js/app/modules/controller/course.js
ptomato: :joy:

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -41,6 +39,31 @@ const ContentGroup = new Module.Class({
...
+        'paginate': GObject.ParamSpec.boolean('paginate',
+            'Paginate', 'Whether to paginate content',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            false),
+        /**
+         * Property: cards-per-page
+         * Number of cards to be displayed per page
+         *
+         * Default:
+         *   10
```
ptomato: Yep, I wanted to be as explicit as possible in doc comments in the past, but we can probably drop this ... autogenerating is the first big task on my list after the SpiderMonkey upgrade

js/app/modules/selection/subsetsForSet.js
```diff
@@ -10,13 +12,27 @@ const Xapian = imports.app.modules.selection.xapian;
 const SubsetsForSet = new Module.Class({
     Name: 'Selection.SubsetsForSet',
     Extends: Xapian.Xapian,
+    Properties: {
+        /**
+         * Property: top-level-only
+         * Whether this content group has more content to show
```
ptomato: Description seems wrong

js/search/mediaObjectModel.js
```diff
@@ -97,10 +97,10 @@ const VideoObjectModel = Lang.Class({
          * The duration of the video in ISO 8601 format. Defaults to empty
```
rmacqueen: The shard I got from @ffarfan had the value as an int. I assume it was meant to be in milliseconds?

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -114,7 +137,23 @@ const ContentGroup = new Module.Class({
             no_show_all: true,
             vexpand: true,
         });
-        this._stack.add_named(this._arrangement, CONTENT_PAGE_NAME);
+
+        let see_more_button = new Gtk.Button({
+            halign: Gtk.Align.CENTER,
+            label: 'See more', // Gettext this
```
rmacqueen: Hmm, I'd like to minimize the customizable surface area as much as possible. Unlike with the trigger, I don't see a reason why this would need to say anything other than "See more"

js/app/modules/card/thumb.js
rmacqueen: ya. this only makes a difference for horizontally orientated thumb cards, which as far as I can see, never actually show up in any existing apps

js/app/interfaces/card.js
rmacqueen: Well this is only meant for background images which are small icons. Yea you could misuse it and would look bad but that's true about a number of our modules

js/app/interfaces/card.js
```diff
@@ -283,6 +275,69 @@ const Card = new Lang.Interface({
...
+            }
+        });
+    },
+
+    _create_inventory_widget: function () {
+        this._child_count = 0;
+        let label = new FormattableLabel.FormattableLabel();
+
+        this._count_set(this.model, (count) => {
+            label.label = count + ' lessons';
```
rmacqueen: > "Lessons" seems overly specific for a context label on all SetObjectModels. Is there a better way we could identify a set of lessons??

Not really - I dont want to make a separate LessonObject I think. I can make this text customizable though

js/app/modules/selection/subsetsForSet.js
rmacqueen: I did not. Now I can't find it. Link?

rmacqueen:
> I never quite understood why the pagination layout had to be reverted and we had to build it into the content group. Can you explain again? (Also would be good to have the explanation in the revert commit message.)

Previously we had made pagination layout its own separate module, with its own scrolled window. The problem was when we put this module inside another scrolled window module, e.g. Layout.Scrolling, and the you would have two layers of scrolling on the same page which doesn't make sense. Having it in the content group also makes more sense because it is there that we have direct access to the selection

rmacqueen:
> For counting the number of articles in a set, I seem to remember you talking about adding Xapian bridge API to do that, did that ever go anywhere?

I dont think that would actually make sense. What constitutes a 'child' articles is just determined by tags, so theres nothing xapian could do that would be quicker/easier than the multiple tag queries we are doing now

ptomato:
> Previously we had made pagination layout its own separate module, with its own scrolled window. The problem was when we put this module inside another scrolled window module, e.g. Layout.Scrolling, and the you would have two layers of scrolling on the same page which doesn't make sense. Having it in the content group also makes more sense because it is there that we have direct access to the selection

Why not simply remove the scrolled window out of the pagination layout, in that case?

I was going to say that now we have two separate mechanisms for pagination, one with the infinite scrolling layout that has a reference to a content group that it should lazy-load, and one built into the content group. However I think I see where you're going with this, it seems like maybe there's not much left of the pagination layout if we remove the scrolled window? other than the pagination button and a means of communicating with the content group, and the latter can be removed entirely if the pagination button is part of the content group? Would you say that's accurate?

I'd still be concerned about the content group turning into a God Class, but that's something that should maybe come up for refactoring in 3.2.

In any case I'd prefer to see the reasons summarized on the commit message of the revert commit.

js/app/interfaces/card.js
```diff
@@ -248,19 +264,7 @@ const Card = new Lang.Interface({
...
-     * associated with this article. Tags are also incidentally the same
-     * as set titles.
-     *
-     * According to designs (and based on what fits on screen), we only want
-     * to show a maximum of two set titles in the context label.
-     * Moreover, we want to exclude the 'Ekn'-prefixed tags which won't mean
-     * anything to the user.
-     */
-    create_context_widget_from_model: function () {
+    _create_contextual_tag_wiget: function () {
```
ptomato: typo: wiget -> widget

js/app/interfaces/card.js
```diff
@@ -157,6 +160,18 @@ const Card = new Lang.Interface({
...
             ''),
         /**
+         * Property: collection-name
+         * The name of whatever it is a set card contains
+         *
+         * For sets, we want to show how many items are contained within it.
+         * To do this we need to know the name of the things being contained,
+         * e.g. 'lessons' or 'articles'.
+         */
+        'collection-name': GObject.ParamSpec.string('collection-name',
```
ptomato: I'm not so sure about this one; first of all it seems like it should be part of the selection. But what to do about sets that contain, let's say, lessons, articles, videos, and podcasts all together?

ptomato: What about singulars and plurals, e.g. `1 LESSONS`?

js/app/interfaces/card.js
```diff
@@ -283,6 +287,69 @@ const Card = new Lang.Interface({
...
+            }
+        });
+    },
+
+    _create_inventory_widget: function () {
+        this._child_count = 0;
+        let label = new FormattableLabel.FormattableLabel();
+
+        this._count_set(this.model, (count) => {
+            label.label = count + ' ' + this.collection_name;
```
ptomato: Use string formatting with Gettext so that it can be localized for languages which put the number after the noun.

js/app/modules/card/thumb.js
```diff
@@ -125,9 +140,19 @@ const Thumb = new Module.Class({
...
+        this.connect('style-set', this._update_custom_style.bind(this));
+
         Utils.set_hand_cursor_on_widget(this);
     },
 
+    _update_custom_style: function () {
+        let context = this.get_style_context();
+        let thumb_background_size = EosKnowledgePrivate.style_context_get_custom_string(
+            context, 'thumbnail-background-size');
+        this.set_thumbnail_frame_from_model(this._thumbnail_frame, thumb_background_size);
```
ptomato: In any case, this should not call `set_thumbnail_frame_from_model` if the background size property didn't change.

Calling it more than once is problematic in the current code, because it doesn't remove the previous widget from the frame, so any subsequent calls will fail because the frame already has a child.

js/app/modules/card/list.js
```diff
@@ -30,6 +30,21 @@ const List = new Module.Class({
...
+
+    StyleProperties: {
+        /**
+         * This property determines how we scale the thumbnail image. It uses a
+         * similar vocabulary as web css' background-size property: 'cover' and
+         * 'center'. 'cover' means we will reposition and rescale the image to
+         * cover the entire thumbnail space. 'center' means we will maintain the
+         * image's original dimensions and center it in the thumbnail space.
+         */
+        'thumbnail-background-size': GObject.ParamSpec.string('thumbnail-background-size',
```
ptomato: Adding something as a CSS property means we have to be prepared for someone to do something preposterous like change it to a different value on hover. I'm thinking for this case it might make more sense to have a construct-only property even though it does feel like more of a theming thing.

js/app/modules/banner/title.js
ptomato: I'm still wondering about this too

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -170,6 +171,7 @@ const ContentGroup = new Module.Class({
             // When the spinner is not being shown on screen, set it to
             // be inactive to help with performance.
             spinner.active = spinner.visible = this._selection.loading;
+            this._error_page.visible = !!this._selection.error;
```
ptomato: This is a great solution and much better than what I was doing previously, which was setting the fonts in the error page to be really small in cases where vertical space made a difference, like in the top menu module. Now that you've improved upon that, consider removing my hacks :smile:

js/app/modules/card/video.js
```diff
@@ -0,0 +1,51 @@
...
+
+        // this.set_title_label_from_model(this._title_label);
+        this.set_label_or_hide(this._synopsis_label, this.model.synopsis);
+
+        this.show_all();
+
+        // this._title_label.visible = this.show_title;
+        let video_player = new EosKnowledgePrivate.MediaBin();
+        video_player.get_style_context().add_class(Utils.get_element_style_class(Video, 'player'));
+        video_player.set_uri('file:///home/endless/checkout/eos-knowledge-lib/YUaYs_7vGAI.mp4')
```
ptomato: Looks like this is left over from debugging

js/app/modules/card/video.js
```diff
@@ -0,0 +1,51 @@
...
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY, true),
+    },
+
+    Template: 'resource:///com/endlessm/knowledge/data/widgets/card/video.ui',
+    InternalChildren: [ 'title-label', 'synopsis-label' ],
+
+    _init: function (props={}) {
+        this.parent(props);
+
+        // this.set_title_label_from_model(this._title_label);
```
ptomato: These commented-out lines should be uncommented?

data/css/_colors.scss
```diff
@@ -1,3 +1,5 @@
+$brand-color: #005a71 !default;
```
ptomato: What's `$brand-color` meant for? It may be worth adding in a separate commit with explanation about what it should be used for.

js/app/modules/selection/contentForSet.js
```diff
@@ -0,0 +1,49 @@
...
+// Copyright 2016 Endless Mobile, Inc.
+
+const Actions = imports.app.actions;
+const Dispatcher = imports.app.dispatcher;
+const HistoryStore = imports.app.historyStore;
+const Module = imports.app.interfaces.module;
+const QueryObject = imports.search.queryObject;
+const Xapian = imports.app.modules.selection.xapian;
+
+const ContentForSet = new Module.Class({
```
ptomato: Test it with `test_selection_compliance`?

js/app/modules/contentGroup/mediaLightbox.js
```diff
@@ -66,44 +73,80 @@ const MediaLightbox = new Module.Class({
...
+        let content_type = media_object.content_type;
+        if (widget === null) {
+            if (media_object instanceof MediaObjectModel.VideoObjectModel) {
+                widget = new EosKnowledgePrivate.MediaBin();
+                widget.set_uri('file:///home/endless/checkout/eos-knowledge-lib/YUaYs_7vGAI.mp4')
+            } else if (content_type === 'application/pdf') {
+                let stream = media_object.get_content_stream();
+                widget = new PDFView.PDFView({
+                    expand: true,
+                    // FIXME: Is this gross? Any way to make it less gross?
```
ptomato: Yep, this is gross :frowning_face:

What needs to be solved? `expand` and `valign: Gtk.Align.FILL` are not working as expected?

js/app/modules/contentGroup/mediaLightbox.js
```diff
@@ -66,44 +73,80 @@ const MediaLightbox = new Module.Class({
...
+
+        let widget = this.create_submodule('card', {
             model: media_object
         });
-        this.lightbox_widget = card;
+
+        // Lightboxes may need to show a variety of different content, in which
+        // case a single card type may not fit all needs. If no card type is
+        // specified, try to determine content type and show it accordingly.
+        let content_type = media_object.content_type;
```
ptomato: I believe this is what previewer.js is meant for. Could you add the PDF and video support there instead?

ptomato:
> This may be nonsense, but could it be that the concept of a top-level set could be eliminated if you used a different pager? Like, a "single-page app" pager that interpreted the history store changing to the "article page" and actually switched to the set page?

Any thoughts about this? The top-level set change sets off alarm bells in my head but I can't quite put my finger on why.

js/app/modules/contentGroup/mediaLightbox.js
rmacqueen: Yea they don't do anything. PDF Viewer still ends up having a very small height

js/app/modules/contentGroup/mediaLightbox.js
rmacqueen: The designs call for this content to be displayed in a lightbox. previewer.js has none of the lightboxing code, so we would have to repeat it all there, not to mention having to create a separate module wrapper class, since previewer is just a widget.

js/app/historyStore.js
```diff
@@ -285,6 +285,7 @@ const HistoryStore = new Lang.Class({
                 this.set_current_item_from_props({
                     page_type: old_item.page_type,
                     model: old_item.model,
+                    context: old_item.model ? old_item.model.resouces : [],
```
ptomato: typo: resouces -> resources

js/app/modules/contentGroup/mediaLightbox.js
ptomato: I thought previewer.js was intended to be used internally in the lightbox for this purpose?

js/app/modules/contentGroup/mediaLightbox.js
ptomato: Probably worth digging into... could Watson or Juan Pablo help?

data/widgets/banner/dynamic.ui
```diff
@@ -6,26 +6,33 @@
     <property name="visible">True</property>
     <property name="can_focus">False</property>
     <child>
-      <object class="EknDynamicLogo" id="logo">
-        <property name="visible">True</property>
-        <property name="can_focus">False</property>
-        <property name="valign">end</property>
-        <property name="expand">True</property>
-        <property name="image-uri">resource:///app/assets/logo</property>
+      <object class="GtkGrid" id="grid">
```
ptomato: Since we have a custom `vfunc_draw` in EknDynamicLogo already, I'd suggest adding a `Gtk.render_background()` there instead.

js/app/interfaces/card.js
```diff
@@ -189,6 +189,15 @@ const Card = new Lang.Interface({
         'sequence': GObject.ParamSpec.uint('sequence', 'Sequence', 'Sequence',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             0, Object.keys(Sequence).length, Sequence.NONE),
+        /**
+         * Property: can-mark-read
+         * Whether to mark cards as having been read
+         *
+         */
+        'can-mark-read': GObject.ParamSpec.boolean('can-mark-read', 'Can mark read',
```
ptomato: Could we just do this unconditionally?

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -267,6 +271,10 @@ const ContentGroup = new Module.Class({
 
     },
 
+    _on_reading_history_changed: function () {
+        this._arrangement.mark_read(ReadingHistoryModel.get_default().get_read_articles());
```
ptomato: At first glance I would rather see the models being marked read, and then that change would propagate to the cards. 

Come to think of it, I would have rather seen this implemented as per-user EKN metadata which has been on our list for quite a long time, but I don't think we'll get around to it in 3.1. In any case we should do something here that's easy to refactor in that direction.

js/app/modules/contentGroup/mediaLightbox.js
rmacqueen: Don't think so. Lightbox is its own separate widget. Previewer only used in mediaCard as far as I can see

js/app/modules/contentGroup/mediaLightbox.js
rmacqueen: We do the same thing in `Card.KnowledgeDocument`

https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/modules/card/knowledgeDocument.js#L218

js/app/interfaces/card.js
rmacqueen: No, because, for example, we don't want the checkmark to show up on list cards in travel app or prensa libre app

data/widgets/banner/dynamic.ui
rmacqueen: done

js/app/modules/contentGroup/contentGroup.js
rmacqueen: Our models have always had all `CONSTRUCT_ONLY` properties - i.e. been immutable. My idea was to stick with that and have mark_read operate similar to how highlight operates

js/app/interfaces/card.js
rmacqueen: >  But what to do about sets that contain, let's say, lessons, articles, videos, and podcasts all together?

Then presumably you would choose a collection-name that encompassed all of those, e.g. "items". That seems like a Content problem rather than an engineering one

js/app/modules/card/list.js
rmacqueen: fixed

js/app/modules/contentGroup/mediaLightbox.js
ptomato: Ugh, let's keep it here as well then. But open a backlog issue and put a link in the FIXME comment so it doesn't drop off the radar

js/app/modules/contentGroup/contentGroup.js
ptomato: Indeed, it was my point that this is the first property of a model that cannot be considered immutable. We've worked around it so far with the ReadingHistoryModel, but now that it needs to influence the UI instead of just the queries, it seems like it should be made more model-like.

Since a real per-user EKN metadata store is probably not realistic, how about having each card check the default ReadingHistoryModel? As far as I can see, that would make it quite easy to refactor in the right direction later.

js/app/modules/card/list.js
```diff
@@ -30,6 +30,7 @@ const List = new Module.Class({
             true),
     },
 
+
```
ptomato: nitpick: Remove this blank line (otherwise this commit is good, so I'll remove the blank line myself and commit it)

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -115,7 +139,23 @@ const ContentGroup = new Module.Class({
...
+            visible: this.paginate,
+        });
+
+        see_more_button.get_style_context().add_class(Utils.get_element_style_class(ContentGroup, 'paginate'));
+        let content_grid = new Gtk.Grid({
+            orientation: Gtk.Orientation.VERTICAL,
+            visible: true,
+        });
+        content_grid.add(this._arrangement);
+        content_grid.add(see_more_button);
```
ptomato: I'd suggest to avoid creating and adding the button if `this.paginate` was false, since we don't ever plan to support the property being non-construct-only.

js/app/interfaces/card.js
ptomato: Discussed on hangout that we'll stick with hardcoded "item"/"items" for now, using `ngettext()`, and open a new issue for customizing this text, possibly through a YAML extension

js/app/historyStore.js
```diff
@@ -49,6 +49,13 @@ const HistoryStore = new Lang.Class({
         'current-set': GObject.ParamSpec.object('current-set', 'current-set', 'current-set',
             GObject.ParamFlags.READABLE, SetObjectModel.SetObjectModel),
         /**
+         * Property: current-subset
+         *
+         * The model for the current subset in the history store.
+         */
+        'current-subset': GObject.ParamSpec.object('current-subset', 'current-subset', 'current-subset',
```
ptomato: Is it a problem that this will always be `null` when not using the CourseHistoryStore? Maybe move the logic that sets it from the CourseHistoryStore up into here?

(Or, now that I think about it some more, maybe it would be better to add a new `current-toplevel-set` property, with a corresponding property to track it in the selection, since then `current-set` would have the same semantics in both CourseHistoryStore and the other HistoryStores?)

js/app/modules/contentGroup/mediaLightbox.js
ptomato: OK, finally grasped the difference between the previewer in Card.Media and the lightbox. Now I've investigated a bit deeper it seems like Card.Media is only ever used as the lightbox card! Could we get rid of Card.Media as a module and just put the previewer directly in the lightbox?

js/app/modules/card/list.js
```diff
@@ -100,6 +107,15 @@ const List = new Module.Class({
             height: image_height,
         });
         this._inner_content_grid.size_allocate(text_alloc);
+
+        let checkmark_alloc = new Gdk.Rectangle({
```
ptomato: You can probably skip this part if the checkmark isn't visible, for better performance.

js/app/modules/card/list.js
```diff
@@ -45,10 +46,14 @@ const List = new Module.Class({
         this.set_thumbnail_frame_from_model(this._thumbnail_frame);
         this.update_card_sizing_classes(Card.MinSize.A, Card.MinSize.D);
         this._synopsis_label.visible = this.show_synopsis;
-
+        this._title_label.vexpand = !this.show_synopsis;
         Utils.set_hand_cursor_on_widget(this);
     },
 
+    mark_read: function (is_read) {
```
ptomato: This would change given the refactors we discussed in the hangout today, but otherwise this commit looks OK to merge.

js/app/historyStore.js
rmacqueen: > Maybe move the logic that sets it from the CourseHistoryStore up into here?

The thing is then we'd be doing the engine query to get the first subset of a set for all apps, not just those that actually need it (right now only capacitate)

rmacqueen: > maybe it would be better to add a new `current-toplevel-set` property

That's what I had originally but moved to this because with `current-toplevel-set` I have to make more changes across other modules. `current-set` then has different semantics across history stores because in the case if course history store, when you click set on home page you _dont_ want to change `current-set`, but in other history stores you do

js/app/modules/contentGroup/mediaLightbox.js
rmacqueen: added

data/css/_colors.scss
rmacqueen: `$brand-color` is the color associated with the course on capacitate.org

js/app/interfaces/card.js
```diff
@@ -3,23 +3,32 @@
...
 const Module = imports.app.interfaces.module;
+const QueryObject = imports.search.queryObject;
 const SearchUtils = imports.search.utils;
 const SetMap = imports.app.setMap;
+const SetObjectModel = imports.search.setObjectModel;
 const SpaceContainer = imports.app.widgets.spaceContainer;
 const Utils = imports.app.utils;
 
+String.prototype.format = Format.format;
+let _ = Gettext.dngettext.bind(null, Config.GETTEXT_PACKAGE);
```
ptomato: Call this function `ngettext`, otherwise it will be scanned incorrectly when generating the pot files

ptomato: Also this file should be added to `po/POTFILES.in`

js/app/historyStore.js
ptomato: OK, got it.

js/app/modules/selection/subsetsForSet.js
```diff
@@ -10,13 +12,28 @@ const Xapian = imports.app.modules.selection.xapian;
 const SubsetsForSet = new Module.Class({
     Name: 'Selection.SubsetsForSet',
     Extends: Xapian.Xapian,
+    Properties: {
+        /**
+         * Property: top-level-only
```
ptomato: Documentation doesn't match code

js/app/historyStore.js
```diff
@@ -167,6 +174,21 @@ const HistoryStore = new Lang.Class({
         return item || null;
     },
 
+    get current_subset () {
+        if (this._current_subset)
+            return this._current_subset;
+        return null;
+    },
+
+    set_current_subset: function (model) {
```
ptomato: Again sorry to nitpick, but this one should be in the section marked "Common helper functions for history stores, not for use from other modules"

data/widgets/card/list.ui
```diff
@@ -73,6 +78,17 @@
             </child>
           </object>
         </child>
+        <child>
+          <object class="GtkImage" id="checkmark">
+            <property name="can_focus">False</property>
+            <property name="visible">False</property>
+            <property name="no_show_all">True</property>
+            <property name="resource">/com/endlessm/knowledge/data/images/checkmark.png</property>
```
ptomato: Sorry for not catching this before! The checkmark should be settable in CSS using `-gtk-icon-source`. Maybe even use `object-select-symbolic` for it in the default theme, and use this PNG in the course theme.

data/css/modules/contentGroup/_contentGroup.scss
```diff
@@ -94,6 +94,17 @@
             -gtk-icon-shadow: none;
         }
     }
+
+    &__paginate {
+        background-color: $brand-color;
```
ptomato: See other comment; other than this, the two paginate commits are ready to go to master.

data/widgets/card/list.ui
rmacqueen: not sure I understand. Are you suggesting to use `ThemeableImage` here?

data/css/_colors.scss
rmacqueen: I need 10 different colors to style this template. We currently only offer 7, so I have to add more. I have no preference as to what the new color variables are called - I tried to follow the names design was giving them in the specs.

data/widgets/card/list.ui
ptomato: Yes.

js/app/historyStore.js
```diff
@@ -216,6 +207,15 @@ const HistoryStore = new Lang.Class({
 
     // Common helper functions for history stores, not for use from other
     // modules...
+    set_current_subset: function (model) {
```
ptomato: I think you amended this change into the wrong commit, squash it into the one titled "Add current-subset property to HistoryStore"

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -15,6 +15,7 @@ const Dispatcher = imports.app.dispatcher;
 const HistoryStore = imports.app.historyStore;
 const Module = imports.app.interfaces.module;
 const Overflow = imports.app.modules.arrangement.overflow;
+const ReadingHistoryModel = imports.app.readingHistoryModel;
```
ptomato: Import not needed?

js/app/readingHistoryModel.js
```diff
@@ -81,8 +81,11 @@ const ReadingHistoryModel= new Knowledge.Class({
     },
 
     mark_article_read: function (article_id) {
+        if (this.is_read_article(article_id))
```
ptomato: Hmm, this looks tidy, but the return value doesn't seem all that useful outside of the context of the property setter in ContentObjectModel. As long as you're calling `is_read_article()` anyway here, I'd say to do it in the property setter instead.

js/app/modules/controller/course.js
```diff
@@ -58,4 +59,11 @@ const Course = new Module.Class({
             this._window.make_ready(cb);
         });
     },
+
+    _on_history_change: function () {
+        let item = HistoryStore.get_default().get_current_item();
+        if (item.media_model) {
```
ptomato: This only marks items read that the user has viewed in the lightbox, I think?

data/preset/course.yaml
```diff
@@ -0,0 +1,150 @@
...
+            selection:
+              type: Selection.AllSets
+              slots:
+                filter: Filter.Featured
+        - shortdef: 'Layout.Box(orientation: vertical, halign: fill, hexpand: true)'
+          styles:
+            - footer
+          slots:
+            contents:
+            - shortdef: 'ContentGroup.StaticText(lines: 1, label: "More courses soon.")'
```
ptomato: This is more of a design problem than a code review problem, but what does "More courses soon" mean? Does it really belong in the preset? I've been meaning to ask this after all three of the demos you gave...

data/css/_colors.scss
ptomato: If we need more colors then let's add them in a consistent way and not bolt them on top of what we have. I guess I need more context around this. I'm afraid this is a perfect example of what I was talking about this morning, about "top-down" design that bypasses the default theme...

Let's check with the designers whether the intention is to add a "brand color" to the default theme, and what components it should be used in, how it is different from the "accent color", etc. If it's only intended to be used for the course theme, then we should (1) push back, because I'm not happy about having a theme that only applies to one preset and a preset that can only use one theme, (2) if that fails then we should move it into the course SCSS file and prefix it with the course name, something like `course-brand-color`.

For now, I would say change the paginate button to use the accent color, then override it in the course theme to use the brand color; then we can merge the pagination changes.

Also, if you need 10 colors, where are the other two? :stuck_out_tongue_winking_eye:

rmacqueen: Wait, is your concern just that it is being added to the `_colors.scss` file (and thus the default theme) as opposed to just living in `course.scss`? Because that I can change by just swapping variable names. As I see it right now, these variables have no consistent meaning: `$primary-medium-color` doesn't have the same semantic meaning across apps - it's just one of the 7 color variables you can use for styling an app.

rmacqueen: you'll notice they specify three shades of accent color (we only offer two) and an additional brand-color which we don't have.

data/css/modules/contentGroup/_contentGroup.scss
```diff
@@ -96,7 +96,7 @@
     }
 
     &__paginate {
-        background-color: $brand-color;
+        background-color: $primary-dark-color;
```
ptomato: Ah, I think you squashed this one into the wrong commit - it should be in "Add pagination to content group". This is currently the only thing preventing me from merging the pagination commits to master, since the build breaks for lack of `$brand-color`

data/preset/course.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+- &main-arrangement-slot
+  arrangement:
+    type: Arrangement.Grid
+    properties:
+      homogeneous: true
+      max-children-per-line: 3
+      min-children-per-line: 3
+    slots:
+      card:
```
ptomato: general comment: on "leaves" of the tree you can leave out `shortdef` for compactness, like `card: 'Card.Thumb(background-size: "center")'` but that's a matter of taste so I will certainly not gripe about that

js/app/modules/controller/course.js
```diff
@@ -0,0 +1,61 @@
...
+            visible: false,
+        });
+
+        this.load_theme();
+    },
+
+    make_ready: function (cb=function () {}) {
+        // Load all sets, with which to populate the set map
+        // FIXME: deduplicate this with Selection.AllSets
+        Engine.get_default().get_objects_by_query(new QueryObject.QueryObject({
```
ptomato: This is duplicated in `Controller.Buffet` as well. I can't remember, was there a reason why we couldn't just set up  the `SetMap` when an app starts up, regardless of the controller?

ptomato:
I committed two more to master

js/app/modules/controller/course.js
rmacqueen: Well I think the issue is `window` has to have a guarantee that the SetMap is ready before it makes itself ready. So I'm not sure where else you could put this and maintain that guarantee.

ptomato: In the Controller parent class?

rmacqueen: Controller parent class? There is no controller parent class - it's just an interface, and we don't have an `interface_make_ready()` now AFAIK

ptomato:
Merged three more, this will be a slightly more difficult rebase as you'll have to squash the checkmark asset into the Course preset commit.

rmacqueen:
ready again

ptomato:
Committed the reading history patch. Now only the final commit remains. I'm not doing a hugely detailed review on that as I think the design QA is more important, so I think we should merge it by tomorrow. I do have one question and one suggestion:
- Checking `media_model` means that only items that get viewed inside a lightbox are marked read. I know you responded to that, but I don't understand yet. It seems very specific to the Capacítate app, whereas marking all items read would be what you expect as a more general history store API consumer, and shouldn't affect the behaviour of the Capacítate app.
- Suggestion would be to move as many things out of nested SCSS rules as possible in the course theme, in cases where that makes sense. For example, adding the checkmark asset to the list card seems like it should not apply only on the set page and not only to list cards, even if that's the only place in this particular preset where the checkmark would show up.

I'm ready to merge this after those two things have been addressed / answered, and let any further changes shake out in design QA.

rmacqueen:
> Checking media_model means that only items that get viewed inside a lightbox are marked read. I know you responded to that, but I don't understand yet. It seems very specific to the Capacítate app, whereas marking all items read would be what you expect as a more general history store API consumer, and shouldn't affect the behaviour of the Capacítate app.

Not sure I understand. Are you asking a) why does every item (that is not a set) get shown in a lightbox? or b) Why do only lightbox'd elements get marked as read? Because the answer to b) is that, because of a), this translates to all items getting marked as read. In the course template, any item clicked on will be shown in lightbox _and_ will be marked as read. 

> Suggestion would be to move as many things out of nested SCSS rules as possible in the course theme, in cases where that makes sense. For example, adding the checkmark asset to the list card seems like it should not apply only on the set page and not only to list cards, even if that's the only place in this particular preset where the checkmark would show up.

OK will do

ptomato:
> Are you asking a) why does every item (that is not a set) get shown in a lightbox? or b) Why do only lightbox'd elements get marked as read? Because the answer to b) is that, because of a), this translates to all items getting marked as read. In the course template, any item clicked on will be shown in lightbox and will be marked as read.

Neither one, I'm asking why not all items in all history stores are marked as read.

ptomato:
Let's do it! The content doesn't build yet, but I think we should leave any further checking to design QA and open follow-up issues for things that need to be changed.

PR #1223, opened 2016-10-21
SetObjectModel has background-uri prop

ffarfan:
In addition to providing a URI for its thumbnail image, our set objects may
require a way to provide a URI to its background image.

https://phabricator.endlessm.com/T13850

### DIFF
--- a/js/search/setObjectModel.js
+++ b/js/search/setObjectModel.js
@@ -1,6 +1,7 @@
 // Copyright 2015 Endless Mobile, Inc.
 /* exported SetObjectModel */
 
+const GObject = imports.gi.GObject;
 const Lang = imports.lang;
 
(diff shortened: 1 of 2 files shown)


rmacqueen:
This looks fine, but let's wait to see if design is gonna keep the idea of having a separate background image per set

rmacqueen:
So I talked to Mariela today. We're gonna go ahead and just have _one_ background image per app, to keep things consistent with prior templates.  So looks like we don't need this change just yet...
Sorry for your having to do unnecessary work :-) FWIW this code looks great!

ptomato:
We may still want this feature eventually, but since this code moved to C the pull request isn't really salvageable... sorry this never got merged!

PR #1225, opened 2016-10-21
control: add gstreamer plugins as build deps

mattdangerw:
Our builds are failing right now without them


ptomato:
**CR pass**

PR #1226, opened 2016-10-21
control: add build dep on gstreamer1.0-tools

mattdangerw:
So we can check for required gstreamer elements


ptomato:
**CR pass**

PR #1228, opened 2016-10-24
EknMediaBin: implemeted EOS design.

xjuan:
Implemented eos design using CSS and added named icons for playback actions

https://phabricator.endlessm.com/T13360

### DIFF
--- a/lib/eosknowledgeprivate/EknMediaBin.ui
+++ b/lib/eosknowledgeprivate/EknMediaBin.ui
@@ -5,7 +5,8 @@
   <object class="GtkImage" id="fullscreen_image">
     <property name="visible">True</property>
     <property name="can_focus">False</property>
-    <property name="icon_name">view-fullscreen</property>
+    <property name="pixel_size">18</property>
+    <property name="icon_name">ekn-media-bin-fullscreen</property>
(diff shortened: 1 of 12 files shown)


xjuan:
Here is a screenshot of the current implementation.
It is not pixel perfect compared with the design mockups since I tried to use em units in the CSS as much as possible.

screenshot from 2016-10-24 17-00-47

tchx84:
Looks awesome! just a few comments:

Design wise: in the specs it looks like the upper bar (where the title is displayed) should have the same height as the bottom bar (where the controls are). Maybe the upper bar should have have less top/bottom margins?

Code wide: I see A LOT of these messages, but I don't see them on the master version.

```
(lt-eos-player:14102): Gtk-WARNING **: Negative content width -2 (allocation 0, extents 1x1) while allocating gadget (node highlight, owner GtkScale)
```

Usability: now that the slider is "invisible" is a little bit harder to play forward/rewind the videos, the behavior is a bit unpredictable, when compared to the version in the master,

xjuan:
Fixed all issues and added some extra polish!

rmacqueen:
CR PASS

PR #1232, opened 2016-10-28
T13678rory

rmacqueen:
https://phabricator.endlessm.com/T13678

### DIFF
--- a/data/css/modules/card/_list.scss
+++ b/data/css/modules/card/_list.scss
@@ -30,6 +30,10 @@
         font-size: 14px;
     }
 
+    &__content_grid {
+        padding: 20px;
+    }
+
     &:hover &__title {
         color: $primary-light-color;
     }
@@ -51,9 +55,8 @@
         font-size: 35px;
     }
 
-    &.width-e &__contentFrame,
-    &.width-f &__contentFrame {
-        padding: 20px;
+    &.width-h &__title {
+        font-size: 35px;
     }
 
     .composite &.width-d &__title,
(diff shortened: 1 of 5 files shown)


js/app/modules/card/list.js
```diff
@@ -43,61 +43,70 @@ const List = new Module.Class({
...
+        let text_width = total_content_width -  image_width;
+
+        let card_alloc = new Gdk.Rectangle({
+            x: alloc.x,
+            y: alloc.y,
+            width: alloc.width,
+            height: alloc.height,
+        });
+
+        this.parent(card_alloc);
```
ptomato: You could just call `this.parent(alloc)` and skip creating `card_alloc`.

data/css/modules/card/_list.scss
```diff
@@ -30,6 +30,14 @@
         font-size: 14px;
     }
 
+    &__content_grid {
+        padding: 20px;
+    }
+
+    &__thumbnail {
+        min-height: 200px;
```
tchx84: I noticed a side-effect of having this value here, when List card is used as navigation card it can't shrink?

screenshot from 2016-11-02 11-15-44

tchx84:
Aside from the philip's comment, there is just one comment regarding the min-height of the thumbnail.

data/css/modules/card/_list.scss
rmacqueen: what are you seeing on master for this?

rmacqueen:
Moved to https://github.com/endlessm/eos-knowledge-lib/pull/1239

PR #1234, opened 2016-10-31
T13981

mattdangerw:
https://phabricator.endlessm.com/T13981


ptomato:
`make distcheck` fails with "No rule to make target 'autobahn'" so I think there's somewhere you are still referring to the old location in the Makefile.

mattdangerw:
@ptomato fixed!

ptomato:
**CR pass**

Note there's still a build failure due to the video player icons not being in EXTRA_DIST but otherwise it works

PR #1238, opened 2016-11-03
Add emeus to the jhbuild moduleset

ebassi:
https://phabricator.endlessm.com/T14015

### DIFF
--- a/eos-knowledge-lib.modules
+++ b/eos-knowledge-lib.modules
@@ -241,6 +241,13 @@ use the latest stable release branch from GNOME if you prefer that.
     </dependencies>
   </autotools>
 
+  <autotools id="emeus">
+    <branch repo="github" module="ebassi/emeus"/>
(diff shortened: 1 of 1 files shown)


eos-knowledge-lib.modules
```diff
@@ -241,6 +241,13 @@ use the latest stable release branch from GNOME if you prefer that.
     </dependencies>
   </autotools>
 
+  <autotools id="emeus">
+    <branch repo="github" module="ebassi/emeus"/>
+    <dependencies>
+      <dep package="gtk+"/>
```
tchx84: @ebassi where should meson and ninja deps go?

ebassi: Both meson and ninja should be part of the development tools needed to use jhbuild. The modules file does not specify a dependency on autoconf/automake/make either.

PR #1239, opened 2016-11-03
Rewrite list card to not be so jittery

rmacqueen:
Previously we were doing all sorts of craziness
in the size allocate routine. This cleans that
up and adds a width_for_height method too.

https://phabricator.endlessm.com/T13216

### DIFF
--- a/data/css/buffet.scss
+++ b/data/css/buffet.scss
@@ -104,6 +104,7 @@ $title-alt-font: 'Lato Thin' !default;
 }
 
 .search-results {
+    margin-top: 20px;
     background-color: $background-light-color;
 
     .search-page-suggestions .Card__contentFrame {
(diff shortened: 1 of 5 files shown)


tchx84:
@rmacqueen Like we talked on slack, min-height: 200px, is kind of big for small screens. Just try reducing the size the window, and you will see.

data/css/buffet.scss
```diff
@@ -119,6 +120,15 @@ $title-alt-font: 'Lato Thin' !default;
         background-color: $separator-color;
     }
 
+    .CardList {
+        &__thumbnail {
+            min-height: 200px;
+        }
+        .composite .CardList__thumbnail {
```
rmacqueen: Why? Isn't this traditional CSS, of which SCSS is a subset?

tchx84: Is not that. is about how scss gets translated/expanded. If you don't use the ```&``` there you will get:

```  .search-results .CardList .composite .CardList__thumbnail ```

which doesn't work (and not what we want either), vs:

```  .composite .search-results .CardList .CardList__thumbnail ```

that you would get by using ```&```.

rmacqueen: fixed!

tchx84:
**CR PASS**

PR #1240, opened 2016-11-08
Use GLib idle_add instead of Mainloop in async task

rmacqueen:
Mainloop.idle_add always adds its callbacks with DEFAULT
priority. In the capacitate app, this has resulted
in cases where the callback is not invoked if many of
them are queued at once. Instead we should use GLib.idle_add,
which allows us to specify a higher priority.

https://phabricator.endlessm.com/T14072

### DIFF
--- a/js/search/asyncTask.js
+++ b/js/search/asyncTask.js
@@ -2,7 +2,6 @@ const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const GObject = imports.gi.GObject;
 const Lang = imports.lang;
-const Mainloop = imports.mainloop;
 
 /**
  * Class: AsyncTask
@@ -26,12 +25,13 @@ const AsyncTask = Lang.Class({
     GTypeName: 'EknAsyncTask',
     Extends: GObject.Object,
 
-    _init: function (source, cancellable, callback) {
+    _init: function (source, cancellable, callback, priority=GLib.PRIORITY_DEFAULT) {
         this.parent();
 
         this._source = source;
         this._callback = callback;
         this._done = false;
+        this._priority = priority;
 
         if (cancellable) {
             let handle_cancel = () => {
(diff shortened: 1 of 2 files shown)


ptomato:
Looks like it will work fine in your case, but what effects will it have on other code using Async.Task?

I would prefer to add an argument to `_init()` such as `priority=GLib.PRIORITY_DEFAULT` so that you can determine the priority when you construct the task.

ptomato:
You can remove the `Mainloop` import now.

rmacqueen:
@ptomato come to think of it, why are we even having to call this function on the idle thread? Why can't we just call it directly?

js/search/asyncTask.js
```diff
@@ -26,6 +25,20 @@ const AsyncTask = Lang.Class({
     GTypeName: 'EknAsyncTask',
     Extends: GObject.Object,
 
+    Properties: {
```
ptomato: Since the other stuff (`source, cancellable, callback`) is all passed directly to the constructor instead of through a properties dict, I think it would be best to do the same with priority. You could provide a default value for the argument. In any case it won't work as-is, because the `this.parent()` call has to get a properties dict for this property to end up set.

That way you wouldn't have to specify a range for the priority either, which is more in line with what GLib does — you can set any integer as the priority.

I would also make the default value `GLib.PRIORITY_DEFAULT`, I don't think it would be good for all tasks to always run on high priority.

js/search/xapianBridge.js
```diff
@@ -118,7 +118,7 @@ const XapianBridge = new Lang.Class({
     // Queues a SoupMessage for *uri* to the current http session. Calls
     // *callback* on any errors encountered and the parsed JSON.
     _send_json_ld_request: function (uri, cancellable, callback) {
-        let task = new AsyncTask.AsyncTask(this, cancellable, callback);
+        let task = new AsyncTask.AsyncTask(this, cancellable, callback, GLib.PRIORITY_HIGH);
```
ptomato: Could you please split this into a separate commit than the feature itself?

PR #1245, opened 2016-11-11
T14121

tchx84:
https://phabricator.endlessm.com/T14121


js/app/widgets/dynamicLogo.js
```diff
@@ -91,7 +91,7 @@ const DynamicLogo = new Knowledge.Class({
         this._layout = null;
         this._image = null;
 
-        if (this._mode !== 'text') {
+        if (this.image_uri) {
```
rmacqueen: `Banner.Dynamic` sets the `image_uri` in its .ui template file so even if your mode is text only, the `image_uri` property will still be non-null and thus will pass this test

rmacqueen:
CR PASS

PR #1246, opened 2016-11-15
debian/control: bump eos-shard version

wgreenberg:
We need 0.1.2 for records_foreach


rmacqueen:
Is version `0.1.2` of shard even published? I did an apt update and tried installing it and it said I was on latest version, but knowledge-lib still complaining about not having `0.1.2`

wgreenberg:
@rmacqueen nope, currently blocked on an OBS build issue https://phabricator.endlessm.com/T14181

rmacqueen:
Looks like it's in now. works for me

PR #1252, opened 2016-11-18
ka-customizer: escape spaces in file paths

rmacqueen:
file paths can have spaces in them. we need to
escape them in order for the arg parser to work.

https://phabricator.endlessm.com/T14250

### DIFF
--- a/tools/ka-customizer
+++ b/tools/ka-customizer
@@ -223,11 +223,14 @@ done_button.connect('clicked', () => {
         show_message('Updating app and libraries, this may take a while...');
         args.push('update');
     }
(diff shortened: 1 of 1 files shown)


mattdangerw:
I think this is actually a problem with the runner script. Gio.subprocess will quote things properly. See
https://github.com/endlessm/eos-knowledge-lib/commit/9bf7e8a6ed4c64b29d8edabed9162f03e707bf32

mattdangerw:
Closing in favor of https://github.com/endlessm/eos-knowledge-lib/pull/1257

PR #1253, opened 2016-11-18
EknMediaBin: handle end of stream message.

xjuan:
https://phabricator.endlessm.com/T14196


rmacqueen:
CR PASS, but this introduces another bug, for which I opened https://phabricator.endlessm.com/T14316

PR #1254, opened 2016-11-18
Initialize Engine with an app-id

rmacqueen:
Engine needs an app-id to load the shards in its
constructor. So let's pass the app-id to the
get_default() function when we first call it.

https://phabricator.endlessm.com/T14073


rmacqueen:
@xjuan FYI

PR #1259, opened 2016-11-23
EknVfs: added support for multiples domains

xjuan:
Added domain property to use as the default domain for ekn uri searches.
Added append-shards action signal to add a list of shards for a particular domain.
Removed obsolete shards property

https://phabricator.endlessm.com/T14073

### DIFF
--- a/js/search/engine.js
+++ b/js/search/engine.js
@@ -1,6 +1,5 @@
 // Copyright 2014 Endless Mobile, Inc.
 
-const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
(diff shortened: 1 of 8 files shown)


js/search/engine.js
```diff
@@ -249,6 +232,18 @@ let get_default = function () {
         the_engine = new Engine({
             language: language,
         });
+
+        Object.defineProperty(the_engine, 'default_app_id', {
+            set: function (id) {
```
rmacqueen: I don't think this needs to be dynamic. I think this can just be a setter on the engine class. And if we don't want it to be called during the constructor then just don't include ` GObject.ParamFlags.CONSTRUCT` in the property definition

rmacqueen:
CR PASS

PR #1260, opened 2016-11-23
Make moltres work with subsets

rmacqueen:
Previously moltres would be able to return subsets
since it had no notion of the hierarchy between sets.
This commit enables that, by creating all the sets
once up front during init() rather than creating
them dynamically at runtime.

https://phabricator.endlessm.com/T14328


tchx84:
Looks good, just one issue:

I tested with news preset and noticed that we should add the corresponding tags to the sets and their articles, otherwise articles won't show in news preset (because of Filter.CurrentSet).

rmacqueen:
took way longer than I expected, but fixed

PR #1261, opened 2016-11-25
T13535

rmacqueen:

### DIFF
--- a/js/app/interfaces/filter.js
+++ b/js/app/interfaces/filter.js
@@ -48,6 +48,14 @@ const Filter = new Lang.Interface({
         'filter-changed': {},
     },
 
+    Slots: {
+        'sub-filter': {},
+    },
+
+    _interface_init: function () {
+        this._sub_filter = this.create_submodule('sub-filter');
+    },
+
     /**
      * Method: include
      * Determine whether an arrangement should show a model as a card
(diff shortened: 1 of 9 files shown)


js/app/modules/filter/multi.js
```diff
@@ -0,0 +1,38 @@
...
+
+/* exported Multi */
+
+const GObject = imports.gi.GObject;
+
+const Filter = imports.app.interfaces.filter;
+const HistoryStore = imports.app.historyStore;
+const Module = imports.app.interfaces.module;
+
+const Multi = new Module.Class({
```
ptomato: drive-by comment, I would much rather see this implemented the same way as we do with Order: https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/interfaces/order.js#L46

rmacqueen:
ready again

js/app/interfaces/filter.js
```diff
@@ -60,6 +68,8 @@ const Filter = new Lang.Interface({
      */
     include: function (model) {
         let retval = this.include_impl(model);
+        if (retval && this._sub_filter)
```
ptomato: I don't think this is correct, because the first filter's `invert` now applies to the sub-filter. E.g. if you have `Filter.Featured(invert: true)` with `Filter.Unread(invert: false)` as the subfilter, expressing "not featured and unread", then this will also include articles that are featured and read, or not featured and read.

ptomato: It would be good to have tests on the `Filter` interface that would catch this.

rmacqueen:
ready again

PR #1262, opened 2016-11-28
debian: add eosknowledgecontent packages

mattdangerw:


debian/libeos-knowledge-content-0-0.install
```diff
@@ -0,0 +1,2 @@
+usr/lib/libeos-knowledge-content-0.so.*
+usr/lib/gio/modules/libeknvfs.so
```
mattdangerw: @xjuan I didn't see the vfs library getting pulled into any existing packages, so I'm adding it here. Which I think is the right place for it anyway.

One thing I wasn't sure about was the gio-querymodules call here https://github.com/endlessm/eos-knowledge-lib/blob/286d25d82d7f1e35b2c7d42be0c359cd047ba1a3/Makefile.am#L302. If that's necessary it probably also needs to go in an post-install debian hook. Could you investigate and tag on a commit here in necessary?

xjuan: @mattdangerw yes we do need to call gio-querymodules in a post install hook
Its needed to update the cache of extension points, that way gio does not have to open the module each time to know which extension point it implements

mattdangerw:
@xjuan Postinst added. I'm not sure if this will actually work as intended though. The gio module (on master and this branch) will get installed to `/usr/lib/gio/modules` but every existing module is in `/usr/lib/x86_64-linux-gnu/gio/modules` depending on the arch.

Does anyone know if ignoring the arch specific libdir is okay in this case? Or what would be needed to actually support multi-arch in knowledge lib like all those other packages installing gio modules do?

debian/control
```diff
@@ -105,3 +106,51 @@ Depends: eos-knowledge-0 (= ${binary:Version}),
...
+ Package config and headers for the Endless Knowledge Content library
+
+Package: libeos-knowledge-content-0-doc
+Section: non-free/doc
+Architecture: any
+Depends: ${misc:Depends}
+Description: Endless Knowledge Content library documentation
+ HTML documentation for the Endless Knowledge Content library
+
+Package: gir1.2-eos-knowledge-content-0
```
ptomato: This should probably be `gir1.2-eosknowledgecontent-0` to fit with other upstream names, like `gir1.2-gdkpixbuf-2.0` (I know we named `gir1.2-eos-shard-0` incorrectly)

debian/rules
```diff
@@ -18,7 +18,7 @@ override_dh_install:
 
 # Build our documentation as well
 override_dh_auto_configure:
-	dh_auto_configure -- --enable-js-doc
+	dh_auto_configure -- --enable-js-doc --enable-gtk-doc
```
ptomato: It's a bit sad that we have to bring gtk-doc in here only to take it out later, but this does indeed seem like the best thing to do.

debian/control
```diff
@@ -105,3 +106,51 @@ Depends: eos-knowledge-0 (= ${binary:Version}),
...
+Architecture: any
+Depends: ${misc:Depends},
+         libeos-knowledge-content-0-0 (= ${binary:Version})
+Description: Endless Knowledge Content library debugging symbols
+ Debugging symbols for the Endless Knowledge Content library
+
+Package: libeos-knowledge-content-0-dev
+Section: non-free/libdevel
+Architecture: any
+Depends: ${shlibs:Depends},
```
ptomato: I don't think the `-dev` package needs `${shlibs:Depends}`, or does it?

debian/control
```diff
@@ -105,3 +106,51 @@ Depends: eos-knowledge-0 (= ${binary:Version}),
...
+         ${misc:Depends},
+         libglib2.0-dev,
+         libeos-knowledge-content-0-0 (= ${binary:Version})
+Description: Endless Knowledge Content development files
+ Package config and headers for the Endless Knowledge Content library
+
+Package: libeos-knowledge-content-0-doc
+Section: non-free/doc
+Architecture: any
+Depends: ${misc:Depends}
```
ptomato: You probably want `Suggests: devhelp` or some such thing here

ptomato:
> every existing module is in `/usr/lib/x86_64-linux-gnu/gio/modules` depending on the arch.

Sounds like we should be doing that as well, how does it work in those modules' Debian packaging?

ptomato:
I just found out we have a distro patch (taken from Debian) that enables the GIO modules path without the architecture as well. Probably for maximum compatibility it would be good to install the module in the right place though.

mattdangerw:
Added multi arch support. Confirmed that everything is building, though I haven't actually tested the debians yet.

xjuan:
CR pass

PR #1263, opened 2016-11-28
T13418

mattdangerw:


ekncontent/Makefile.am
```diff
@@ -83,7 +83,11 @@ INTROSPECTION_SCANNER_ARGS = --add-include-path=$(srcdir) --warn-all $(EOS_KNOWL
 INTROSPECTION_COMPILER_ARGS = --includedir=$(srcdir)
 
 EosKnowledgeSearchPrivate-1.0.gir: libeosknowledgesearchprivate.la
-EosKnowledgeSearchPrivate_1_0_gir_INCLUDES = Gio-2.0
+EosKnowledgeSearchPrivate_1_0_gir_INCLUDES = \
+	EosShard-0 \
+	Gio-2.0 \
+	GLib-2.0 \
+	GObject-2.0
```
ptomato: nitpick: backslash and `$(NULL)`

ekncontent/Makefile.am
```diff
@@ -19,6 +23,25 @@ lib_LTLIBRARIES =
...
 
+# Make sure to run Gtk-doc tests and build the documentation when doing
+# 'make distcheck'
+AM_DISTCHECK_CONFIGURE_FLAGS = --enable-gtk-doc
+if EOS_ENABLE_COVERAGE
+AM_DISTCHECK_CONFIGURE_FLAGS += --enable-coverage --with-coverage-dir=@EOS_COVERAGE_DIR@
+endif EOS_ENABLE_COVERAGE
+
+# Make sure that 'make dist' includes documentation
+if CAN_MAKE_DIST
```
ptomato: I think gtk-doc.m4 does this automatically, we only had to add this for `--enable-gir-doc` and `--enable-js-doc` in the parent project.

ekncontent/docs/reference/ekncontent/Makefile.am
```diff
@@ -0,0 +1,91 @@
...
+# e.g. EXTRA_DIST += version.xml.in
+EXTRA_DIST +=
+
+# Files not to distribute
+# for --rebuild-types in $(SCAN_OPTIONS), e.g. $(DOC_MODULE).types
+# for --rebuild-sections in $(SCAN_OPTIONS) e.g. $(DOC_MODULE)-sections.txt
+#DISTCLEANFILES +=
+
+if ENABLE_GTK_DOC
+TESTS_ENVIRONMENT = \
```
ptomato: Use `AM_TESTS_ENVIRONMENT` nowadays

mattdangerw:
> Since you're making a new configure.ac anyway, we may as well. If you agree, then I'll do another review with that in mind.

Sounds great!

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+AC_SUBST(EKN_CONTENT_API_NAME)
+EKN_CONTENT_INTERFACE_AGE=_EKN_CONTENT_INTERFACE_AGE_MACRO
+AC_SUBST([EKN_CONTENT_MAJOR_VERSION], [_EKN_CONTENT_API_VERSION_MACRO])
+AC_SUBST([EKN_CONTENT_MINOR_VERSION], [_EKN_CONTENT_MINOR_VERSION_MACRO])
+AC_SUBST([EKN_CONTENT_MICRO_VERSION], [_EKN_CONTENT_MICRO_VERSION_MACRO])
+AC_DEFINE([EKN_CONTENT_VERSION], [_EKN_CONTENT_VERSION_MACRO], [Full version number])
+
+# Shared library versioning; calculated automatically.
+# See http://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
+EKN_CONTENT_LT_CURRENT=m4_eval(100 * _EKN_CONTENT_MINOR_VERSION_MACRO + _EKN_CONTENT_MICRO_VERSION_MACRO - _EKN_CONTENT_INTERFACE_AGE_MACRO)
```
ptomato: I'd go with https://developer.gnome.org/programming-guidelines/stable/versioning.html.en#libtool-versioning instead of this calculation

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+EKN_CONTENT_LT_VERSION_INFO="$EKN_CONTENT_LT_CURRENT:$EKN_CONTENT_LT_REVISION:$EKN_CONTENT_LT_AGE"
+AC_SUBST(EKN_CONTENT_LT_VERSION_INFO)
+
+# Required build tools
+# --------------------
+# Make sure we can create directory hierarchies
+AC_PROG_MKDIR_P
+# Make sure we can symlink
+AC_PROG_LN_S
+AS_IF([test "$LN_S" = "ln -s"], [], [AC_MSG_ERROR([Symlink support required.])])
```
ptomato: Is this really required?

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+# C compiler
+AC_PROG_CC
+AC_PROG_CC_C99
+# Make sure the C compiler supports per-target CFLAGS
+AM_PROG_CC_C_O
```
ptomato: This is included in `AC_PROG_CC` since Automake 1.14

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+# Library configuration tool
+PKG_PROG_PKG_CONFIG
+# Gtk-doc; 0.18 required for Markdown parsing
+GTK_DOC_CHECK([1.18], [--flavour no-tmpl])
```
ptomato: The second argument is completely ignored these days

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+Try installing "glib" package.])])
+
+AC_CACHE_SAVE
+
+# Check that the compiler really supports C99
+AS_IF([test "x$ac_cv_prog_cc_c99" = "xno"], [AC_MSG_ERROR([C99 is required.])])
+
+# Configure options
+# -----------------
+# --enable-strict-flags: Compile with strict compiler flags. Done automatically
```
ptomato: I would get rid of this whole option and just use the default settings of [`AX_COMPILER_FLAGS`](https://www.gnu.org/software/autoconf-archive/ax_compiler_flags.html)

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+AC_ARG_VAR([JASMINE_JUNIT_REPORTS_DIR], [Where to put test reports])
+AS_IF([test -n "$JASMINE_JUNIT_REPORTS_DIR"],
+    [JASMINE_REPORT_ARGUMENT="--junit $JASMINE_JUNIT_REPORTS_DIR/\$\${log/%.log/.js.xml}"
+    AC_MSG_RESULT([in $JASMINE_JUNIT_REPORTS_DIR])],
+    [JASMINE_REPORT_ARGUMENT=
+    AC_MSG_RESULT([nowhere])])
+AC_SUBST([JASMINE_REPORT_ARGUMENT])
+
+# ------------------
+# These go into the pkg-config file as Requires: and Requires.private:
```
ptomato: Use [`AX_PKG_CHECK_MODULES`](https://www.gnu.org/software/autoconf-archive/ax_pkg_check_modules.html) instead of this section. (and I think only `eos-shard-0` can be in requires.private)

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+
+AC_CACHE_SAVE
+
+# Output
+# ------
+dnl List files here that the configure script should output
+AC_CONFIG_FILES([
+    Makefile
+    ekncontent/eknc-version.h
+    $EKN_CONTENT_API_NAME.pc
```
ptomato: Use colons here, to take the API level out of the `.pc.in`'s filename

ekncontent/configure.ac
```diff
@@ -0,0 +1,196 @@
...
+
+PKG_CHECK_MODULES([EKN_VFS], [
+    eos-shard-0
+    gio-2.0
+    gmodule-2.0
+])
+
+# Code Coverage
+# -------------
+EOS_COVERAGE_REPORT([c])
```
ptomato: I would go ahead and switch to [`AX_CODE_COVERAGE`](https://www.gnu.org/software/autoconf-archive/ax_code_coverage.html) entirely here

ekncontent/Makefile.am
```diff
@@ -0,0 +1,97 @@
...
+	$(NULL)
+libeknvfs_la_CFLAGS = $(AM_CFLAGS)
+libeknvfs_la_LIBADD  = @EKN_VFS_LIBS@
+libeknvfs_la_LDFLAGS = \
+	-export-symbols-regex "^g_io_module_(load|unload|query)" \
+	-avoid-version \
+	$(NULL)
+
+uninstall-hook:
+	if test -z "$(DESTDIR)"; then \
```
mattdangerw: I'll update the hooks to properly include DESTDIR when calling gio-querymodules below. In that case it should be harmless. The cache file it generates just won't be included in any of the debian packages, and we'll used the postinst to update the real cache file.

ekncontent/ekncontent/eknc-subtree-dispatcher.c
```diff
@@ -191,12 +202,20 @@ const GDBusSubtreeVTable subtree_vtable = {
...
+ * eknc_subtree_dispatcher_register:
+ * @dispatcher: the subtree dispatcher
+ * @connection: the dbus connection
+ * @subtree_path: a subtree path to register with
+ *
+ * Register the dispatcher.
+ */
 void
-ekns_subtree_dispatcher_register (EknsSubtreeDispatcher *dispatcher,
+eknc_subtree_dispatcher_register (EkncSubtreeDispatcher *dispatcher,
```
xjuan: NIT, @ptomato should we use *self here instead of dispatcher

BTW I personally prefer dispatcher but it is my understanding that all our libs should use 'self'

ekncontent/ekncontent/eknc-subtree-dispatcher.h
```diff
@@ -0,0 +1,43 @@
...
+
+/* Copyright 2015 Endless Mobile, Inc. */
+
+#ifndef EKNC_SUBTREE_DISPATCHER_H
+#define EKNC_SUBTREE_DISPATCHER_H
+
+#include <gio/gio.h>
+
+#define EKNC_TYPE_SUBTREE_DISPATCHER               (eknc_subtree_dispatcher_get_type ())
+#define EKNC_SUBTREE_DISPATCHER(obj)                               (G_TYPE_CHECK_INSTANCE_CAST ((obj), EKNC_TYPE_SUBTREE_DISPATCHER, EkncSubtreeDispatcher))
```
xjuan: you could use G_DECLARE_FINAL_TYPE() macro here

ekncontent/ekncontent/eknc-subtree-dispatcher.c
ptomato: I don't think it's an absolute rule but much of the other code uses it, so I usually flag it in reviews if I notice it

xjuan:
CR pass

PR #1264, opened 2016-11-29
contentObjectModel: don't depend on reading history store

mattdangerw:
Having search libraries depend on app code is incorrect and with
the impending port to C about to become impossible.

The implementation to cover the existing functionality is not very
efficient, but I doubt it will be the bottleneck in terms of
performance.

### DIFF
--- a/js/app/historyStore.js
+++ b/js/app/historyStore.js
@@ -13,6 +13,7 @@ const Dispatcher = imports.app.dispatcher;
 const HistoryItem = imports.app.historyItem;
 const MediaObjectModel = imports.search.mediaObjectModel;
 const Pages = imports.app.pages;
+const ReadingHistoryModel = imports.app.readingHistoryModel;
 const SetObjectModel = imports.search.setObjectModel;
 const Utils = imports.app.utils;
 
@@ -100,7 +101,7 @@ const HistoryStore = new Lang.Class({
                     this.go_forward();
                     break;
                 case Actions.ITEM_CLICKED:
-                    payload.model.mark_read();
+                    ReadingHistoryModel.get_default().mark_article_read(payload.model.ekn_id);
                     break;
             }
         });
--- a/js/app/modules/card/list.js
+++ b/js/app/modules/card/list.js
@@ -7,6 +7,7 @@ const Gtk = imports.gi.Gtk;
 const Card = imports.app.interfaces.card;
 const NavigationCard = imports.app.interfaces.navigationCard;
 const Module = imports.app.interfaces.module;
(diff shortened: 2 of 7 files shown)


tests/compliance.js
```diff
@@ -19,87 +19,93 @@ const WidgetDescendantMatcher = imports.tests.WidgetDescendantMatcher;
 const Utils = imports.tests.utils;
 
 function test_card_compliance(CardClass) {
-    describe(CardClass.$gtype.name + ' implements Card correctly', function () {
+    describe(CardClass.$gtype.name + ' compliance', function () {
         beforeEach(function () {
```
ptomato: Can be moved to the inner `beforeEach`?

ptomato:
As discussed in the standup, this is necessary because we overlooked the introduced dependency on js/app, but it's worth looking at this while considering how we would implement the per-user EKN metadata store.

If the line for deciding what goes in ekncontent is "it should work in Electron prototypes as well", then we definitely need the per-user metadata to be on the ekncontent side of the line. Probably the ReadingHistoryModel in that case too, although that would be a larger refactor probably not in scope for this pull request.

mattdangerw:
Agreed. The problem the reading history store is it mandates not just a gapplication but an eos-sdk application. It uses `Gio.Application.get_default().config_dir.get_path()` when `config-dir` is an eosapplication prop. It's quite literally "app" code, and I don't think belongs in ekncontent without a bigger refactor.

> If the line for deciding what goes in ekncontent is "it should work in Electron prototypes as well"

I would encourage us to start thinking more about our "knowledge platform" in which case the dividing line should not single out Electron. Something more like "is it useful for a general purpose library for serving up soma content to applications." The goal IMO would be to put out independently useful parts that can be combined in surprising ways. An electron app, a native gtk app built with glade and adwaita, a qt app, a terminal interface :), all valid.

Per-user EKN metadata is totally still valid in that case, but to have it in ekncontent it should not be per application or mandating any specific tech stack. Probably with some thought into how it could be used from apps that don't "own" the  content. Then you can use this for an electron prototype, or a teacher tool that monitored reading habits across all content, all sorts of craziness we should not try to predict.

Mainly a philosophical point, but that's probably good to clear up while shaping this library. `searchProvider` is actually something I'm hoping to ditch from this library as well and move straight into eos-knowledge-services. Global search integration should be an use of the library that serves up content, but not part of it.

mattdangerw:
High level aside, I'm just in a race against Friday and looking for a quick way to detangle and port to C. And then land lots of follow ups improving the library api :)

mattdangerw:
Fixed up the small comment.

xjuan:
CR pass

PR #1267, opened 2016-12-01
T14257

katelk28:
**Phabricator Issue:**
https://phabricator.endlessm.com/T14257

Yay my first PR! Wooooohoooo 
o<|:~)--/--<

### DIFF
--- a/data/css/modules/card/_media.scss
+++ b/data/css/modules/card/_media.scss
@@ -1,35 +1,30 @@
 .CardMedia {
-    padding: 10px;
     background-color: white;
 
-    separator {
-        background-color: $black-20-color;
-        min-height: 1px;
-    }
-
     &__caption {
-        font-size: .81em;  /* 13pt * 1/12 * 0.75 DPI correction factor */
-        font-family: $display-font;
+        font-size: 20px;
+        font-family: $body-font;
         color: #666666;
+        padding: 20px 20px 0px 20px;
 
         .composite & {
-            font-size: 14.1px;  /* 12px spec * 88% because spec was at 110% scale not 125% */
+            font-size: 20px;
         }
     }
 
     &__attributionText {
         font-weight: bold;
-        font-size: .62em;  /* 10pt * 1/12 * 0.75 DPI correction factor */
+        font-size: 16px;
         font-family: $display-font;
         color: $accent-light-color;
-        transition: color 250ms ease-in-out;
+        padding: 10px 20px 20px 20px;
 
         &:hover {
             color: $accent-dark-color;
         }
 
         .composite & {
-            font-size: 10.5px;  /* 9px spec * 88% */
+            font-size: 20px;
         }
     }
 }
--- a/data/css/modules/layout/_sidebar.scss
+++ b/data/css/modules/layout/_sidebar.scss
@@ -6,57 +6,48 @@
 
         .CardTitle {
             margin: 0px;
-            padding: 6px;
-            border: 2px solid transparentize(white, 1);
-            background-color: transparentize(white, 1);
-            transition: background-color 250ms ease-in-out,
-                        background-image 250ms ease-in-out,
-                        border 250ms ease-in-out;
+            padding: 10px;
 
             &__title {
                 padding: 0em 0em 0em 1.5em;
                 font-family: $display-font;
+                font-weight: 700;
                 font-size: 1.125em; /* 18pt * 1/12 * 0.75 DPI correction factor */
-                font-weight: normal;
                 color: $primary-light-color;
-                transition: color 250ms ease-in-out;
             }
 
             &:hover {
-                border: 2px solid white;
-                background-color: transparentize(white, 1 - 0.9);
-                background-image: linear-gradient(-179deg,
-                                                  transparentize(white, 1 - 0.05) 0%,
-                                                  transparentize(black, 1 - 0.05) 100%);
 
                 .Card__title {
                     color: $primary-medium-color;
                 }
             }
 
+            &:active {
+                background: transparentize(black, 1 - 0.3);
+            }
+
             &.highlighted {
-                background: transparentize(white, 1 - 0.8);
-                text-shadow: 0px 1px 0px white;
-                font-weight: bold;
-                transition-property: none;
+                background: transparentize(black, 1 - 0.5);
+                font-weight: 700;
 
                 .Card__title {
-                    color: $primary-dark-color;
-                    transition-property: none;
+                    color: white;
+                    transition-property: color;
+                    transition-timing-function: ease;
+                    transition-duration: 0.5s;
+
                 }
 
                 &:hover {
-                    border: 2px solid white;
-                    background-color: transparentize(white, 1 - 0.9);
-                    background-image: linear-gradient(-179deg,
-                                                      transparentize(white, 1 - 0.05) 0%,
-                                                      transparentize(black, 1 - 0.05) 100%);
-
-                    .Card__title {
-                        color: $primary-dark-color;
-                        transition-property: none;
-                    }
+                    background-color: transparentize(black, 1 - 0.9);
                 }
+
+                &:active {
+                background: transparentize(black, 1 - 1);
+                box-shadow: none;
+                transition: none;
+            }
             }
         }
     }
--- a/data/templates/css/wikihow.scss
+++ b/data/templates/css/wikihow.scss
@@ -4,14 +4,13 @@
 
 @include global-reset;
 
-//
+//////////////////////////////
 // Variables
-//
+//////////////////////////////
 
-$font1: "Lato", sans-serif;
-$font1-heading: "Lato";
-$font1-composite: "DejaVu Sans";
-$font2: "Montserrat", sans-serif;
+$font-primary: "Merriweather", sans-serif;
+$font-primary-composite: "Roboto";
+$font-secondary: "Fira Sans", sans-serif;
 
 $cols: 100% / 36;
 
@@ -22,24 +21,21 @@ $xga: 1024px * 36 / 56;
 $wxga: 1366px * 36 / 56;
 $fhd: 1920px * 36 / 56;
 
-$gray-darker: #191919;
-$gray-dark: #333;
-$gray: #666;
-$gray-light: #959595;
-$gray-lighter: #ccc;
+$text-color-body: #505050;
+$text-color-title: #282828;
 
-$fast: 0.25s;
-$normal: 0.5s;
-$slow: 1s;
+$section-color: #e9e9e3;
 
-//
-//
-//
+$divider-color: #ccc;
+$divider-color-composite: #959595;
+
+//////////////////////////////
+// Styling
+//////////////////////////////
 
 * {
     outline: none;
     -webkit-font-smoothing: antialiased;
-    -moz-osx-font-smoothing: grayscale;
     &[align="center"] {
         text-align: center;
     }
@@ -47,14 +43,11 @@ $slow: 1s;
 
 html, body {
     height: 100%;
+    font: 20px $font-primary;
 }
 
 body {
-    background: #ececec image-url("noise.png");
-}
-
-img {
-    vertical-align: middle;
+    background: white;
 }
 
 strong, b {
@@ -69,10 +62,6 @@ a[href] {
     color: #d16550;
     text-decoration: none;
 }
-
-//
-//
-//
 
 .hidden,
 .editsection,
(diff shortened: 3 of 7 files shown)


data/css/modules/card/_media.scss
```diff
@@ -1,35 +1,30 @@
...
 
-    separator {
-        background-color: $black-20-color;
-        min-height: 1px;
-    }
-
     &__caption {
-        font-size: .81em;  /* 13pt * 1/12 * 0.75 DPI correction factor */
-        font-family: $display-font;
+        font-size: 20px;
```
nedrichards: Drive by review with no actual knowledge - but do we want to keep using em's rather than px here?

data/css/modules/layout/_sidebar.scss
```diff
@@ -6,57 +6,49 @@
...
         .CardTitle {
             margin: 0px;
-            padding: 6px;
-            border: 2px solid transparentize(white, 1);
-            background-color: transparentize(white, 1);
-            transition: background-color 250ms ease-in-out,
-                        background-image 250ms ease-in-out,
-                        border 250ms ease-in-out;
+            padding: 10px;
+            // border-bottom: 1px solid transparentize(black, 1 - 0.7);
```
nedrichards: Any reason why we need to keep commented out code?

data/css/modules/layout/_sidebar.scss
```diff
@@ -6,57 +6,49 @@
...
+                background: transparentize(black, 1 - 0.5);
+                font-weight: 700;
 
                 .Card__title {
-                    color: $primary-dark-color;
-                    transition-property: none;
+                    color: white;
+                    transition-property: color;
+                    transition-timing-function: ease;
+                    transition-duration: 0.5s;
```
nedrichards: Nice! Have we tested perf on the mini or another lower end device for this? Or will it be disabled by default on those devices, I forget how we performance bucket.

data/templates/css/wikihow.scss
```diff
@@ -22,39 +21,32 @@ $xga: 1024px * 36 / 56;
...
+    font: 20px $font-primary;
 }
 
 body {
-    background: #ececec image-url("noise.png");
-}
-
-img {
-    vertical-align: middle;
+    background: #fff;
```
nedrichards: Stupid consistency question, but in some files we go for #fff and in some files, white. Any particular reason?

data/templates/css/wikihow.scss
```diff
@@ -221,16 +261,17 @@ h2 + ol, h3 + ol, .steps_list_2 {
...
 
-//
+
+//////////////////////////////
 // Responsive
-//
+//////////////////////////////
 
-@media (min-width: $vga) {
+@media (min-width: $vga) {  // FIXME: Haven't tested for vga yet
```
nedrichards: Do we need to fix this before merging? Or is it OK.

data/templates/css/wikimedia.scss
```diff
@@ -387,15 +458,15 @@ html {
...
 
-//
+//////////////////////////////
 // Responsive
-//
+//////////////////////////////
 
-@media (min-width: $vga) {
-    .mw-body {
+@media (min-width: $vga) {  // FIXME: Haven't tested for vga yet
```
nedrichards: Same as above, do we need to fix this before merging the PR?

data/templates/css/wikimedia.scss
```diff
@@ -411,6 +415,7 @@ html {
     .mw-body {
         $in-cols: 100% / 36;
         padding: 30px $cols * 2;
+        // padding: 10px 10px 10px 10px;
```
tchx84: Hi! If these are not necessary why leaving them :)?

data/templates/css/wikimedia.scss
```diff
@@ -283,28 +326,87 @@ html {
...
+    	
+    	// th {
+     //    	font: 700 1.5em $font-heading !important; /* 24px / 16px/em */
+     //    	background-color: $tan-color !important;
+     //        padding: 22px;
+     //    }
+
+     //    th a.eos-no-link {
+     //    	font: 700 1.5em $font-heading !important; /* 24px / 16px/em */
+     //    }
```
tchx84: If there are no needed, then just remove it.

data/templates/css/wikimedia.scss
```diff
@@ -283,28 +326,87 @@ html {
...
+
         // &.geography
         *[style*="nowrap"] {
             white-space: normal !important;
         }
         td, th {
-            border-top: 1px solid $gray-light;
-            padding: 0.4em 0.6em;
+            // border-top: 1px solid $gray-light;
+            // padding: 0.4em 0.6em;
```
tchx84: same here!

data/templates/css/wikihow.scss
```diff
@@ -8,10 +8,16 @@
...
-$font1-composite: "DejaVu Sans";
-$font2: "Montserrat", sans-serif;
+$font-primary: "Lato", sans-serif;
+$font-primary-composite: "Roboto";
+$font-secondary: "Fira Sans", sans-serif;
+
+
+
+
+
```
tchx84: we probably don't need all these blank lines :)

data/templates/css/wikihow.scss
```diff
@@ -145,25 +193,26 @@ $rhythm: 22px;
 h2 + ol, h3 + ol, .steps_list_2 {
     $in-cols: 100% / (36 - 2 * 2);
     > li {
-        border-left: 1px solid $gray-lighter;
+        // border-left: 1px solid $gray-lighter;
         clear: both;
-        margin-bottom: $rhythm;
-        padding-left: $in-cols * 2;
+        margin-bottom: 50px;;
+        // padding-left: $in-cols * 2;
```
tchx84: same here.

data/templates/css/wikihow.scss
```diff
@@ -145,25 +193,26 @@ $rhythm: 22px;
...
         @include box-sizing(border-box);
         @include pie-clearfix;
     }
     .mwimg {
         margin: 0 auto;
         margin-bottom: $rhythm;
         position: relative;
-        width: $in-cols * 30;
+        // width: $in-cols * 30;
```
tchx84: same here, and for the others.

data/css/modules/card/_media.scss
```diff
@@ -22,7 +22,6 @@
         font-size: .62em;  /* 10pt * 1/12 * 0.75 DPI correction factor */
         font-family: $display-font;
         color: $accent-light-color;
-        transition: color 250ms ease-in-out;
```
tchx84: The commit description mentions changes in the sidebar and text styling, but the commit contains many changes that are not related to that. Including changes to the non-GTK styling sheets.

Can you double check whether these changes belong here? and if they do, elaborate a bit more in the description to explain what this commit achieves.

data/css/modules/layout/_sidebar.scss
```diff
@@ -7,7 +7,8 @@
         .CardTitle {
             margin: 0px;
             padding: 10px;
-            border-bottom: 2px solid transparentize(black, 1 - 0.1);
+            border-bottom: 1px solid transparentize(black, 1 - 0.2);
+            transition: all 0.2s ease;
```
tchx84: I think some changes don't corresponding to this commit? Please double check! :)

data/templates/css/wikihow.scss
```diff
@@ -368,14 +364,31 @@ h2 + ol, h3 + ol, .steps_list_2 {
...
+        .eos-article-title {
+            padding: 20px 0px 40px 0px; // 40px because of top padding of 20px
         }
     }
-}
+
+    // .section li,
+    // .section ul {
+    //     padding: 20px;
+    // }
```
tchx84: Same thing re. leaving these here.

data/css/modules/card/_media.scss
tchx84: @mattdangerw your opinion is more than welcome here :)

data/templates/css/wikihow.scss
```diff
@@ -266,31 +261,17 @@ h2 + ol, h3 + ol, .steps_list_2 {
...
-    padding-top: 40px;
-    padding-bottom: 40px;
-    margin-top: 40px;
-    border-top: 1px solid $gray-lighter;
-    clear: both;
-    p {
-        width: 100% !important;
-    }
-}
-
```
tchx84: I noticed you added this earlier, maybe it would better to aggregate these changes?

data/css/modules/layout/_sidebar.scss
```diff
@@ -18,20 +18,14 @@
             }
 
             &:hover {
-                // background: transparentize(white, 1 - 0.05);
```
tchx84: It would be better to aggregate these changes to the commit where you commented these line.

data/css/modules/card/_media.scss
mattdangerw: I think going `px` is fine here and lots of other places in a framework css dos the same.

The problem with `em` is since it cascades its a really bad fit for our modular setup. `2em` inside a module that set font size to be `1.5em` and your fonts are looking way different then in another modular hierarchy.

`rem` could be a possibility going forward, but its kinda useless unless we are doing some responsiveness with the the base font size which we are not. And that's out of scope for this issue anyway.

nedrichards: lgtm

data/css/modules/card/_media.scss
```diff
@@ -1,35 +1,30 @@
...
-    separator {
-        background-color: $black-20-color;
-        min-height: 1px;
-    }
-
     &__caption {
-        font-size: .81em;  /* 13pt * 1/12 * 0.75 DPI correction factor */
-        font-family: $display-font;
+        font-size: 20px;
+        font-family: Merriweather;
```
tchx84: Just a reminder of what we talked on slack about using specific fonts in modules partials, we should ideally replace it with the proper scss font variables unless we really want all of our apps to use this specific font.

data/templates/css/wikimedia.scss
```diff
@@ -43,6 +43,7 @@ $divider-color-composite: #959595;
 }
```
tchx84: Try to break your commit messages lines, you can find some good guidelines [here](https://github.com/endlessm/eos-documentation/wiki/Commit-Message-Practices)

data/templates/css/wikimedia.scss
```diff
@@ -76,253 +77,317 @@ div[style*="float:right"] {
...
+            max-width: 100%;
+        }
+
+        h2, .firstHeading, h3, h4, h5 {
+        	color: $text-color-title;
+            line-height: 1.125;
+        }
+
+        h2, .firstHeading {
+        	font: 700 2.2rem $font-secondary;  // 44px // Root: 20px
```
tchx84: watch out with tabs :) we want spaces here! (and everywhere else)

data/templates/css/wikimedia.scss
```diff
@@ -76,253 +77,317 @@ div[style*="float:right"] {
...
+                width: 100% !important;
+                display: inline-block;
+                margin-bottom: 20px;
+
+                div, .gallerytext p, .gallerytextwrapper {
+                    width: auto !important;
+                }
+
+                .gallerytext {
+                	font-family: $font-secondary; // 16 px
```
tchx84: idem! No tabs!

Note that you can make a commit that fixes this, and then use the same interactive rebase trick to unify these fixes in this same commit.

data/templates/css/wikimedia.scss
```diff
@@ -76,253 +77,317 @@ div[style*="float:right"] {
...
+                }
+                &:last-child:after {
+                    display: none;
+                }
             }
         }
-        &[style*="text-align: center"] {
-            td, th {
+        .wikitable {
+        	font-size: 0.8rem;	// 16px // Root: 20px
```
tchx84: Idem.

data/templates/css/wikimedia.scss
```diff
@@ -76,253 +77,317 @@ div[style*="float:right"] {
...
-            display: inline;
-            margin: 0;
-        }
-        dd, li {
-            &:after {
-                content: " · ";
-                font-weight: bold;
+
+        tbody {
+    		font-size: 0.8rem;	// 16px // Root: 20px
```
tchx84: There are probably more, so just make sure to search for \t

data/templates/css/wikimedia.scss
```diff
@@ -76,253 +77,317 @@ div[style*="float:right"] {
...
-        th {
-            font-weight: 700;
-        }
-        .maptable {
-            td, th {
-                border-top: 0 none;
+            li {
+                display: inline;
+                &:before {
+                    // content: " · ";
```
tchx84: what this selector do?

data/templates/legacy-article.mst
```diff
@@ -11,13 +11,12 @@
     {{/title}}
     <section id="inside-content" class="section">
         <div>{{{body-html}}}</div>
-        <hr>
```
tchx84: I noticed you removed this, but also that you moved the disclaimer out of this section, wouldn't that be a problem considering that the selectors [here](https://github.com/endlessm/eos-knowledge-lib/pull/1267/commits/94249c5614659360e65fe4e47e1996337759d42b#diff-db16cecbc2ecc49d958cd74c4b659d68R526) assume that this going inside that section?

Please let me know if I am missing something :)

data/css/modules/card/_media.scss
```diff
@@ -3,7 +3,7 @@
 
     &__caption {
         font-size: 20px;
-        font-family: Merriweather;
+        font-family: $body-font;
```
tchx84: You can use interactive rebase to fix your previous commit (basically by moving this commit below the original commit and set it as fixup)

data/templates/css/wikimedia.scss
```diff
@@ -431,15 +496,48 @@ html {
...
+    .wikitable,
+    tbody,
+    .thumbcaption,
+    .disclaimer {
+        font-size: 1.0rem !important;  // 20px // Root: 20px
     }
 }
+
+
+
```
tchx84: I didn't see this on wikihow articles though, so maybe is something related only to the wikipedia scss?

data/templates/legacy-article.mst
katelk28: Good point, but as far as I know, `<hr>` and borders are different. 
In the .disclaimer class, I added a top-border, which adds a border to the .disclaimer div --- it's not manipulating the `<hr>`. They just look very similar.

data/templates/css/wikimedia.scss
katelk28: My mistake. Forgot to take the comment off.

data/templates/css/wikimedia.scss
katelk28: So this is actually a separate problem, and I've noticed that too.

Here's what happened: I made a design decision that, in composite mode, adds a `border-bottom` to the `h2` and `eos-article-title` elements, b/c it's very hard to differentiate between sections when the screen is that small.

BUT for some reason, maybe in the ingestor, when a section is hidden (like the "References" section or something), it seems that the `h2` element is still there in the HTML, even though it doesn't display anything. That's why the `border-bottom` is still applied.

I'm not sure if that's something that can be fixed with CSS, but I guess for now, since it's making that ugly bug, I will take out the `border-bottom` until I come up with a better solution.

data/templates/legacy-article.mst
tchx84: I meant something else. What I meant is that before this patch we have:
```
<inside-content>
  <disclaimer>
</inside-content>
```

After this patch we have:

```
<inside-content>
  ...
</inside-content>
<disclaimer>
```

But, the CSS selectors still assume the previous structure [here](https://github.com/endlessm/eos-knowledge-lib/pull/1267/commits/cac136f4c92bc190d9eb830154bb58aa0f23d377#diff-db16cecbc2ecc49d958cd74c4b659d68R521), and I this is correct that selector should not be working.

tchx84:
Looking better now, just noticed one more thing though:

screenshot from 2016-12-13 13-15-45

As you can see in the screenshot, there is a big gap between title and the content.

I tested with random articles from encyclopedia, astronomy and howto apps, but only noticed this in astronomy, some examples are "earth science" and "physical cosmology" articles. Try comparing with master (let me know if you need help getting these shards).

Also, since not all the articles use the exact same structure is not always possible to cover every single case. Sometimes we fix something and possibly break other things, so I would suggest to give a quick try and see if you add a selector for cover this case (unless you find something evidently wrong).

tchx84:
**CR PASS**

PR #1269, opened 2016-12-01
EknMediaBin: workaround fullscreen bug

xjuan:
GtkGstGLWidget does not support reparenting to a different toplevel
because the gl context is different and the pipeline does not know it
changes, so as a temporal workaround we simply reconstruct the whole
pipeline.

https://bugzilla.gnome.org/show_bug.cgi?id=775045

https://phabricator.endlessm.com/T14099

### DIFF
--- a/lib/eosknowledgeprivate/ekn-media-bin.c
+++ b/lib/eosknowledgeprivate/ekn-media-bin.c
@@ -96,6 +96,7 @@ typedef struct
   GtkAdjustment  *playback_adjustment;
 
   /* Support Objects */
+  GtkWidget *video_widget;      /* Created at runtime from sink */
   GtkWindow *fullscreen_window;
   GdkCursor *blank_cursor;
   GtkWidget *tmp_image;      /* FIXME: remove this once we can derive from GtkBin in Glade */
@@ -112,7 +113,7 @@ typedef struct
 
   /* Gst support */
   GstElement *play;          /* playbin element */
-  GstElement *video_sink;    /* The video sink element used (gtkglsink or gtksink) */
+  GstElement *video_sink;    /* The video sink element used (glsinkbin or gtksink) */
   GstElement *vis_plugin;    /* The visualization plugin */
   GstBus     *bus;           /* playbin bus */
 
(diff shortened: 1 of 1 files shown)


lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -637,26 +601,208 @@ ekn_media_bin_video_sink_new (EknMediaBin *self)
...
+   * We can not wait for a GtkWidget::draw event, because that might not
+   * happen if the window is fully obscured by windowed child widgets.
+   * Alternatively, we could wait for an expose event on widget's window.
+   * Both of these are rather hairy, not sure what's best. */
+  gdk_event_handler_set (check_for_draw, loop, NULL);
+  g_main_loop_run (loop);
+
+  /* give the WM/server some time to sync. */
+  gdk_display_sync (gdk_window_get_display (window));
+  g_timeout_add (12, quit_when_idle, loop);
```
rmacqueen: what's this 12? Should that be put in a constant somewhere?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -637,26 +601,208 @@ ekn_media_bin_video_sink_new (EknMediaBin *self)
...
+}
+
+static void
+ekn_media_bin_fullscreen_apply (EknMediaBin *self, gboolean fullscreen)
+{
+  EknMediaBinPrivate *priv = EMB_PRIVATE (self);
+  gint64 position = -1;
+
+  /*
+   * To avoid flickering, this will make the widget pack a image with the last
```
rmacqueen: *an image

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -637,26 +601,208 @@ ekn_media_bin_video_sink_new (EknMediaBin *self)
...
+      GdkPixbuf *pixbuf = ekn_media_bin_video_pixbuf_new (self);
+      priv->tmp_image = gtk_image_new_from_pixbuf (pixbuf);
+      g_object_set (priv->tmp_image, "expand", TRUE, NULL);
+      g_object_unref (pixbuf);
+    }
+
+  /*
+   * FIXME: GtkGstGLWidget does not support reparenting to a different toplevel 
+   * because the gl context is different and the pipeline does not know it
+   * changes, so as a temporal workaround we simply reconstruct the whole
```
rmacqueen: *temporary workaround

rmacqueen: Unless you mean to say that this workaround is of this material world, and not of the kingdom of heaven :-)

rmacqueen:
See the video I sent on Slack - I still get that weird resizing bug when returning to normal from full screen. I also see the control bar flashing while in full screen mode.

xjuan:
The function I used to wait for changes to show up in the screen was problematic, and never returning control back to the main loop, so I simply removed it.

rmacqueen:
This looks good to me, and is working, but I'm also not able to now reproduce the original bug on master. I think this is because it only shows up on real hardware.

tchx84:
I tested in my VM and I see the full screen bug happening in master and I see the full screen working with this patch, so CR PASS.

PR #1270, opened 2016-12-01
debian/control: fix eos-knowledge-0 dependency of content lib

mattdangerw:


ptomato:
**CR pass**

PR #1272, opened 2016-12-01
Fix jasmine test runners

mattdangerw:
We won't run the ekncontent tests from the main directory. Those
can just be trigger by runing jasmine in the ekncontent repo.

That should make it a little easier to handle import paths and
decouple the two test suites.

### DIFF
--- a/ekncontent/jasmine.json
+++ b/ekncontent/jasmine.json
@@ -1,0 +1,11 @@
+{
+    "spec_files": ["tests"],
+    "options": "--verbose",
+    "environment": {
+        "GJS_PATH": ".:js",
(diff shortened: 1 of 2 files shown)


rmacqueen:
I can now run `jasmine <path_to_test>` but I still get this message when doing so:

```
Configuration loaded from /home/endless/checkout/eos-knowledge-lib/jasmine.json
'g_io_module_load': ekncontent/.libs/libeos-knowledge-content-0.so: undefined symbol: g_io_module_load
Failed to load module: ekncontent/.libs/libeos-knowledge-content-0.so
Configuration loaded from /home/endless/checkout/eos-knowledge-lib/jasmine.json
```

PR #1274, opened 2016-12-05
T13418

mattdangerw:


ekncontent/ekncontent/eknc-utils.c
```diff
@@ -14,6 +16,86 @@
...
+                                                     &error);
+      g_value_set_variant (&param.value, variant);
+    }
+  else if (JSON_NODE_HOLDS_VALUE (node))
+    {
+      GValue value = { 0, };
+      GType type = json_node_get_value_type (node);
+
+      // For some reason some of our integer properties are stored in json as
+      // strings. Try to parse if we hit this case.
```
ptomato: :thinking: This looks like it's covering up a bug somewhere else

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -14,6 +16,86 @@
...
+        }
+
+      if (g_value_type_transformable (pspec->value_type, type))
+        g_value_transform (&value, &param.value);
+      else
+        g_set_error (&error, 0, 0, "Unexpected type %s", g_type_name (type));
+    }
+  else
+    {
+      g_set_error (&error, 0, 0, "Unexpected json value.");
```
ptomato: nitpick: remove period, as other error handlers often append a colon to an error message

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -14,6 +16,86 @@
...
+        }
+      else
+        {
+          json_node_get_value (node, &value);
+        }
+
+      if (g_value_type_transformable (pspec->value_type, type))
+        g_value_transform (&value, &param.value);
+      else
+        g_set_error (&error, 0, 0, "Unexpected type %s", g_type_name (type));
```
ptomato: Yes, I agree with your other comment that you should have a real error quark and error code enum

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -14,6 +16,86 @@
...
+/**
+ * eknc_utils_free_gparam_array:
+ *
+ * Frees a GArray of GParameters.
+ */
+void
+eknc_utils_free_gparam_array (GArray *params)
+{
+  for (gint i = 0; i < params->len; i++) {
+    GParameter *param = &g_array_index (params, GParameter, i);
```
ptomato: This looks weird, but as far as I can tell it's correct.

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -0,0 +1,578 @@
...
+                                     NPROPS,
+                                     eknc_content_object_model_props);
+}
+
+static void
+eknc_content_object_model_init (EkncContentObjectModel *self)
+{
+}
+
+void
```
ptomato: gtk-doc will probably complain if you don't put a doc comment here, but I'd suggest annotating it with `(skip)`

ekncontent/ekncontent/eknc-object-model.h
```diff
@@ -0,0 +1,13 @@
+/* Copyright 2016 Endless Mobile, Inc. */
+
+#pragma once
+
+#include "eknc-content-object-model.h"
+
+G_BEGIN_DECLS
+
+EkncContentObjectModel *
+eknc_object_model_from_string (const gchar *str,
```
ptomato: Consider making a companion function to create one from a Json node (or even instead of the from-string function)

tests/js/app/testArticleHTMLRenderer.js
```diff
@@ -6,13 +6,33 @@ const SearchUtils = imports.search.utils;
...
 describe('Article HTML Renderer', function () {
-    let wikihow_model, wikibooks_model;
+    let wikihow_model, wikibooks_model, wikipedia_model, wikisource_model;
     let renderer;
 
     beforeEach(function () {
         Utils.register_gresource();
 
         renderer = new ArticleHTMLRenderer.ArticleHTMLRenderer();
+        wikipedia_model = Eknc.ArticleObjectModel.new_from_props({
```
ptomato: CR pass on this commit too, but it doesn't apply without the other commits. I'd merge it if you ported it to use the old API

ekncontent/ekncontent/eknc-utils.c
mattdangerw: Just an inconsistency in our data. A lot of our other number (duration, width, height) are store in our json metadata as string. The toc item `hasIndex` is the only thing I can find that's an actual number there.

Would be great to cleanup on the server side either all number in json as number or none. Then we'd probably need to keep a compat layer, but don't want this issue to extend to changing our on disk data format.

Can open up some follow ups though.

ekncontent/ekncontent/eknc-object-model.h
mattdangerw: Hmm, maybe I'll just move this to json node for now? And engine can just call `imports.gi.Json.from_string`, and then call the function.

Might be nice to expose a from_string function anyway, but I'll wait till engine/domain get ported to C before making any calls on API here. We don't need to expose anything function like this by then end of the port if we don't want to.

mattdangerw:
> Consider squashing "move in media object tests" into "port media object model to C"

The media object tests are actually testing image object model and video object model. So I could squash it but the tests will fail until the next two commits land. Probably bad for bisect.

I'm thinking of rewriting a bit so each c object gets its own jasmine unit test file, but for a first pass was focusing on preserving the existing test coverage in C.

ekncontent/ekncontent/eknc-utils-private.h
```diff
@@ -0,0 +1,18 @@
...
+
+#pragma once
+
+#include <gio/gio.h>
+#include <json-glib/json-glib.h>
+
+G_BEGIN_DECLS
+
+void
+eknc_utils_append_gparam_from_json_node (JsonNode *node,
```
mattdangerw: erg do I have to :P
One of my least favorite parts of gnome style because it always ends up breaking with time. https://git.gnome.org/browse/gtk+/tree/gtk/gtkwidget.h#n1110

I had vaguely planned on using a new style here--just copy the decl from your c file, add a semi colon, done.

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -0,0 +1,578 @@
...
+    }
+}
+
+static void
+eknc_content_object_model_finalize (GObject *object)
+{
+  EkncContentObjectModel *self = EKNC_CONTENT_OBJECT_MODEL (object);
+  EkncContentObjectModelPrivate *priv = eknc_content_object_model_get_instance_private (self);
+
+  g_free (priv->ekn_id);
```
mattdangerw: g_variant_unref does not handle NULL as far as i know.
g_free does.

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -0,0 +1,578 @@
...
+  /**
+   * EkncContentObjectModel:ekn-id:
+   *
+   * Unique ID of the model
+   *
+   * This is an internal ID assigned by EKN.
+   * If none is provided, the model will generate its own id.
+   */
+  eknc_content_object_model_props[PROP_EKN_ID] =
+    g_param_spec_string ("ekn-id", "Object's ID",
```
mattdangerw: I'd like to keep the port as close as possible to a direct port, just in terms of diff cleanliness. Then I can just copy in test files in, rename construction calls in js as nice clear commits.

But I can add a commit on top moving to id if that's something we want.

ekncontent/ekncontent/eknc-utils.c
magcius: I believe this is because ImageObject, which our schemas are based on, have width/height defined as a "Distance": https://schema.org/ImageObject which is supposed to be a unit-ful value like "500px" or "5 inches". We don't use any of that, but width/height are indeed defined as strings in our object models: https://github.com/endlessm/endless-content-factory/blob/master/src/eos-knowledge-db-build/database_build/packer/models/media_object.py#L17-L18

ekncontent/tests/ekncontent/testDatadir.js
```diff
@@ -0,0 +1,14 @@
...
+
+describe('Data dir finder', function () {
+    it('returns null if the domain was not found', function () {
+        let found_dir = Eknc.get_data_dir('this-knowledge-app-should-never-exist');
+        expect(found_dir).toBeNull();
+    });
+
+    // FIXME: no easy way to test these without a way to mock file URIs
+    // during tests.
+    it('returns a GFile pointing to a valid ekn data dir');
```
mattdangerw: Kinda thinking out loud here, but maybe a way to do this would be to define some method or environment variable...
`EKN_CONTENT_USE_TEST_URI`
or
`eknc_set_use_test_uri (TRUE)`
which would load everything with a `test://` uri, we could then do a `vfs.register_uri_scheme('test')` to serve whatever content we want.

Could play with that more. But since it involves more vfs stuff I'd rather wait till all that shakes out.

Alternately could pass a prefix along to `Eknc.get_data_dir` to use instead of the rootfs if set. And then write some directories to disk like the old tests did. But I am not a fan of either changing the exposed API for tests, or writing to disk during tests if avoidable.

ekncontent/ekncontent/eknc-content-object-model.c
ptomato: Up to you, but it does seem desirable to review the API as long as we have the chance. Once this is fully landed I'd like to start treating it as public API.

ekncontent/ekncontent/eknc-content-object-model.c
ptomato: Well, add a docstring in any case. (I'm surprised gtk-doc isn't marking it as missing.) If you don't `(skip)` then add real documentation as well :wink: otherwise it can be a skeleton docstring.

ekncontent/ekncontent/eknc-utils-private.h
ptomato: I dislike that style as well, but then I dislike the whole GNOME C style...

ekncontent/ekncontent/eknc-utils.c
ptomato: OK, makes sense. In that case replace the "For some reason" comment with the reason :smile:

ekncontent/tests/ekncontent/testDatadir.js
ptomato: Still hoping to eventually use https://github.com/ptomato/gt/blob/master/src/mockfile.vala for that...

ekncontent/ekncontent/eknc-utils.c
ptomato: And that possibly means that the property should be string-valued here, rather than integer valued?

ekncontent/ekncontent/eknc-content-object-model.c
ptomato: Oh, it's probably because you have it in `<SUBSECTION Private>`. Never mind then

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -0,0 +1,580 @@
...
+    g_param_spec_string ("copyright-holder", "Copyright Holder",
+      "The copyright holder of the object",
+      "", G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS);
+  /**
+   * EkncContentObjectModel:source-uri:
+   *
+   * URI where this content was downloaded from during database build
+   *
+   * TODO Should we check to always have a value for source-uri?
+   * Don"t use this property for user-visible things.
```
ptomato: nitpick: couple instances of `"` where `'` should be

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -0,0 +1,580 @@
...
+   * The source content"s mimetype
+   */
+  eknc_content_object_model_props[PROP_CONTENT_TYPE] =
+    g_param_spec_string ("content-type", "Object Content Type",
+      "Mimetype of the source content",
+      "", G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS);
+  /**
+   * EkncContentObjectModel:synopsis:
+   *
+   * The synopsis for this content object. Defaults to an empty string.
```
ptomato: nitpick: As long as you're touching this code, get rid of the "Defaults to ..." comments; the default values are automatically extracted by gtk-doc.

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -0,0 +1,580 @@
...
+  /**
+   * EkncContentObjectModel:last-modified-date:
+   *
+   * The date of last modification for this content object. It treats dates
+   * according to the ISO8601 standard.
+   */
+  eknc_content_object_model_props[PROP_LAST_MODIFIED_DATE] =
+    g_param_spec_string ("last-modified-date", "Last Modified Date",
+      "Last Modified Date of the document or media object",
+      "", G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS);
```
ptomato: Is there a boxed type for dates in GLib we could use here? Maybe with an override to convert to/from `Date` objects in JS?

ekncontent/ekncontent/eknc-content-object-model.c
ptomato: Oh, I got it backwards. Never mind.

ekncontent/overrides/EosKnowledgeContent.js
```diff
@@ -0,0 +1,116 @@
+const Json = imports.gi.Json;
+const GLib = imports.gi.GLib;
+
+function define_propery (klass, name, descriptor) {
```
ptomato: typo: `define_propery` -> `define_property` throughout

ekncontent/tests/ekncontent/testObjectModel.js
```diff
@@ -0,0 +1,75 @@
+const Eknc = imports.gi.EosKnowledgeContent;
+const Json = imports.gi.Json;
+
+const InstanceOfMatcher = imports.tests.InstanceOfMatcher;
+
+describe ('Object model from json node', function () {
```
ptomato: nitpick: no spaces after `describe` and `it`

ekncontent/ekncontent/eknc-utils.h
```diff
@@ -8,6 +8,19 @@
...
+ * @EKNC_CONTENT_OBJECT_ERROR_BAD_FORMAT: unexpected format in metadata json
+ *
+ * Error enumeration for object model creation.
+ */
+typedef enum _EkncContentObjectError {
+  EKNC_CONTENT_OBJECT_ERROR_BAD_FORMAT,
+} EkncContentObjectError;
+
+#define EKNC_CONTENT_OBJECT_ERROR eknc_content_object_error_quark ()
+GQuark eknc_content_object_error_quark (void);
```
ptomato: Should this be moved up into an earlier commit so that you can use the error quark there as well? There are some more instances where you just passed 0

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -198,6 +198,106 @@ eknc_utils_parallel_init (GSList        *initables,
...
+ * eknc_get_running_under_flatpak
+ *
+ * Checks if we are running in a flatpak sandbox.
+ *
+ * Returns: true if we are running in a flatpak sandbox.
+ */
+gboolean
+eknc_get_running_under_flatpak (void)
+{
+  g_autoptr(GError) error = NULL;
```
ptomato: nitpick: space after `g_autoptr`

ekncontent/ekncontent/eknc-content-object-model.c
mattdangerw: oh yeah, not saying it won't work, just that it adds nothing. You'd be calling `g_clear_pointer (mystring, g_free)` with a useless extra function call you didn't need.

But I don't feel strongly so I can flip. What the benefit of g_clear_pointer?

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -198,6 +198,106 @@ eknc_utils_parallel_init (GSList        *initables,
...
+                                                     app_id, NULL);
+  GFile *file = g_file_new_for_path (database_dir);
+  if (g_file_query_exists (file, NULL))
+      return file;
+  g_object_unref (file);
+  return NULL;
+}
+
+/**
+ * eknc_get_running_under_flatpak
```
ptomato: missing `:`, doesn't gtk-doc complain about this?

ekncontent/ekncontent/eknc-utils.c
mattdangerw: We could consider changing this `hadIndex` to a string, but we'd probably still want the compatibility with the old int version. Probably not worth the effort, I'd rather look at a more generally friendly table of contents format.

But again, any discussions of data format changes probably best done in a follow up.

ekncontent/ekncontent/eknc-content-object-model.c
mattdangerw: Sure. Commit wise I would do this as a new thing tagged on the end. It'll make this branch easier for me to drive/reason about. But I can add that on when I get through the bulk of the port.

As to renaming, it does not seem super clear cut. `Id` would be more consistent with other libraries, shard included. 'ekn-id' is a name that ends up sticking in our heads and discussions though. You can say 'ekn-id' and everyone knows what you are talking about without needed to contextualize. Though maybe not a outsider friendly as a term.

magcius: eh, we have multiple names. EKN URI and EKN ID are both synonymous. Shard does not have any notion of ID, it indexes its records by "names", e.g. hex_name or raw_name.

If we are doing this, I would prefer "ekn URI" over "ekn ID" or "ID".

ekncontent/ekncontent/eknc-utils.c
mattdangerw: no, which is odd. gtk-doc is definitely working in make check though, im using that regularly to get all my symbols properly in sections

ekncontent/ekncontent/eknc-utils.c
ptomato: in that case, remove the space after `g_autoptr` a couple lines below here

mattdangerw: done

ekncontent/ekncontent/eknc-content-object-model.c
mattdangerw: Ohh I see, yeah that make sense, probably good for dangling refs to the object. I'm sold. And gnome-builder seems to use this style everywhere.

ekncontent/ekncontent/eknc-content-object-model.c
mattdangerw: Didn't find any boxed type or good example of date properties. I'd vote lets stick with the status quo for now, but totally fine with that as a future cleanup.

ekncontent/ekncontent/eknc-media-object-model.c
```diff
@@ -0,0 +1,205 @@
...
+      break;
+
+    default:
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
+    }
+}
+
+static void
+eknc_media_object_model_set_property (GObject *object,
+                                        guint prop_id,
```
ptomato: wrong alignment

ekncontent/tests/ekncontent/testObjectModel.js
ptomato: Still needs addressing

ptomato:
Just needs the comment in the top commit fixed so it doesn't say "For some reason". I think it can probably be merged after that, I still need to poke at the tests a bit though

mattdangerw:
Fixed up!

ptomato:
OK, I think this can be merged with the understanding that it's a 1-for-1 port of the JS and opportunities for cleanups will be taken advantage of in the future.

PR #1275, opened 2016-12-05
EknVfs: fixed crash when running inside flatpak

xjuan:
Handle NULL from vfs->get_supported_uri_schemes()
Use local vfs instance instead of using g_vfs_get_local()

https://phabricator.endlessm.com/T14455


mattdangerw:
Discussed on slack. Seems like none of fully understand the differences in the flatpak sandbox with vfs extension points. But this does seem to fix the problem, and we need our runtime back up and running for other work.

So lets do it!

PR #1276, opened 2016-12-06
build: include ekncontent builddir for g-ir-scanner

ptomato:
This is necessary to make the build work when srcdir != builddir,
otherwise eknc-version.h can't be found during the g-ir-scanner step.


mattdangerw:
CR pass

PR #1279, opened 2016-12-06
EknVfs: fixed segfault while initializing instance

xjuan:
Add reference to priv->local object.
Use g_initable_new() for initiable types.

https://phabricator.endlessm.com/T14455


mattdangerw:
Works outside sandbox
Works inside sandbox
make check passes

Looks good to me! Thanks @xjuan

PR #1280, opened 2016-12-06
T14472

magcius:
https://phabricator.endlessm.com/T14472

### DIFF
--- a/js/search/domain.js
+++ b/js/search/domain.js
@@ -78,10 +78,6 @@ const Domain = new Lang.Class({
 
         this._load_sync_internal();
         this._shard_inited = true;
-
-        /* Append shards to default EknVfs for ekn:// uri to work */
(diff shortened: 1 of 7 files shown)


js/search/domain.js
```diff
@@ -78,6 +78,10 @@ const Domain = new Lang.Class({
 
         this._load_sync_internal();
         this._shard_inited = true;
+
+        /* Append shards to default EknVfs for ekn:// uri to work */
+        if (this._shards)
+            Eknc.default_vfs_set_shards (this._shards);
```
magcius: This seems wrong, considering just creating a new domain will overwrite the default shards.

xjuan: You are right, the only place we should set the shards is in the default engine

xjuan:
CE pass

PR #1286, opened 2016-12-08
Add background image for activity PDFs

rmacqueen:
Since they dont have any image of their own to show.


ptomato:
Either this should be part of the overrides.scss in the apps themselves, or the asset should be included in eos-knowledge-lib and not in the individual apps. I prefer the former.

PR #1288, opened 2016-12-10
Fix deadlocked builds

ptomato:
https://phabricator.endlessm.com/T14477

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -223,8 +223,11 @@ EOS_CHECK_GJS_GIR([GLib], [2.0])
 EOS_CHECK_GJS_GIR([GObject], [2.0])
 EOS_CHECK_GJS_GIR([Gtk], [3.0])
 EOS_CHECK_GJS_GIR([Pango], [1.0])
+EOS_CHECK_GJS_GIR([Rsvg], [2.0])
 EOS_CHECK_GJS_GIR([WebKit2], [4.0])
 EOS_CHECK_GJS_GIR_API([Gdk], [Rectangle])
+AS_IF([test "x$enable_metrics" != "xno"],
+    [EOS_CHECK_GJS_GIR([EosMetrics], [0])])
 
 # GStreamer plugins needed for EknMediaBin widget
 m4_pattern_allow([AM_GST_ELEMENT_CHECK])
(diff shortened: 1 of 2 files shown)


tests/js/app/interfaces/testCard.js
```diff
@@ -141,9 +141,8 @@ describe('Card interface', function () {
 
     it('sets a thumbnail frame visible if model has a thumbnail uri', function () {
         let engine = MockEngine.mock_default();
-        engine._lookup_ekn_uri.and.callFake(() => {
-            return Gio.Vfs.get_local().get_file_for_path('/dev/zero');
-        });
+        let zeroes = Gio.Vfs.get_local().get_file_for_path('/dev/zero');
```
mattdangerw: I wonder if we even need to single out vfs local for this. Shouldn't `Gio.File.new_for_uri('file:///dev/zero')` work?

mattdangerw: Do we even need to single out Gio.Vfs.get_local() here? Wouldn't `let zeroes = Gio.File.new_for_uri('file:///dev/zero');` work?

Maybe I'm not understanding the vfs setup completely...

configure.ac
```diff
@@ -223,8 +223,11 @@ EOS_CHECK_GJS_GIR([GLib], [2.0])
 EOS_CHECK_GJS_GIR([GObject], [2.0])
 EOS_CHECK_GJS_GIR([Gtk], [3.0])
 EOS_CHECK_GJS_GIR([Pango], [1.0])
+EOS_CHECK_GJS_GIR([Rsvg], [2.0])
```
magcius: unrelated???

mattdangerw:
CR pass

tests/js/app/interfaces/testCard.js
dbnicholson: Or even just `Gio.File.new_for_path('/dev/zero')` should be fine here, right?

mattdangerw: Yar true. But just merged

ptomato: `new_for_path` was what was actually merged

PR #1289, opened 2016-12-12
Downloader ndn

xaiki:
Keeping this as pending, DO NOT MERGE until NDN has been pushed to the rest of the OS.

### DIFF
--- a/js/search/downloader.js
+++ b/js/search/downloader.js
@@ -1,144 +1,20 @@
 // Copyright 2016 Endless Mobile, Inc.
 
-const Format = imports.format;
-const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
-const Lang = imports.lang;
 
-const AsyncTask = imports.search.asyncTask;
-const Config = imports.search.config;
-const Utils = imports.search.utils;
-
-// Returns items that are in A but not in B.
-function subtract_set (A, B, key) {
-    let result = [];
-    let mapped_A = A.map(key);
-    let mapped_B = B.map(key);
-
-    for (let i = 0; i < mapped_A.length; i++) {
-        let item = mapped_A[i];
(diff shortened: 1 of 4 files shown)


ptomato:
@xaiki Are you intending this to be reviewed or does "pending" mean you are still working on it?

xaiki:
@ptomato please review the last one. that said DO NOT MERGE IT, we need this to be the *last* part of the NDN integration to get in.

xaiki:
by last i mean 8f573ac

js/search/downloader.js
```diff
@@ -313,57 +120,16 @@ const SubscriptionDownloader = new Lang.Class({
...
-                this._download_new_shards(directory, old_manifest, new_manifest, cancellable, task.catch_callback_errors((source, download_task) => {
-                    let have_new_manifest = this._download_new_shards_finish(download_task);
-                    if (!have_new_manifest)
-                        new_manifest_file.delete(cancellable);
-
-                    task.return_value(true);
-                }));
-            }));
+            let proxy = new DownloaderProxy(Gio.DBus.session, 'com.endlessm.EknSubscriptionsDownloader', '/com/endlessm/EknSubscriptionsDownloader');
+            proxy.DownloadSubscriptionRemote(subscription_id);
```
ptomato: This code will never call the callback, except if either of these statements throws an exception. If you want to call the callback, then you should add a `() => task.return_value()` to the `DownloadSubscriptionRemote` invocation here. But, it might be the case that the downloader should be fire-and-forget and the app has nothing sensible to do for error handling. In that case you could just remove the callback and the `fetch_update_finish` method altogether.

xaiki:
rebased. (still not for merge)

ptomato:
This code doesn't exist anymore. Unfortunately we never got around to fixing this up.

PR #1292, opened 2016-12-13
Makefile: include ekncontent overrides in main testsuite

mattdangerw:
By setting the GJS_PATH. We shouldn't rely on make install to be
run before make check


ptomato:
Let's give it a try, looks reasonable to me...

PR #1293, opened 2016-12-14
Style fixes for escola template

rmacqueen:
Based on Design QA

https://phabricator.endlessm.com/T14133

### DIFF
--- a/js/app/interfaces/card.js
+++ b/js/app/interfaces/card.js
@@ -272,6 +272,18 @@ const Card = new Lang.Interface({
         label.visible = !!text;
     },
 
+    set_duration_label: function (label, duration) {
+        if (typeof duration !== 'undefined') {
+            let minutes = Math.floor(duration / (1000 * 60));
+            let seconds = Math.floor((duration / 1000)) % 60;
+            let count = (minutes === 1 && seconds === 0) ? 1 : minutes;
+            label.visible = true;
+            label.label = ngettext("%s:%s minute", "%s:%s minutes", count).format(minutes, ("00" + seconds).substr(-2, 2));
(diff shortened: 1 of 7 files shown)


js/app/interfaces/card.js
```diff
@@ -272,6 +272,18 @@ const Card = new Lang.Interface({
         label.visible = !!text;
     },
 
+    set_duration_label: function (label, duration) {
```
tchx84: There is nothing really card specific with this method right? Wouldn't it make more sense as helper in utils.js? Specially since we only use it for one specific card.

What do you think?

js/app/modules/card/list.js
```diff
@@ -46,6 +47,13 @@ const List = new Module.Class({
         if (this.navigation_context)
             this.set_label_or_hide(this._navigation_context_label, this.navigation_context);
         this.set_thumbnail_frame_from_model(this._thumbnail_frame);
+
+        // FIXME: It's not clear what should go in the context label if model
+        // is not a video. This may need to be changed later on once we clarify
+        // that with other presets.
+        if (this.model instanceof Eknc.VideoObjectModel) {
```
tchx84: Why don't we just check for the duration property? I mean, instead of making it specific for Video, I think we plan to add audio as well, so the more flexible the better.

js/app/modules/card/list.js
```diff
@@ -1,5 +1,6 @@
 // Copyright 2015 Endless Mobile, Inc.
 
+const Eknc = imports.gi.EosKnowledgeContent;
```
tchx84: no longer necessary!

PR #1294, opened 2016-12-14
T13418 kill domain v2

mattdangerw:
Don't merge this till tests have been fixed!!!

Here's some API changes I'm planning to make the rest of the C port easier and generally clean up our content API a bit.

Plenty of different ways we could go about this, so putting this up quickly without looking at tests (or testing much) to get feedback.

Major changes
 - Kill domain object v2
 - Application calls downloader directly (needed to port domain to C but leave downloader in js for the time being)
 - Remove broken command line option setting datadir
 - Some general cleanups

Probably a bit easier to go commit by commit. The diff on domain is kinda terrible, but its basically just moving things around.

### DIFF
--- a/js/app/application.js
+++ b/js/app/application.js
@@ -1,7 +1,9 @@
 imports.gi.versions.WebKit2 = '4.0';
 
+const Eknc = imports.gi.EosKnowledgeContent;
 const Endless = imports.gi.Endless;
 const EvinceDocument = imports.gi.EvinceDocument;
+const Format = imports.format;
 const Gdk = imports.gi.Gdk;
 const Gettext = imports.gettext;
 const Gio = imports.gi.Gio;
@@ -13,6 +15,7 @@ const System = imports.system;
 const Actions = imports.app.actions;
 const Config = imports.app.config;
 const Dispatcher = imports.app.dispatcher;
+const Downloader = imports.search.downloader;
 const Engine = imports.search.engine;
 const Knowledge = imports.app.knowledge;
 const ModuleFactory = imports.app.moduleFactory;
@@ -142,8 +145,10 @@ const Application = new Knowledge.Class({
         if (engine instanceof MoltresEngine.MoltresEngine)
             return;
 
+        this._check_for_update();
         try {
-            engine.update_and_preload_default_domain();
+            let shards = engine.get_default_domain().get_shards();
+            Eknc.default_vfs_set_shards(shards);
         } catch (e if e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND)) {
             // No content. If updates are pending then show a nice dialog
             let subs = engine.get_default_domain().get_subscription_entries();
@@ -159,6 +164,36 @@ const Application = new Knowledge.Class({
             }
             // if no updates pending, then proceed and let Xapian queries fail
         }
+    },
+
+    _check_for_update: function () {
+        if (GLib.getenv('EKN_DISABLE_UPDATES'))
+            return;
+
+        let engine = Engine.get_default();
+        let downloader = Downloader.get_default();
+        let subs = engine.get_default_domain().get_subscription_entries();
+        subs.forEach(function (entry) {
+            let id = entry.id;
+
+            if (entry.disable_updates)
+                return;
+
+            // Synchronously apply any update we have.
+            downloader.apply_update(id, null, (downloader, result) => {
+                downloader.apply_update_finish(result);
+
+                // Regardless of whether or not we applied an update,
+                // let's see about fetching a new one...
+                downloader.fetch_update(id, null, (downloader, result) => {
+                    try {
+                        downloader.fetch_update_finish(result);
+                    } catch(e) {
+                        logError(e, Format.vprintf("Could not update subscription ID: %s", [id]));
+                    }
+                });
+            });
+        });
     },
 
     vfunc_startup: function () {
--- a/js/app/historyStore.js
+++ b/js/app/historyStore.js
@@ -262,10 +262,10 @@ const HistoryStore = new Lang.Class({
     // the same after a link click, factoring out this common function. When we
     // diverge in future interactions we should revisit this decomposition.
     show_ekn_id: function (ekn_id) {
-        Engine.get_default().get_object_by_id(ekn_id, null, (engine, task) => {
+        Engine.get_default().get_object(ekn_id, null, (engine, task) => {
             let model;
             try {
-                model = engine.get_object_by_id_finish(task);
+                model = engine.get_object_finish(task);
             } catch (error) {
                 logError(error);
                 return;
@@ -295,9 +295,9 @@ const HistoryStore = new Lang.Class({
     },
 
     load_dbus_item: function (ekn_id, query, timestamp) {
-        Engine.get_default().get_object_by_id(ekn_id, null, (engine, task) => {
+        Engine.get_default().get_object(ekn_id, null, (engine, task) => {
             try {
-                let model = engine.get_object_by_id_finish(task);
+                let model = engine.get_object_finish(task);
                 if (model instanceof Eknc.ArticleObjectModel) {
                     this.set_current_item_from_props({
                         page_type: Pages.ARTICLE,
--- a/js/search/domain.js
+++ b/js/search/domain.js
@@ -230,7 +230,7 @@ const Domain = new Lang.Class({
         return false;
     },
 
-    get_object_by_id: function (id, cancellable, callback) {
+    get_object: function (id, cancellable, callback) {
         let task = new AsyncTask.AsyncTask(this, cancellable, callback);
         task.catch_errors(() => {
             let [hash] = Utils.components_from_ekn_id(id);
@@ -245,7 +245,7 @@ const Domain = new Lang.Class({
         return task;
     },
 
-    get_object_by_id_finish: function (task) {
+    get_object_finish: function (task) {
         return task.finish();
     },
 
@@ -295,7 +295,7 @@ const Domain = new Lang.Class({
         return task.finish();
     },
 
-    get_objects_by_query: function (query_obj, cancellable, callback) {
+    get_objects_for_query: function (query_obj, cancellable, callback) {
         let task = new AsyncTask.AsyncTask(this, cancellable, callback);
         task.catch_errors(() => {
             let domain_params = this._get_domain_query_params();
@@ -314,8 +314,8 @@ const Domain = new Lang.Class({
 
                 AsyncTask.all(this, (add_task) => {
                     json_ld.results.forEach((result) => {
-                        add_task((cancellable, callback) => this.get_object_by_id(result, cancellable, callback),
-                                 (task) => this.get_object_by_id_finish(task));
+                        add_task((cancellable, callback) => this.get_object(result, cancellable, callback),
+                                 (task) => this.get_object_finish(task));
                     });
                 }, cancellable, task.catch_callback_errors((source, resolve_task) => {
                     let results = AsyncTask.all_finish(resolve_task);
@@ -326,7 +326,7 @@ const Domain = new Lang.Class({
         return task;
     },
 
-    get_objects_by_query_finish: function (task) {
+    get_objects_for_query_finish: function (task) {
         return task.finish();
     },
 });
--- a/js/search/domain.js
+++ b/js/search/domain.js
@@ -8,7 +8,6 @@ const Gio = imports.gi.Gio;
 const Lang = imports.lang;
 
 const AsyncTask = imports.search.asyncTask;
-const Downloader = imports.search.downloader;
 const Utils = imports.search.utils;
 
 // This hash is derived from sha1('link-table'), and for now is the hardcoded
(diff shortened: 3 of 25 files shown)


js/search/engine.js
```diff
@@ -67,17 +67,22 @@ const Engine = Lang.Class({
...
      *              The function will be called with the engine and a task object,
      *              as parameters. The task object can be used with
      *              <get_object_by_id_finish> to retrieve the result.
      */
-    get_object_by_id: function (id, cancellable, callback) {
+    get_object_by_id: function (id, app_id, cancellable, callback) {
         let task = new AsyncTask.AsyncTask(this, cancellable, callback);
         task.catch_errors(() => {
+            if (!app_id)
+                app_id = this.default_app_id;
```
magcius: perhaps this should be in `_get_domain`?

js/app/historyStore.js
```diff
@@ -262,7 +262,7 @@ const HistoryStore = new Lang.Class({
     // the same after a link click, factoring out this common function. When we
     // diverge in future interactions we should revisit this decomposition.
     show_ekn_id: function (ekn_id) {
-        Engine.get_default().get_object_by_id(ekn_id, null, (engine, task) => {
+        Engine.get_default().get_object_by_id(ekn_id, null, null, (engine, task) => {
```
magcius: the `null, null` is a bit too microsoft-y for my taste. you can easily do `let cancellable = null;` as i commonly do to document this param, but app_id being null falling back to the default app id just seems.... weird? i really dont know what to do though

mattdangerw: Yeah I think with all the C writing I'm getting too used to passing three nulls and a couple 0s.

I'm fine with ditching this last commit entirely, if we can't get it the way we want. Its just our engine API is a bit confusing as is...

Could pass the `app_id` at the end of the arg list here optionally, but it seemed confusing and pattern breaking to put an argument after the cancellable and callback.

js/app/application.js
```diff
@@ -142,12 +144,13 @@ const Application = new Knowledge.Class({
...
+        this._check_for_update();
         try {
-            engine.update_and_preload_default_domain();
+            engine.preload_default_domain_shards();
         } catch (e if e.matches(Gio.IOErrorEnum, Gio.IOErrorEnum.NOT_FOUND)) {
             // No content. If updates are pending then show a nice dialog
-            let subs = engine.get_default_domain().get_subscription_entries();
-            if (subs.some(entry => !entry.disable_updates)) {
+            let entry = engine.get_default_domain().get_subscription_entry();
+            if (!entry.disable_updates) {
```
mattdangerw: I can leave it in. Just felt weird to have the update/downloader part of the application.js clearly written for one subscription, but this part written for multi subscription.

js/app/historyStore.js
mattdangerw: I guess on the more verbose side there's
get_object_by_id_for_default_app
And
get_object_by_id_for_app

But that's quite a mouthful

mattdangerw: Maybe just
get_object_for_default_app
get_object_for_app

ptomato: Not sure if you want this, as it seems pretty confusing to me, but spidermonkey 31 now allows parameters with default values to come before parameters without them: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

js/search/domain.js
```diff
@@ -22,7 +20,6 @@ const LINK_TABLE_ID = '4dba9091495e8f277893e0d400e9e092f9f6f551';
  */
 const Domain = new Lang.Class({
     Name: 'Domain',
-    Abstract: true,
```
ptomato: Do you really want to remove the `get_domain_impl()` stuff? Even if we drop the code for Domain v2, doesn't this mean we'd have to separate it all back out again if we wanted Domain v4? Or does Flatpak render that totally unnecessary?

mattdangerw: Have some semi formed thoughts here.

With the flatpak sandbox I think we could try to have a EKN_VERSION strongly coupled to a runtime version, so that all apps built with version say 3.5 of our runtime *have* to have content with EKN_VERSION 4. But particularly with global search and perhaps more generally, I can see us getting into a position where our content library needs to support multiple on disk formats of our offline content again.

However, a new version of our offline format would probably be coming from new requirements on our content, and I a little afraid of trying to predict the future here. Our current API surface on domains is fairly big
```
load_sync
get_domain_query_params
get_fixed_query[_finish]
resolve_xapian_result[_finish]
get_objects_by_query[_finish]
get_object_by_id[_finish]
check_for_updates
get_subscription_entries
```
If we have some new assumption-breaking content requirement that requires a new on disk format, how confident are we that API holds somewhat stable? And that domain continues to be the one stop shop for handle on disk format dependencies? Taking the full abstract plus subclass approach is gonna be a lot of boilerplate in C, which is totally fine that's what I signed up for, but I don't want us carry a lot of code we don't currently need unless we are reasonably confident in its longevity.

And of course we could keep around just the `get_domain_impl()` and a switch that errors if not version 3 if we want just that piece.

@magcius might have some thoughts here...

js/app/application.js
mattdangerw: How about I rewrite the update function to use `get_subscription_entries` when I move it to application (and update multiple subscriptions if there were any). Then we can keep `get_subscription_entries` as the exposed API from domain and extend it to handle such internally down the line.

mattdangerw:
Revised. Still no testing work done, waiting for more of an ack on approach.

Changed the last commit reworking engine API to avoid the weird double null, which I think is an improvement. Involves some bigger function renames though.

Then there's that big comment above. Totally fine going back to the abstract class if we want, just wanted to explain my thinking.

js/search/domain.js
magcius: Agreed with @mattdangerw's comments above -- the port to C would have rather large changes for implementation, and I see no reason to carry that over -- we can reintroduce a V4 Domain with an abstract class implementation later.

I'm not convinced that EKN_VERSION should be tied to runtime version -- I can see it argued both ways. We might want to upgrade e.g. Prensa Libre to a new version for a new module without having to stop old versions of the app from working, doing large content migration, or changing subscriptions. But that's probably for another time.

js/search/domain.js
```diff
@@ -217,23 +180,55 @@ const Domain = new Lang.Class({
...
+            return new EosShard.ShardFile({
+                path: file.get_path(),
+            });
+        });
         // Don't allow init() to be cancelled; otherwise,
         // cancellation will spoil the object for future use.
         Eknc.utils_parallel_init(this._shards, 0, null);
 
+        // Append shards to default EknVfs for ekn:// uri to work
+        Eknc.default_vfs_set_shards(this._shards);
```
magcius: Loading a random domain shouldn't cause it to stomp over the VFS. Please leave this in engine.

ptomato:
Approach looks fine to me. You asked specifically for comments about the mass rename in the last commit:
- `get_objects_by_query` to `get_objects_for_query` — seems fine to me either way, slight preference for the old name since you are getting objects BY the query FOR the default app, but you could make just as strong a case the other way.
- `get_object_by_id` to `get_object_for_default_app` — I'm not so fond of this one, it seems to me the default app should be implicit. This seems like a lot of typing only to confirm that you want the default. Maybe given `get_object_for_app` just call it `get_object`?
- `get_object_for_app` — seems fine

mattdangerw:
> This seems like a lot of typing only to confirm that you want the default. Maybe given get_object_for_app just call it get_object

I guess I just wanted the word default in anything requiring the default app id to be set, but fine with the shorter version. I'll just document that it needs the property set.

> get_objects_by_query to get_objects_for_query — seems fine to me either way, slight preference for the old name since you are getting objects BY the query FOR the default app

Well not necessarily, the query object can contain an app id, default will only be used if no app id was set in the query. "This method gets the object models for a given query object" feels like better english than "this method gets the object models by a given query object", but not a big deal either way.

js/search/domain.js
magcius: There are some cases, like the xapian DB or link tables, where we explicitly have SHA collisions between shards. We don't load those from the VFS, so it should be fine, but it feels weird to me to do that.

mattdangerw: Cool. I'll just rename stuff to make explicit for now.

js/app/application.js
```diff
@@ -161,6 +166,37 @@ const Application = new Knowledge.Class({
...
+    _check_for_update: function () {
+        if (GLib.getenv('EKN_DISABLE_UPDATES'))
+            return;
+
+        let engine = Engine.get_default();
+        let downloader = Downloader.get_default();
+        let subs = engine.get_default_domain().get_subscription_entries();
+        subs.forEach(function (entry) {
+            let id = entry.id;
+            let disable_updates = !!entry.disable_updates;
```
ptomato: Doesn't seem like this is used more than once below, just replace it with
```js
if (entry.disable_updates)
    return;
```

js/search/engine.js
```diff
@@ -57,48 +56,76 @@ const Engine = Lang.Class({
...
     },
 
     /**
-     * Function: get_objects_by_query
+     * Function: get_object
+     *
+     * Asynchronously fetches an object with ID for a given application.
+     *
+     * Parameters:
+     *   id - The unique ID for this object, of the form ekn://domain/sha
```
ptomato: form should be `ekn:///sha`?

ptomato:
Merged the top two commits to master.

mattdangerw:
Revised

ptomato:
I'm getting an error that I don't get on master, here's the full backtrace. (Admittedly this is running an app from jhbuild, so it's certainly possible that something is messed up about my setup.)
```
(com.endlessm.prensa_libre.es_GT:1082): Gjs-WARNING **: JS ERROR: Failed to load sets from database: TypeError: record is null
Domain<.get_object/<@resource:///com/endlessm/knowledge/js/search/domain.js:238:17
AsyncTask<.catch_errors@resource:///com/endlessm/knowledge/js/search/asyncTask.js:65:13
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Domain<.get_object@resource:///com/endlessm/knowledge/js/search/domain.js:235:9
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Domain<.get_objects_for_query/</</</</<@resource:///com/endlessm/knowledge/js/search/domain.js:317:61
all/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:181:32
all@resource:///com/endlessm/knowledge/js/search/asyncTask.js:181:5
Domain<.get_objects_for_query/</<@resource:///com/endlessm/knowledge/js/search/domain.js:320:1
AsyncTask<.catch_callback_errors/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:88:17
AsyncTask<._callback_in_idle/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:134:13
@<command line>:10:18
--- Called from: ---
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
all/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:177:9
Domain<.get_objects_for_query/</</</<@resource:///com/endlessm/knowledge/js/search/domain.js:317:25
Domain<.get_objects_for_query/</</<@resource:///com/endlessm/knowledge/js/search/domain.js:316:21
all@resource:///com/endlessm/knowledge/js/search/asyncTask.js:163:5
Domain<.get_objects_for_query/</<@resource:///com/endlessm/knowledge/js/search/domain.js:320:1
AsyncTask<.catch_callback_errors/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:88:17
AsyncTask<._callback_in_idle/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:134:13
@<command line>:10:18
--- Called from: ---
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Domain<.get_objects_for_query/</<@resource:///com/endlessm/knowledge/js/search/domain.js:320:1
AsyncTask<.catch_callback_errors/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:88:17
AsyncTask<._callback_in_idle/<@resource:///com/endlessm/knowledge/js/search/asyncTask.js:134:13
@<command line>:10:18
--- Called from: ---
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Engine<.get_objects_for_query/</do_query@resource:///com/endlessm/knowledge/js/search/engine.js:152:17
Engine<.get_objects_for_query/<@resource:///com/endlessm/knowledge/js/search/engine.js:173:17
AsyncTask<.catch_errors@resource:///com/endlessm/knowledge/js/search/asyncTask.js:65:13
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Engine<.get_objects_for_query@resource:///com/endlessm/knowledge/js/search/engine.js:143:9
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Controller<.initialize_set_map@resource:///com/endlessm/knowledge/js/app/interfaces/controller.js:98:1
Buffet<.make_ready@resource:///com/endlessm/knowledge/js/app/modules/controller/buffet.js:44:9
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Application<._ensure_controller@resource:///com/endlessm/knowledge/js/app/application.js:246:13
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
Application<.vfunc_activate@resource:///com/endlessm/knowledge/js/app/application.js:225:9
wrapper@resource:///org/gnome/gjs/modules/lang.js:178:13
@<command line>:10:18
```

ptomato:
Determined that the error was pre-existing, and showed up by chance on my tests with this branch, so that can go to another ticket. **CR pass**

PR #1295, opened 2016-12-15
T14576

tchx84:
https://phabricator.endlessm.com/T14576

### DIFF
--- a/ekncontent/overrides/EosKnowledgeContent.js
+++ b/ekncontent/overrides/EosKnowledgeContent.js
@@ -89,9 +89,10 @@ function _init() {
     });
     define_property(Eknc.ArticleObjectModel, 'table-of-contents', {
         get: function () {
-            let toc = this.get_table_of_contents().deep_unpack();
+            let toc = this.get_table_of_contents();
             if (!toc)
                 return [];
+            toc = toc.deep_unpack();
             toc.forEach(item => {
                 for (let prop in item)
                     item[prop] = item[prop].deep_unpack();


ekncontent/overrides/EosKnowledgeContent.js
```diff
@@ -89,7 +89,9 @@ function _init() {
     });
     define_property(Eknc.ArticleObjectModel, 'table-of-contents', {
         get: function () {
-            let toc = this.get_table_of_contents().deep_unpack();
+            let toc = this.get_table_of_contents();
+            if (toc)
+                toc = toc.deep_unpack();
```
mattdangerw: Just move this line under the if statement below, and you don't need this second if right?

mattdangerw:
Thanks! Thought this mighta been my fault.

One small comment

ekncontent/overrides/EosKnowledgeContent.js
tchx84: I thought those were different checks, I mean, toc could be null but it could also be something "empty" (but we only know that after we unpack it?) , so If I get rid of the first  "if" I would only be checking the first null case, but not for the second case.

mattdangerw: Well the gvariant there is gauranteed to be a array. So if its not null, the case you are talking about is where its any empty array right? If that case the forEach below would be a no-op so an early return is fine by not necessary.

Even if we do need extra check for a reason I am missing I'd write everything with early retruns
```
if (!toc)
  return [];
toc = toc.deep_unpack();
if (!toc)
  return [];
```

But do think we can just ditch the second if.

mattdangerw:
CR pass

PR #1296, opened 2016-12-17
eknvfs: Fix for missing schemes

magcius:
Some code, like file-roller, first tries to open any inputs with
g_file_new_for_uri, and then if that returns NULL, uses
g_file_new_for_path.

If we try to open a string without a scheme, we'll segfault here when we
look it up in the hash table. Just do a quick test first.

https://phabricator.endlessm.com/T14613


xjuan:
CR Pass

PR #1297, opened 2016-12-20
EknMediaBin: ported to playbin3

xjuan:
With playbin3 we always get eos event, even if we initially set the state to paused
which is needed to show the first frame of the video and make the widget request
the video size.

https://phabricator.endlessm.com/T14316


rmacqueen:
Looks good. Added a commit on top to revert the workaround we had in place before this.

CR PASS

PR #1300, opened 2016-12-22
ekncontent: Add EkncQueryResults object

ptomato:
This object is a replacement for the property bag that's returned in the
JS version of Engine.get_objects_for_query(). It is immutable and also
does not expose any constructor API other than g_object_new.

### DIFF
--- a/ekncontent/ekncontent/eknc-query-results.c
+++ b/ekncontent/ekncontent/eknc-query-results.c
@@ -1,0 +1,188 @@
+/* Copyright 2016 Endless Mobile, Inc. */
+
+#include "eknc-query-results.h"
+#include "eknc-content-object-model.h"
+
+/**
+ * SECTION:query-results
+ * @title: Query Results
+ * @short_description: Results and metadata from Xapian
+ *
+ * The #QueryResults class is returned from search operations.
+ * It provides a list of search results, as well as metadata about the search,
+ * such as the number of total search results.
+ *
+ * This class has no functionality, but is just a bag of properties.
+ * Instances are immutable after creation and all properties must be passed in
+ * on construction.
+ *
+ * You are not intended to create instances of this class; it is returned as
+ * a result of search operations.
+ */
+struct _EkncQueryResults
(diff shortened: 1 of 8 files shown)


ekncontent/ekncontent/eknc-query-results.c
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * eknc_query_results_new_for_testing:
+ *
+ * For testing within eos-knowledge-content only.
+ *
+ * Returns: (transfer full):
+ */
+EkncQueryResults *
+eknc_query_results_new_for_testing (void)
```
mattdangerw: If we take in a GSList here of content object, would that marshal properly from gjs? With `Eknc.QueryResults.new_for_testing([model1, model2])`

Seems like that would be nice just to be able to keep the testing contents local to the testing file.

ekncontent/ekncontent/eknc-query-results.c
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * eknc_query_results_get_upper_bound:
+ * @self: the #EkncQueryResults
+ *
+ * See #EkncQueryResults:upper-bound.
+ *
+ * Returns: the upper bound on the number of results
+ */
+gint
+eknc_query_results_get_upper_bound (EkncQueryResults *self)
```
mattdangerw: Fine with this, but not really needed as the property getter will work fine. Haven't been adding getters except for when needed for introspection workarounds for the content object models.

ptomato: If it's already written and you don't mind, then it may as well stay...

mattdangerw: Sounds good

mattdangerw:
One last thing, this should be included in the `eos-knowledge-content.h` header file.

ptomato:
Revised.

mattdangerw:
Thanks! CR pass

PR #1301, opened 2016-12-22
debian/control: added libgstreamer-plugins-base1.0-dev dep

xjuan:
https://phabricator.endlessm.com/T14316


ptomato:
**CR pass**

PR #1302, opened 2016-12-23
T14104

tchx84:
https://phabricator.endlessm.com/T14104

### DIFF
--- a/js/app/modules/arrangement/constrained.js
+++ b/js/app/modules/arrangement/constrained.js
@@ -1,0 +1,127 @@
+// Copyright 2016 Endless Mobile, Inc.
+
+/* exported Constrained */
+
+const Emeus = imports.gi.Emeus;
+const Gtk = imports.gi.Gtk;
+const Lang = imports.lang;
+
+const Arrangement = imports.app.interfaces.arrangement;
+const Card = imports.app.interfaces.card;
+const Module = imports.app.interfaces.module;
+
+const _DEFAULT_CARD_COUNT = 10;
+const _DEFAULT_COLUMNS = 5;
+const _DEFAULT_ROWS = 6;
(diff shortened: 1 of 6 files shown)


js/app/modules/arrangement/constrained.js
```diff
@@ -0,0 +1,124 @@
...
+        let child = card.get_parent();
+        child.remove(card);
+        this.remove(child);
+    },
+
+    /**
+     * Method: get_description
+     * Get Emeus VFL description for this arrangement.
+     */
+    get_description: function () {
```
xjuan: any reason not to use a property instead of a function here?

js/app/modules/arrangement/constrained.js
```diff
@@ -0,0 +1,124 @@
...
+            'H:|[view3][view5(view4 * 2)][view6(view4)]|',
+            'H:|[view8(view4)][view9(view4)][view5][view7(view4)]|',
+            'V:|[view0(view4 * 3)][view3(view4 * 2)][view8(view4)]|',
+            'V:[view3][view9(view4)]|',
+            'V:|[view0][view4][view5(view4 * 2)]|',
+            'V:|[view1(view4 * 2)][view2(view4 * 2)][view6(view4)][view7(view4)]|',
+        ];
+    },
+
+    _setup_constraints: function () {
```
xjuan: If you do not want to have a property for the VFL, then we could make this method public and have a string parameter for it and let subclasses specify their own VFL when overriding set_models()

tchx84: As we agreed on the hangout call: a) I can't leave it as a property because we don't want to expose the VFL yet, and b) that override of set_models hides implementation details, so we don't want other classes to re-implement it.

PR #1304, opened 2016-12-27
KnowledgeDocument: use EknMediaBin to playback audio links

xjuan:
https://phabricator.endlessm.com/T14205

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -188,7 +188,7 @@ PKG_CHECK_MODULES([EOS_KNOWLEDGE_PRIVATE], [
     glib-2.0
     gobject-2.0
     gio-2.0
-    gtk+-3.0 >= 3.12
+    gtk+-3.0 >= 3.22
     webkit2gtk-4.0
     gstreamer-1.0 >= 1.10
     gstreamer-audio-1.0 >= 1.10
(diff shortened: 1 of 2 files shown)


js/app/modules/card/knowledgeDocument.js
```diff
@@ -219,6 +221,35 @@ const KnowledgeDocument = new Module.Class({
...
+                    if (GLib.uri_parse_scheme (uri) !== 'file')
+                        return false;
+
+                    let mediabin = new EosKnowledgePrivate.MediaBin({
+                        audioMode: true,
+                        uri: uri,
+                    });
+                    mediabin.show();
+
+                    let lightbox = new Lightbox.Lightbox({
```
tchx84: Would it be possible to use the existing MediaLightbox ?

I am not sure it is, considering the audio files are not ObjectModels (and don't go through the HistoryStore), but one thing I noticed is that we can open this new Lightbox and it would only cover the document area, then a user could click on something else and the lightbox would be inaccessible (still playing the audio).

xjuan: Right, once the audio files are in the shard we will be able to use a MediaLightBox

Th other option would be to pu the MediaBin in a popover but there is no easy way to get the link position from a PDFView (at least I havent found)

PR #1309, opened 2017-01-03
Moduleset fixes

ptomato:
https://phabricator.endlessm.com/T14821

### DIFF
--- a/eos-knowledge-lib.modules
+++ b/eos-knowledge-lib.modules
@@ -22,13 +22,6 @@ use the latest stable release branch from GNOME if you prefer that.
 
   <!-- SYSTEM DEPENDENCIES -->
 
-  <systemmodule id="libtool">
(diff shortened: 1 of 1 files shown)


eos-knowledge-lib.modules
```diff
@@ -199,6 +199,7 @@ use the latest stable release branch from GNOME if you prefer that.
       <dep package="gjs"/>
       <dep package="glib"/>
       <dep package="gtk+"/>
+      <dep package="emeus"/>
```
tchx84: The only thing missing was a dependency on meson system package (required for building emeus), after I added it manually, I was able to successfully follow all steps of the contributing file.

ptomato: Should the module type just be `meson` in that case? I don't have a meson system dependency here.

PR #1314, opened 2017-01-06
Changes to escola template based on Design QA

rmacqueen:
https://phabricator.endlessm.com/T14133

### DIFF
--- a/data/preset/escola.yaml
+++ b/data/preset/escola.yaml
@@ -17,22 +17,33 @@ defines:
       shortdef: 'Layout.Box(orientation: vertical)'
       slots:
         contents:
-        - shortdef: Layout.Box
+        - type: Layout.Overlay
           styles:
             - landing
           slots:
-            contents:
-            - shortdef: 'Banner.Dynamic(show-subtitle: true, mode: text, expand: true)'
-        - shortdef: 'ContentGroup.ContentGroup(expand: true)'
+            content:
+              shortdef: Layout.Box
+              styles:
+                - background
+            overlays:
+            - shortdef: 'Layout.Box(orientation: vertical, valign: center, vexpand: true)'
+              slots:
+                contents:
+                - shortdef: 'Banner.Dynamic(mode: text, expand: true)'
+                - shortdef: 'ContentGroup.StaticText(label: "TV ESCOLA")'
+        - shortdef: 'Layout.Box(orientation: vertical)'
           slots:
-            <<: *main-arrangement-slot
-            selection: Selection.AllSets
-        - shortdef: 'Layout.Box(orientation: vertical, halign: fill, hexpand: true)'
-          styles:
-            - footer
-          slots:
             contents:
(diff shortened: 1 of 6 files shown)


data/preset/escola.yaml
```diff
@@ -17,22 +17,33 @@ defines:
...
+            content:
+              shortdef: Layout.Box
+              styles:
+                - background
+            overlays:
+            - shortdef: 'Layout.Box(orientation: vertical, valign: center, vexpand: true)'
+              slots:
+                contents:
+                - shortdef: 'Banner.Dynamic(mode: text, expand: true)'
+                - shortdef: 'ContentGroup.StaticText(label: "TV ESCOLA")'
```
ptomato: This makes the preset unsuitable for any apps but TV Escola. Isn't the app title already part of the `Banner.Dynamic` above this label?

nedrichards: Yeah, escola was the generic version of capacitate that we wanted to use in the portal. Obviously we can fix this in post with another yaml but unexpected interactions may occur.

rmacqueen: > Isn't the app title already part of the Banner.Dynamic above this label?

Yes but this text, "TV ESCOLA" is not a title, nor a subtitle. It's just text that we want to include on all tv escola apps, and since we don't yet have a way of parameterizing yaml files, I have to just stick it in as a constant. We're not making any other non-tv-escola apps for 3.1 that use this template, so I figure by the time we need to use it for something else we'll have a way to parameterize the yaml

js/app/widgets/dynamicLogo.js
```diff
@@ -150,6 +150,10 @@ const DynamicLogo = new Knowledge.Class({
     _load_image: function () {
         try {
             let file = Gio.File.new_for_uri(this._image_uri);
+            if (!file.query_exists(null) && this.mode === 'full') {
+                this.mode = 'text';
+                return;
+            }
```
tchx84: I would really prefer to move this logic to the dynamic banner module, because in module we can tell in advance what is the proper mode to use in the widget.. In the module, we know if we have a real image (the uri lives there) and if we have a text (we extract the text there as well).

Deciding to fallback in this widget, only for 1 of the 3 modes we have, can be confusing.

data/preset/escola.yaml
ptomato: I disagree. In fact it's quite likely we'll be asked to make apps for 3.1.x point releases that use this template, and I don't want to backport the parametrization stuff if we can help it. I think this is simply a place where we should compromise on the design.

nedrichards: To clarify what I meant earlier - 'escola' was supposed to be the generic template that we could expose in the portal and allow other people to use. That was the whole reason we forked it from capacitate (since that had the logos in there).

So for our release by the end of march we either need 'another fork' or to have worked out how to parametarise that stuff.

PR #1316, opened 2017-01-09
Destroy lightbox widget when closing lightbox

rmacqueen:
Currently, if the lightbox widget is a video, it will
keep playing in the background after the lightbox has
closed. So we need to destroy the widget when lightbox
closes or when we switch to another piece of content.

https://phabricator.endlessm.com/T14931

### DIFF
--- a/js/app/modules/contentGroup/mediaLightbox.js
+++ b/js/app/modules/contentGroup/mediaLightbox.js
@@ -57,6 +57,10 @@ const MediaLightbox = new Module.Class({
     },
 
     _on_close: function () {
+        if (this.lightbox_widget instanceof EosKnowledgePrivate.MediaBin) {
+            this.lightbox_widget.stop();
+        }
+        this.lightbox_widget = null;
         Dispatcher.get_default().dispatch({
             action_type: Actions.LIGHTBOX_CLOSED,
         });
@@ -115,8 +119,13 @@ const MediaLightbox = new Module.Class({
         if (this._current_index === -1)
             return;
 
-        if (this.lightbox_widget)
+        if (this.lightbox_widget) {
+            if (this.lightbox_widget instanceof EosKnowledgePrivate.MediaBin) {
+                this.lightbox_widget.stop();
+            }
             this.drop_submodule(this.lightbox_widget);
+            this.lightbox_widget = null;
+        }
 
         let widget = this.create_submodule('card', {
             model: media_object


rmacqueen:
This sometimes gives me warnings in the terminal about gtk_revealer not being a revealer. Not sure if that's something we need to worry about or not. @mattdangerw

js/app/modules/contentGroup/mediaLightbox.js
```diff
@@ -57,6 +57,8 @@ const MediaLightbox = new Module.Class({
     },
 
     _on_close: function () {
+        this.lightbox_widget.destroy();
```
ptomato: I don't think you should call `destroy()` manually, since it will destroy the widget in C while still keeping a dangling JS object. Instead, `this.lightbox_widget = null` should be enough, since the setter in `Lightbox.Lightbox` will remove the widget from its container and the widget will get destroyed at the next garbage collection if nothing else is holding on to it.

js/app/modules/contentGroup/mediaLightbox.js
```diff
@@ -115,8 +117,11 @@ const MediaLightbox = new Module.Class({
         if (this._current_index === -1)
             return;
 
-        if (this.lightbox_widget)
+        if (this.lightbox_widget) {
             this.drop_submodule(this.lightbox_widget);
+            this.lightbox_widget.destroy();
```
ptomato: Same here.

js/app/modules/contentGroup/mediaLightbox.js
rmacqueen: Yea I tried that, and still it would play in the background after closing lightbox

xjuan: I guess it will play until it gets GC

You can stop playback with this.lightbox_widget.stop() and let the GC do its work eventually :S

ebassi: If you're keeping a reference to a widget in your JS code, you *must* connect to the `destroy` signal and release your reference. This should always happen.

ptomato: But if you release the reference here already, that shouldn't be necessary...

ebassi: You should still protect yourself from destruction coming from other layers — e.g. one of the ancestors being destroyed.

rmacqueen:
@ptomato ready again

js/app/modules/contentGroup/mediaLightbox.js
```diff
@@ -57,6 +57,10 @@ const MediaLightbox = new Module.Class({
     },
 
     _on_close: function () {
+        if (this.lightbox_widget instanceof EosKnowledgePrivate.MediaBin) {
```
ptomato: In the future consider `if ('stop' in this.lightbox_widget)`, I think it's more Javascript-y, but no big deal

PR #1320, opened 2017-01-10
T14128

rmacqueen:
https://phabricator.endlessm.com/T14128

### DIFF
--- a/data/css/escola.scss
+++ b/data/css/escola.scss
@@ -33,8 +33,8 @@ $display-font: 'Raleway Black' !default;
       &__logo {
         -EknDynamicLogo-sizing: "size-min";
         color: white;
-        min-height: 120px;
-        min-width: 400px;
+        min-height: 200px;
+        min-width: 600px;
         font-weight: normal;
         font-family: $display-font;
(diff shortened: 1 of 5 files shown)


data/css/library_list.scss
```diff
@@ -8,8 +8,8 @@
     &__logo {
       -EknDynamicLogo-sizing: "size-min";
       min-height: 250px;
+      min-width: 800px;
```
ptomato: This would prevent the app from working on composite (720x480)

data/css/escola.scss
```diff
@@ -33,8 +33,8 @@ $display-font: 'Raleway Black' !default;
       &__logo {
         -EknDynamicLogo-sizing: "size-min";
         color: white;
-        min-height: 120px;
-        min-width: 400px;
+        min-height: 200px;
+        min-width: 600px;
```
ptomato: If the logo is min 600px then I'm pretty sure this will also break composite 720x480.

ptomato:
I rebased this, but all the code that was modified is now gone.

PR #1321, opened 2017-01-11
Fix i18n of "See more" label

ptomato:
contentGroup.js needs to go in POTFILES.in, and the string should be in
double quotes because otherwise xgettext gets confused.

https://phabricator.endlessm.com/T14969


rmacqueen:
CR PASS

PR #1322, opened 2017-01-11
Fix content updates (eos3.1 branch only)

ptomato:
https://phabricator.endlessm.com/T14643

### DIFF
--- a/js/app/application.js
+++ b/js/app/application.js
@@ -180,18 +180,15 @@ const Application = new Knowledge.Class({
                 return;
 
             // Synchronously apply any update we have.
-            downloader.apply_update(id, null, (downloader, result) => {
-                downloader.apply_update_finish(result);
-
-                // Regardless of whether or not we applied an update,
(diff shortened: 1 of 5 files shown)


mattdangerw:
Approach looks fine to me. Can definitely apply the same set of changes to the new C version without much trouble.

A little curious where this regression actually came from. Cause this synchronous update apply is new right? How was this actually working before?

Also @magcius if you want, might be good to look over the downloader changes here.

I'll work on a master version of this change.

magcius:
Both of these patches look fine to me. I think the answer is that we always "won the race" before.

ptomato:
The commit that "broke" things (i.e. changed from a race condition to a situation where things were initialized in the wrong order, period) was 860285713257c1bf65b46f8f8bd014a051b59725, but as Jasper said, this was already present.

mattdangerw:
Alright lets go for it

CR pass

PR #1325, opened 2017-01-11
T14993

mattdangerw:
https://phabricator.endlessm.com/T14993

### DIFF
--- a/ekncontent/ekncontent/eknc-utils.c
+++ b/ekncontent/ekncontent/eknc-utils.c
@@ -453,3 +453,39 @@ eknc_default_vfs_set_shards (GSList *shards)
 
   return TRUE;
 }
+
+/**
+ * eknc_get_subscriptions_json:
+ * @app_id: knowledge app ID, such as "com.endlessm.health-es"
+ * @cancellable: (allow-none): optional #GCancellable object, %NULL to ignore.
+ * @error: #GError for error reporting.
+ *
+ * Reads the subscriptions json file for a given app_id.
+ *
+ * Returns: (transfer full): the JsonNode for the given app id
+ */
+JsonNode *
+eknc_get_subscriptions_json (const gchar *app_id,
+                             GCancellable *cancellable,
+                             GError **error)
+{
+  g_return_val_if_fail (app_id && *app_id, NULL);
+  g_return_val_if_fail (G_IS_CANCELLABLE (cancellable) || cancellable == NULL, NULL);
+  g_return_val_if_fail (error == NULL || *error == NULL, NULL);
+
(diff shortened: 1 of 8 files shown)


ekncontent/ekncontent/eknc-utils.c
```diff
@@ -453,3 +453,39 @@ eknc_default_vfs_set_shards (GSList *shards)
...
+ * @cancellable: (allow-none): optional #GCancellable object, %NULL to ignore.
+ * @error: #GError for error reporting.
+ *
+ * Reads the subscriptions json file for a given app_id.
+ *
+ * Returns: (transfer full): the JsonNode for the given app id
+ */
+JsonNode *
+eknc_get_subscriptions_json (const gchar *app_id,
+                            GCancellable *cancellable,
```
ptomato: nitpick: indent 1 more space

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -453,3 +453,39 @@ eknc_default_vfs_set_shards (GSList *shards)
...
+      return NULL;
+    }
+
+  g_autoptr(GFile) subscriptions_file = g_file_get_child (data_dir, "subscriptions.json");
+  g_autofree gchar *contents = NULL;
+  if (!g_file_load_contents (subscriptions_file, cancellable, &contents,
+                             NULL, NULL, error))
+    return NULL;
+
+  return json_from_string (contents, error);
```
ptomato: Does seem a bit wasteful to load this into a string in memory, convert to a `JsonNode`, convert back to a string on the JS side, then convert to a JS object... but that's out of the scope of this issue. I think it would be good to convert `JsonNode` to JS object directly in GJS, but I'm not sure whether it should be part of marshalling automatically, or should happen in some `toObject()` override.

mattdangerw: Well agreed generally that this is wasteful, but want to be careful about optimizing just for the sake of it and not in our real bottlenecks. This is a one off startup cost, and I suspect the blocking file i/o on startup will be much more noticable to the user than the time spent converting a small amount of json data between formats. Just guessing though, don't have real numbers here.

Keep in mind that once the downloader is moved into the content library this can become private and never called from js, so this is a temporary state.

ptomato: I do think that some sort of marshalling for json-glib would be good to add in GJS though...

mattdangerw: Ah yeah, didn't read closely enough. I'd be wary about landing a fix in e-k-l about this, but as a general introspection fix in gjs I think that makes sense.

PR #1327, opened 2017-01-13
Don't rely on integer overflow!

rmacqueen:
Previously we were implicitly relying on integer
overflow to convert -1 to a large int. This is dangerous
and was partly responsible for a serious bug we found on
ARM. Instead we should explicitly set the integer
to be the max int.

https://phabricator.endlessm.com/T15009

### DIFF
--- a/js/app/interfaces/card.js
+++ b/js/app/interfaces/card.js
@@ -311,7 +311,7 @@ const Card = new Lang.Interface({
     _count_set: function (set_obj, callback) {
         let query = Eknc.QueryObject.new_from_props({
             tags_match_any: set_obj.child_tags,
-            limit: -1,
+            limit: GLib.MAXUINT32,
         });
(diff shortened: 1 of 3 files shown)


js/app/interfaces/card.js
```diff
@@ -311,7 +311,7 @@ const Card = new Lang.Interface({
     _count_set: function (set_obj, callback) {
         let query = Eknc.QueryObject.new_from_props({
             tags_match_any: set_obj.child_tags,
-            limit: -1,
+            limit: GLib.MAXUINT32,
```
tchx84: Not working, somehow this is getting converted to -1 before we create the uri for xapian-bridge. I double checked, changing it to 1000 or 10000 then it works fine.

ebassi: This should have been fixed by my commit.

PR #1329, opened 2017-01-17
T14516

rmacqueen:
https://phabricator.endlessm.com/T14516

### DIFF
--- a/js/app/interfaces/documentCard.js
+++ b/js/app/interfaces/documentCard.js
@@ -71,4 +71,13 @@ const DocumentCard = new Lang.Interface({
      *
      */
     clear_content: Lang.Interface.UNIMPLEMENTED,
+
+    /**
+     * Method: set_active (is_active)
+     * Activate or deactivate the content being displayed on this page.
+     * An example use case would be the video player starting or stopping
+     * its playback.
+     *
+     */
+    set_active: Lang.Interface.UNIMPLEMENTED,
 });
--- a/js/app/modules/layout/articleStack.js
+++ b/js/app/modules/layout/articleStack.js
@@ -129,6 +129,7 @@ const ArticleStack = new Module.Class({
     _set_document_card: function (document) {
         this.visible_child = document;
         document.content_view.grab_focus();
+        document.set_active(true);
         this._clear_old_views();
     },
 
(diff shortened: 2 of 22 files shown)


js/app/modules/selection/search.js
```diff
@@ -26,7 +26,6 @@ const Search = new Module.Class({
         return Eknc.QueryObject.new_from_props({
             query: query,
             limit: limit,
-            tags_match_all: ['EknArticleObject'],
```
rmacqueen: Even though I am no longer restricting the search to only Articles, I still do not get video content returned in search results. i think maybe video titles aren't getting indexed? @ffarfan

js/app/interfaces/documentCard.js
```diff
@@ -71,4 +71,15 @@ const DocumentCard = new Lang.Interface({
...
+    /**
+     * Method: stop_content
+     * Stop the content for this article. Different from
+     * clear in the sense that we don't want to destroy
+     * the content itself, since it may be used again if user
+     * returns to article page. But for e.g. video and audio
+     * content we want to stop it playing.
+     *
+     */
+    stop_content: Lang.Interface.UNIMPLEMENTED,
```
xjuan: I am not really good at naming things but is there any way to make this method more generic?
So that it could eventually used for other things

maybe set_active (is_active)

rmacqueen: good idea. I had a feeling it was a bad name but couldn't think of anything else

rmacqueen:
The first four commits are bug fixes that can be merged right away. The next five are the changes necessary to support video content in templates A and B. They can also be merged though nothing will actually change yet until we tell the apps to start fetching video content from the engine, which happens in the final two commits. Those final two shouldn't be merged until we have indexed video titles in xapian: https://phabricator.endlessm.com/T15117

js/app/modules/layout/articleStack.js
```diff
@@ -239,8 +239,10 @@ const ArticleStack = new Module.Class({
 
     _on_history_changed: function () {
         let item = HistoryStore.get_default().get_current_item();
-        if (item.page_type !== Pages.ARTICLE)
+        if (item.page_type !== Pages.ARTICLE) {
+            this.get_children().forEach((view) => view.set_active(false));
```
xjuan: We also need to call view.set_active(true) the the current view dont we?
even if there is no current user for it

rmacqueen: well doing that now won't work because `set_active` plays the video and I don't want the video to start playing when the view loads - only when the user actually clicks play

rmacqueen:
updated commit message!

rmacqueen:
ready

xjuan:
CR Pass

PR #1330, opened 2017-01-17
EknMediaBin: print FPS in gst info level

xjuan:
https://phabricator.endlessm.com/T15106

### DIFF
--- a/lib/eosknowledgeprivate/ekn-media-bin.c
+++ b/lib/eosknowledgeprivate/ekn-media-bin.c
@@ -41,6 +41,8 @@
 
 #define INFO_N_COLUMNS           6  /* Number of info columns labels */
 
+#define FPS_WINDOW_SIZE          2  /* Window size in seconds to calculate fps */
+
 #define EMB_ICON_SIZE            GTK_ICON_SIZE_BUTTON
 
 #define EMB_ICON_NAME_PLAY       "ekn-media-bin-play"
@@ -113,7 +115,11 @@ typedef struct
   guint timeout_id;          /* Autohide timeout source id */
   gint  timeout_count;       /* Autohide timeout count since last move event */
 
-  guint tick_id;             /* Widget frame clock tick callback (used to update UI) */
+  guint  tick_id;           /* Widget frame clock tick callback (used to update UI) */
+  gint64 tick_start;
+  gint64 frames_window_start;
+  guint  frames_window;     /* Frames "rendered" in the last FPS_WINDOW_SIZE seconds window */
+  guint  frames_rendered;   /* Total frames "rendered" */
   GdkEventType pressed_button_type;
 
   gint video_width;
@@ -124,6 +130,7 @@ typedef struct
   GstElement *video_sink;    /* The video sink element used (glsinkbin or gtksink) */
   GstElement *vis_plugin;    /* The visualization plugin */
   GstBus     *bus;           /* playbin bus */
+  GstBuffer  *last_buffer;
 
   GstTagList *audio_tags;
   GstTagList *video_tags;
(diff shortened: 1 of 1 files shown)


lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,66 @@ ekn_media_bin_update_position (EknMediaBin *self)
   gtk_label_set_label (priv->audio_position_label, format_time (position));
 }
 
+static inline void
+log_fps (EknMediaBin *self)
+{
+  EknMediaBinPrivate *priv = EMB_PRIVATE (self);
+  GSource *source = g_main_current_source ();
```
ebassi: I'm not entirely sure it's a good idea to use the time of the current GSource, here. You're not in a GSource callback, you're in a tick callback provided by GtkWidget.

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,66 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
+      gst_sample_unref (sample);
+    }
+
+  if (priv->last_buffer == buffer)
+    return;
+
+  priv->last_buffer = buffer;
+  priv->frames_window++;
+
+  time = (g_source_get_time (source) - priv->tick_start) / 1000000.0;
```
ebassi: You should use `gdk_frame_clock_get_frame_time()` instead.

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,64 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
 {
+  static GstDebugLevel level = 0;
+
   ekn_media_bin_update_position ((EknMediaBin *)widget);
 
+  if (level == 0)
+    level = gst_debug_category_get_threshold (ekn_media_bin_debug);
+
+  if (level >= GST_LEVEL_INFO)
+    log_fps ((EknMediaBin *)widget);
```
ebassi: Minor nitpick: you could pass the frame clock to `log_fps()` here, instead of having to call `gtk_widget_get_frame_clock()`

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,64 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
+  if (priv->last_buffer == buffer)
+    return;
+
+  clock = gtk_widget_get_frame_clock (GTK_WIDGET (self));
+  priv->last_buffer = buffer;
+  priv->frames_window++;
+
+  time = (gdk_frame_clock_get_frame_time (clock) - priv->tick_start) / 1000000.0;
+
+  if (priv->playback_time == (gint)time)
```
ebassi: You're doing a lot of double/int conversions, and you're also using `guint`.

You should just use `gint64` everywhere and keep the value in milliseconds.

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,64 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
+    }
+
+  if (priv->last_buffer == buffer)
+    return;
+
+  clock = gtk_widget_get_frame_clock (GTK_WIDGET (self));
+  priv->last_buffer = buffer;
+  priv->frames_window++;
+
+  time = (gdk_frame_clock_get_frame_time (clock) - priv->tick_start) / 1000000.0;
```
ebassi: You're converting the frame time in seconds, here. Why?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,64 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
+
+  if (priv->playback_time == (gint)time)
+    return;
+
+  priv->playback_time = time;
+
+  if (priv->playback_time % FPS_WINDOW_SIZE == 0)
```
ebassi: You're computing the frames per second every two seconds?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,64 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
+    {
+      priv->frames_rendered += priv->frames_window;
+
+      GST_INFO ("FPS: %lf average: %lf",
+                priv->frames_window/(gdouble)FPS_WINDOW_SIZE,
```
ebassi: Coding style: missing spaces. Also, you converted everything to unsigned integers and now you're converting them out in doubles.

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1401,13 +1408,64 @@ ekn_media_bin_update_position (EknMediaBin *self)
...
+      priv->frames_window = 0;
+    }
+}
+
 static gboolean
 ekn_media_bin_tick_callback (GtkWidget     *widget,
                              GdkFrameClock *frame_clock,
                              gpointer       user_data)
 {
+  static GstDebugLevel level = 0;
```
ebassi: There's no need to initialise variables in static storage; they are guaranteed to be set to 0 by the C standard.

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -113,7 +115,11 @@ typedef struct
   guint timeout_id;          /* Autohide timeout source id */
   gint  timeout_count;       /* Autohide timeout count since last move event */
 
-  guint tick_id;             /* Widget frame clock tick callback (used to update UI) */
+  guint   tick_id;           /* Widget frame clock tick callback (used to update UI) */
+  guint64 tick_start;
+  guint   playback_time;
```
ebassi: You should use `gint64` for every time-related field.

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -113,7 +115,11 @@ typedef struct
...
+  guint   frames_rendered;   /* Total frames "rendered" */
+  guint   frames_window;     /* Frames "rendered" in the last 4s window */
```
ebassi: Here you say you're using a 4 seconds window, but below you check every 2 seconds.

PR #1331, opened 2017-01-17
EknMediaBin: do not use GST_SEEK_FLAG_TRICKMODE flag

xjuan:
Using GST_SEEK_FLAG_TRICKMODE flag while performing a seek makes
gst performance to drop considerably.

https://phabricator.endlessm.com/T15110


ptomato:
**CR pass**

To be merged Friday.

PR #1335, opened 2017-01-19
application: css file supersedes theme

mattdangerw:
If both a css/scss file and a theme are specified via the command
line, let the manual theme override the theme.

https://phabricator.endlessm.com/T15008


rmacqueen:
CR PASS

PR #1336, opened 2017-01-19
T15065 (part 1)

tchx84:
https://phabricator.endlessm.com/T15065

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -568,7 +568,7 @@ on_xapian_bridge_query_fixed (GObject *source,
 {
   EkncXapianBridge *bridge = EKNC_XAPIAN_BRIDGE (source);
   g_autoptr(GTask) task = user_data;
-  GError *error;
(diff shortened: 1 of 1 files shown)


ekncontent/ekncontent/eknc-domain.c
```diff
@@ -662,6 +662,10 @@ on_object_response (GObject *source,
   QueryState *state = g_task_get_task_data (task);
   GError *error = NULL;
 
+  // do nothing if we already returned error
+  if (g_task_had_error(task))
```
mattdangerw: Could this more generally just be `g_task_get_completed`?

mattdangerw: That way if we hit an early return that's not an error we are still covered.

tchx84: sure, will change it now!

mattdangerw:
CR pass! (not merging yet though)

mattdangerw:
Heh nice catch. Looks good! CR pass again

PR #1338, opened 2017-01-19
T15173

mattdangerw:
https://phabricator.endlessm.com/T15173

### DIFF
--- a/tools/ka-customizer
+++ b/tools/ka-customizer
@@ -182,7 +182,17 @@ app_chooser.connect('custom-item-activated', (chooser, app_id) => {
     selected_app_id = app_id;
 });
 grid.add(app_chooser);
-add_label('Update:');
+
+add_label('Framework version:');
+let framework_version_chooser = new Gtk.ComboBoxText({
+    halign: Gtk.Align.START,
+});
+framework_version_chooser.append('master', 'Nightly');
+framework_version_chooser.append('production', 'Production');
+framework_version_chooser.set_active_id('master');
+grid.add(framework_version_chooser);
+
+add_label('Update application and framework:');
 let update_switch = new Gtk.Switch({
     halign: Gtk.Align.START,
     active: true,
@@ -229,6 +239,10 @@ done_button.connect('clicked', () => {
     }
 
     let args = ['./ka-customizer-runner'];
+    let branch = '';
+    if (framework_version_chooser.get_active_id() === 'master')
+        branch = 'master';
+    args.push(branch);
     if (update_switch.get_active()) {
         show_message('Updating app and libraries, this may take a while...');
         args.push('update');
--- a/tools/ka-customizer
+++ b/tools/ka-customizer
@@ -200,7 +200,7 @@ add_label('Choose a stock theme (overrides application theme):');
 let stock_theme_chooser = new Gtk.ComboBoxText({
     halign: Gtk.Align.START,
 });
-['', 'default', 'encyclopedia', 'news', 'buffet', 'preset_a', 'preset_b'].forEach((item) => {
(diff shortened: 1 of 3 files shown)


tools/ka-customizer
```diff
@@ -190,7 +190,7 @@ add_label('Choose a stock theme (overrides application theme):');
 let stock_theme_chooser = new Gtk.ComboBoxText({
     halign: Gtk.Align.START,
 });
-['', 'default', 'encyclopedia', 'news', 'buffet', 'preset_a', 'preset_b'].forEach((item) => {
+['', 'default', 'buffet', 'course', 'encyclopedia', 'escola', 'library_list', 'news', 'picard', 'preset_a', 'preset_b'].forEach((item) => {
```
tchx84: can we exclude picard theme here? that theme is not really meant for apps, and could cause confusion?

tools/ka-customizer-runner
```diff
@@ -1,11 +1,14 @@
 #!/bin/bash -ex
 
+BRANCH=$1
```
tchx84: is this script supposed to be run by another script or humans? if humans, then we should probably add some usage message?

tools/ka-customizer-runner
```diff
@@ -1,11 +1,14 @@
...
+shift
+
 if [ "$1" = "update" ]; then
     shift
-    flatpak update com.endlessm.Platform master
-    flatpak update com.endlessm.Sdk master
+    flatpak update "com.endlessm.Platform//$BRANCH"
+    flatpak update "com.endlessm.Sdk//$BRANCH"
```
tchx84: I got an error when running the ka-customizer:

```
+ BRANCH=
+ shift
+ '[' update = update ']'
+ shift
+ flatpak update com.endlessm.Platform//
Looking for updates...
Updating: com.endlessm.Platform/x86_64/eos3.1 from eos-runtimes
./ka-customizer-runner: line 11:  1703 Segmentation fault      (core dumped) flatpak update "com.endlessm.Platform//$BRANCH"
```

The segmentation fault part is probably unrelated, but I think the BRANCH is not getting replaced?

tools/ka-customizer-installer
```diff
@@ -3,6 +3,9 @@
 APPDIR=$HOME/.local/share/applications
 BINDIR=$HOME/.local/bin
 
+if ! flatpak info --show-commit com.endlessm.Sdk eos3.1; then
+    flatpak install eos-runtimes com.endlessm.Sdk eos3.1
+fi
```
tchx84: shouldn't we check for and include com.endlessm.Platform eos3.1 as well?

btw, we could also use a variable here ($BRANCH?) so we can easily replace eos3.1 later on (or even have a installer.in version that does it automatically)?

tools/ka-customizer-runner
mattdangerw: Nope, not intended for human invocation. ka-customizer runs this, and it handles chaining commands easier done in bash than gjs.

Presumably if you are a human and comfortable with flatpak and the command line, you can just use the flatpak and knowledge app cli options directly.

tools/ka-customizer-runner
mattdangerw: BRANCH is intentionally left blank in that case, to run the default branch for the app. I don't think there's an easy way to explicitly get the platform version a flatpak app has been built against, so its not easy to write the branch out explicitly at the moment. See discussion here https://phabricator.endlessm.com/T15173#321665

tools/ka-customizer-installer
mattdangerw: Can check for Platform sure. This is meant to be run on non converted eos images by design where it should already be installed, but can't hurt to check.

I can add a variable here, though honestly I'm not sure what this looks like come 3.2. Do we need to make sure all of 3.1, 3.2 and master are installed? Will some apps still be targeting the 3.1 runtime? Kinda suspect so, but I'm not totally worried about future proofing this quite yet.

PR #1339, opened 2017-01-21
T14909

mattdangerw:
https://phabricator.endlessm.com/T14909


ekncontent/tests/ekncontent/testMediaObjectModel.js
```diff
@@ -15,12 +15,12 @@ describe('Media Object Model', function () {
...
     beforeEach(function () {
         jasmine.addMatchers(InstanceOfMatcher.customMatchers);
-        imageObject = Eknc.MediaObjectModel.new_from_json(MOCK_IMAGE_DATA);
+        imageObject = Eknc.MediaObjectModel.new_from_json(MOCK_MEDIA_DATA);
     });
 
     describe('type', function () {
-        it('should be an MediaObjectModel', function () {
-            expect(imageObject).toBeA(Eknc.MediaObjectModel);
+        it('should be an ContentObjectModel', function () {
```
ptomato: optional, pre-existing typo: an -> a

ekncontent/Makefile.am
```diff
@@ -163,8 +163,16 @@ javascript_tests = \
 	tests/ekncontent/testXapianBridge.js \
 	$(NULL)
 
+test_content = \
+	tests/testcontent/ekn/share/com.endlessm.fake_test_app.en/EKN_VERSION \
+	tests/testcontent/ekn/share/com.endlessm.fake_test_app.en/subscritions.json \
```
ptomato: typo: subscritions -> subscriptions (unlike the other typo, this one has consequences :smile:)

ekncontent/tests/ekncontent/testDatadir.js
```diff
@@ -6,9 +6,13 @@ describe('Data dir finder', function () {
...
+        let found_dir = Eknc.get_data_dir('com.endlessm.fake_test_app.en');
+        expect(found_dir.get_path()).toBeDefined();
+    });
+
     // FIXME: no easy way to test these without a way to mock file URIs
     // during tests.
-    it('returns a GFile pointing to a valid ekn data dir');
-    it('discovers data dirs inside the flatpak sandbox');
-    it('discovers data dirs in system data dirs');
+    xit('discovers data dirs inside the flatpak sandbox');
```
ptomato: This is fine, though not strictly necessary; `it('description')` without a function body is equivalent to `xit()`.

ekncontent/tests/ekncontent/testDomain.js
```diff
@@ -2,41 +2,75 @@ const Eknc = imports.gi.EosKnowledgeContent;
...
         });
     });
+
+    describe('get_object', function () {
+        beforeEach(function () {
+            domain.init(null);
+        });
+
+        it('returns an object model for an ID in our database', function (done) {
+            domain.get_object("ekn:///02463d24cb5690af2c8e898736ea8c80e0e77077", null, function (domain, result) {
```
ptomato: Where do these IDs come from? is it a requirement that they exist in the test shard?

mattdangerw: Yup. Not sure if there's a better way...

mattdangerw:
@ptomato so the shard here is the canary app from eos-content-factory with a modified db.json with only 4 articles.

Seemed like the easiest way to get a valid shard with xapian db and a links dictionary and all that. Trying to script all that setup would be doomed to failure IMO, getting eos-content-factory up and running is pretty tricky and changes pretty frequently.

I could put a comment roughly describing the process though.
```
// Build from the "canary app" in endless-content-factory with a modified db.json with only four articles.
// Use the kermit tool to list the articles and app IDs available in the shard.
```
and put that in the Makefile I guess?

ptomato:
That would be a shame though since external contributors wouldn't be able to modify it. It wouldn't be in violation of the license because the test code isn't part of the end-user software, but still not great.
Would it be possible to build a working shard with just a small JS script using xapian-glib and eos-shard directly?

mattdangerw:
> Would it be possible to build a working shard with just a small JS script using xapian-glib and eos-shard directly?

Probably, though there's some tricky details to get right such as getting the ids right for the xapian database and the link table dictionary, setting up the link table and dictionary in the first place.

I'd rather us take a step back there. I've had many conversations with many people about the problems of having an "open" knowledge framework without any way to make content shards with open source tools. We could look at some sort of tool that could take a a directory of html and metadata content and shardify it with a xapian database and links properly set up. Hopefully something that soma could call into directly but also suitable for other open tool chains or even manual content creation to plug into. Make that well documented and open.

Or something like that, I've got no strong opinion on the details but seems like we have a growing need there. Then we could plug into that tool here in the ekncontent test shard instead of trying to build some one off script that would probably take a decent amount of effort, and inevitably grow stale as soma evolves.

I'd love to look at this, though not totally confident I'll have the time, and least in the short term. Seems like it should be a separate issue, though I'm fine if we want to block this on that one. Though getting the extra security of some amount of coverage on engine/domain (so if someone pushes a segfault on domain initialization e.g., we don't deploy), would be nice. So I would vote for landing the improved coverage with a "magic shard," and solving the whole "how the hell do I make a shard for our framework without closed soma stuff" as soon as we have bandwidth.

mattdangerw:
https://phabricator.endlessm.com/T15214

mattdangerw:
Well besides the how do i make a shard discussion, CR ready again

ptomato:
OK. Some shard-building script is definitely part of the push to get this in a good shape for open-source, but judging from the response to the ticket you opened, apparently this is a sore topic, so we'll probably have to do it ourselves... I'm reluctantly OK with merging this in the meantime so that we can have a better functioning early warning system.

PR #1340, opened 2017-01-23
T14932

tchx84:
https://phabricator.endlessm.com/T14932


ebassi:
I like this, but maybe it should be called `Banner.AppMetadata`? Metadata is a bit generic.

ebassi:
Looks good.

PR #1342, opened 2017-01-23
control: content dev package depends on shard dev package

ptomato:
The content library now serves up shards as part of its public
api, so we should add this dependency
https://phabricator.endlessm.com/T15211


mattdangerw:
CR pass

PR #1345, opened 2017-01-25
T14391

mattdangerw:
https://phabricator.endlessm.com/T14391


js/app/promiseWrapper.js
```diff
@@ -0,0 +1,44 @@
+const Eknc = imports.gi.EosKnowledgeContent;
+const Gio = imports.gi.Gio;
+const Soup = imports.gi.Soup;
+
+function wrapPromise (obj, cancellable, async_name) {
```
mattdangerw: Moved away from supporting cancellation directly in promises for now. Reading more (here http://bluebirdjs.com/docs/api/cancellation.html for example) there's some tough cases where you have multiple consumers of the same promise.

Easier to just take in an optional GCancellable when calling into one of these `_promise` functions and handle passing the cancellable manually. In our case that's only with downloader.js

mattdangerw:
Need to actually test downloader more before merging, but should be ready for CR at least

mattdangerw:
Downloader does appear to be working fine with promises

js/app/courseHistoryStore.js
```diff
@@ -82,17 +82,12 @@ const CourseHistoryStore = new GObject.Class({
...
         });
-        Eknc.Engine.get_default().query(query, null, (engine, task) => {
-            let results;
-            try {
-                results = engine.query_finish(task);
-            } catch (error) {
-                logError(error);
-                return;
-            }
+        Eknc.Engine.get_default().query_promise(query).then((results) => {
```
ptomato: Style nit: when we were doing promises in the pantheon in node.js, we had settled on this style:
```js
Eknc.Engine.get_default().query_promise(query)
.then(results => {
    blah blah
})
.catch(error => {
    logError(error)
});
```
But since this is the first promise code we have, I'm open to setting new standards.

Also you may be able to do `.catch(logError)` which would be quite concise and clear IMO.

js/app/modules/selection/xapian.js
```diff
@@ -76,31 +76,20 @@ const Xapian = new Module.Class({
...
-            } catch (e) {
-                logError(e, 'Failed to load content from engine');
-                this._exception = e;
-                if (!this._error_state) {
-                    this._error_state = true;
-                    this.notify('in-error-state');
-                }
-                return;
-            }
+            models = results.models;
```
ptomato: nitpick: Combine these two lines with the previous two (or even better, `.then({models, upper_bound} => {...})` should work in GJS now)

js/app/widgets/eknWebview.js
```diff
@@ -147,13 +143,11 @@ const EknWebview = new Knowledge.Class({
...
         let components = Utils.components_from_ekn_id(id);
         if (components.length === 1) {
-            this._load_object(id, cancellable, (source, load_task) => {
-                try {
-                    let [stream, content_type] = load_task.finish();
-                    req.finish(stream, -1, content_type);
-                } catch (error) {
-                    fail_with_error(error);
-                }
+            this._load_object(id).then((results) => {
```
ptomato: Same here, `.then([stream, content_type] => { ... })`

tests/compliance.js
```diff
@@ -245,7 +245,7 @@ function test_selection_compliance (SelectionClass, setup=function () {}, extra_
             // Not all Selections might use the Xapian engine, but I can't think
             // of a better place to put this
             let engine = MockEngine.mock_default();
-            engine.query_finish.and.returnValue({ models: [] });
+            engine.query_promise.and.returnValue(Promise.resolve({ models: [] }));
```
ptomato: Nothing wrong with this, but out of curiosity, I suppose we could just leave this as-is and write an implementation of `MockEngine.query_promise()` to call `query_finish()`?

js/app/dominantColor.js
```diff
@@ -1,31 +1,19 @@
...
 const GdkPixbuf = imports.gi.GdkPixbuf;
 const Gio = imports.gi.Gio;
 
-const AsyncTask = imports.app.asyncTask;
 const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
 
 // Note: this is a temporary solution until the color extraction is done
 // at build-time. The use of this function should be limited to specific cases,
 // to avoid performance penalties.
-function get_dominant_color (model, cancellable, callback) {
```
ptomato: I think the rationale for wrapping it in the AsyncTask was to make it easier to refactor to async in the future.

js/app/downloader.js
```diff
@@ -36,74 +35,33 @@ const FileDownloader = new Lang.Class({
...
-            function bytes_read(stream, res) {
-                let bytes = stream.read_bytes_finish(res);
-                csum.update(bytes.get_data());
-
-                if (bytes.get_size() !== 0)
-                    continue_checksum();
-                else
-                    finish();
-            }
+            function continue_checksum () {
```
ptomato: Do you still need `continue_checksum` as a separate function?

js/app/downloader.js
```diff
@@ -114,44 +72,45 @@ const FileDownloader = new Lang.Class({
...
+                let csum = file_download_request.csum;
 
-    download_file_finish: function (task) {
-        return task.finish();
+                if (!csum)
+                    return true;
+                // XXX: If the file didn't checksum correctly, what should we do? Trash and re-download?
+                return this.verify_checksum_promise(tmp_file, csum, cancellable);
+            }).then(() => {
+                tmp_file.move(out_file, Gio.FileCopyFlags.OVERWRITE, cancellable, null);
```
ptomato: Consider wrapping the `move()` call in a promise as well, now that that's trivial

js/app/dominantColor.js
mattdangerw: Well fine with leaving sync for now? Could write this as a promise, but that seems over future proof as far as I know our only concrete plans with this are to move this to ingestion someday in which case there will be no need for async.

mattdangerw:
> Hmm, I didn't consider the case of a cancelled promise with two consumers. I suppose the GIO-specific thing to do would be to throw a Gio.ErrorEnum.CANCELLED in that case, although in the comments on the bluebird doc they seemed to object to that.

Yeah definitely worth thinking about more for a future gjs patch. You could have a tree of sorts of promises where everything was rigged to the same cancel, or one where each branch has its own. Probably want the ability to pass them down the chain, or make specific calls non cancellable if there are crucial for whatever reason. Definitely seemed worth some decent thought, there's been proposal (backed out on I think) for handling this withing promises. https://medium.com/@benlesh/promise-cancellation-is-dead-long-live-promise-cancellation-c6601f1f5082#.g7yohi2tu

I think for knowledge lib, just exposing the ability to pass a cancellable in on promise creation is perfect. That is completely flexible and doesn't get in the way of our code, which don't give two shits about cancellation.

But for gjs a can of worms worth cracking I think.

js/app/courseHistoryStore.js
mattdangerw: Fine with moving to that, it will explode the diff a bit with indentation changes.

tests/compliance.js
mattdangerw: Did that at first, fully embracing promises made for less code in mock engine so I preferred it. And the nice thing here is you can easily bounce between Promise.resolve and Promise.reject for the simple cases, as well as actually supporting more of an async flow (waiting some ticks on the mainloop e.g.) if you wanted to test application code with a real delay for some reason...

js/app/downloader.js
mattdangerw: We do still need to recursively call into it until we hit the end of our stream. Though if you see another way to unroll it let me know.

js/app/courseHistoryStore.js
mattdangerw: Ah misread, are you sure we decided on straight ruler and not
```
Eknc.Engine.get_default().query_promise(query)
    .then(results => {
        blah blah
    })
    .catch(error => {
        logError(error)
    });
```
Fine with whatever, but don't want to rearrange the style here more than once so I'll wait till consensus to touch this.

mattdangerw: https://github.com/airbnb/javascript/issues/216

mattdangerw:
All revised except for deciding on promise style i think...

js/app/courseHistoryStore.js
ptomato: Yes, we did decide on left-aligned. I can't find the place where we documented that, it was probably in an eos-sdk issue which are erased now.

I dig airbnb's point of staying close to `async`/`await` (they are landing in SpiderMonkey 52). Although that would mean more indentation churn as the above code would look like
```js
try {
    let results = await Eknc.Engine.get_default().query(query);
    blah_blah(results);
} catch(error) {
    logError(error);
}
```

mattdangerw: Alright lets go with what decided before. I'll make the change now.

mattdangerw:
> Promise.resolve().then(() => ...), is that idiomatic? I'd have thought new Promise((resolve, reject) => ...) for the first operation, but I haven't worked with a lot of Promise code lately.

They are slightly different, just a style I picked up from post I read I think. With `new Promise((resolve, reject) => { .... }` you have to be careful about wrapping any dangerous code in a try catch and rejecting on it. Starting with a `Promise.resolve().then(() => { .... }` you can just throw with the first chunk of your code same as any of the later `then`s.

> Have you thought about naming all the methods without the _promise on the end since you're not keeping the non-Promise versions?

I kind of like having the post fix so you know what you are getting. Think it helps readability. And things like `engine.query` are kept around, just not used.

mattdangerw:
> The proposal was just sugar to make the exception different from a normal exception, which I think we actually don't want to do in GJS unless they actually implement it in ECMA since it would be awkward with the way GIO works.

Yeah this makes sense to me. So in that case the only thing we might want to add to gjs in the future is some automatic promisification of _async/_finish functions that would remove the need for `promiseWrapper.js` in the future. Might need more introspection annotations thought?

js/app/downloader.js
```diff
@@ -204,51 +169,38 @@ const SubscriptionDownloader = new Lang.Class({
...
-
-            task.return_value(true);
-        }));
-
-        return task;
-    },
-
-    _download_new_shards_finish: function (task) {
-        return task.finish();
+            return Promise.all(downloads).then(function () { return true; });
```
ptomato: Simply `return Promise.all(downloads)`? Or is this intentional to keep the resolved promises inside this API?

js/app/courseHistoryStore.js
```diff
@@ -82,17 +82,14 @@ const CourseHistoryStore = new GObject.Class({
...
-                logError(error);
-                return;
-            }
+        Eknc.Engine.get_default().query_promise(query)
+        .then((results) => {
             if (results.models.length > 0) {
                 this.set_current_subset(results[0]);
             }
+        })
+        .catch(function (error) {
```
ptomato: nitpick: there's a lot of `.catch(function (error)` and `.catch(error =>` mixed

mattdangerw: Ill try to stick to function (error) where there is no need to bind this (usually the case).

js/app/downloader.js
mattdangerw: I think I was trying to keep the old API, where the true would be returned if an update was actually applied, false if no update was applied, and throwing if an error occurred.

Though I need to make sure to actually return that value in `fetch_update_promise` ill fix that up.

mattdangerw:
Thanks for review! nits fixed, distcheck passing, merging now

PR #1346, opened 2017-01-25
eknc-domain: make robust to parallel initialization

mattdangerw:
There's a few places where we had code which looked like
if (!g_file_exists)
  create_the_file;
We can run into problems here if two domains are racing to setup
directories and symlinks for the same subscriptions at the same
time. In that case you might make it through the existence check,
but by the time you get to actual file creation another domain
in another process has already completed the check.

I believe this is happening with multi-threaded make in our test
runner on ci at the moment, and could also happen when spinning up
the global search provider at the same time as an application.

Instead we move to
create_the_file;
if (!error matches G_IO_ERROR_EXISTS)
  error out;
This way if another domain has one the file creation race, we can
still proceed as normal.
https://phabricator.endlessm.com/T14909

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -215,7 +215,6 @@ eknc_parse_subscription_id (EkncDomain *self,
       return FALSE;
     }
   self->subscription_id = g_strdup (json_node_get_string (id_node));
-
   return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


mattdangerw:
Kind of a shot in the dark here. The error from `ci` looks like a race on directory creation and this code *should* fix that up. But everything still paces A-ok on my machine on master and this branch (even with multithreaded test runner), so not totally sure this will clear things up.

Either way it's probably a good change to get in.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -299,9 +307,14 @@ eknc_domain_initable_init (GInitable *initable,
   g_autoptr(GFile) bundle_subscriptions_dir = g_file_get_child (self->content_dir, "com.endlessm.subscriptions");
   g_autoptr(GFile) bundle_dir = g_file_get_child (bundle_subscriptions_dir, self->subscription_id);
 
-  if (!g_file_query_exists (subscription_dir, cancellable) &&
-      !g_file_make_directory_with_parents (subscription_dir, cancellable, error))
-      return FALSE;
+  g_file_make_directory_with_parents (subscription_dir, cancellable, error);
+  if (*error)
+    {
+      if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_EXISTS))
```
mattdangerw: It feels a little weird to be setting and unsetting the error passed in from the invoker here over the course of initialization.

The other option would be to use a local error and call g_propogate_error in all the early returns. That would be a little more verbose, I couldn't think of an actual practical reason where setting then unsetting the passed in error would bite us. But maybe makes sense to switch.

ptomato: This will crash if the caller of this function passes `NULL` in to indicate they aren't interested in the error. I think that's the reason for the `local_error` pattern.

mattdangerw:
Fixed up!

PR #1349, opened 2017-02-01
T15366

katelk28:
Card.LegacyPolaroid PDF icon and label styling

- Changed valign and vexpand properties in _legacyPolaroid.ui
- Deleted .CardLegacyPolaroid__icon padding in _mesh.scss
- Added .CardLegacyPolaroid__label padding in Card.LegacyPolaroid

https://phabricator.endlessm.com/T15366

### DIFF
--- a/data/css/_mesh.scss
+++ b/data/css/_mesh.scss
@@ -58,10 +58,6 @@ $display-font-composite: 'Roboto' !default;
 }
 
 .CardLegacyPolaroid {
-    &__icon {
-        padding-top: 10px;
-    }
-
     &__label {
         font-size: 0.9em;
         color: #666666;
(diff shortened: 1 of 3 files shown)


data/css/_mesh.scss
```diff
@@ -58,12 +58,9 @@ $display-font-composite: 'Roboto' !default;
 }
 
 .CardLegacyPolaroid {
-    &__icon {
-        padding-top: 10px;
-    }
-
     &__label {
         font-size: 0.9em;
+        padding: 0.50em 0em 0.50em 0.70em;
```
ptomato: I would suggest moving this entire section to `_legacyPolaroid.scss`, I can't think of a case where we would not want this padding.

katelk28: @ptomato good point, will do!

PR #1351, opened 2017-02-06
Make sure background stretches to fill width of set page

rmacqueen:
https://phabricator.endlessm.com/T15252

### DIFF
--- a/data/css/escola.scss
+++ b/data/css/escola.scss
@@ -69,6 +69,7 @@ $display-font: 'Raleway Black' !default;
   .background {
     background-image: url($secondary-background-image);
     background-repeat: no-repeat;
+    background-size: 100%;
   }
 
   .ContentGroupContentGroup {


data/css/escola.scss
```diff
@@ -69,6 +69,7 @@ $display-font: 'Raleway Black' !default;
   .background {
     background-image: url($secondary-background-image);
     background-repeat: no-repeat;
+    background-size: 100%;
```
ptomato: Don't you want `cover` here?

rmacqueen: As I mentioned on call, not all these background images are same size. But we always want them to cover full _width_, but not necessarily height, of `.background`, and not cut off parts of the image. 

`background-size: 100%` expands to `background-size: 100% auto`: http://www.w3schools.com/cssref/css3_pr_background-size.asp

PR #1352, opened 2017-02-06
Don't add card to arrangement map twice.

rmacqueen:
We already add it once in the create_card function.
No need to do it again in set_models.

https://phabricator.endlessm.com/T13398


xjuan:
CR pass

PR #1353, opened 2017-02-08
T15443

ptomato:
https://phabricator.endlessm.com/T15443

### DIFF
--- a/js/app/courseHistoryStore.js
+++ b/js/app/courseHistoryStore.js
@@ -69,7 +69,7 @@ const CourseHistoryStore = new GObject.Class({
                     break;
                 case Actions.DBUS_LOAD_ITEM_CALLED:
                     this.load_dbus_item(payload.ekn_id, payload.query,
-                        payload.timestamp);
+                        payload.timestamp || Gdk.CURRENT_TIME);
                     break;
             }
         });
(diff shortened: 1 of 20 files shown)


js/app/courseHistoryStore.js
```diff
@@ -85,7 +85,7 @@ const CourseHistoryStore = new GObject.Class({
         Eknc.Engine.get_default().query_promise(query)
         .then((results) => {
             if (results.models.length > 0) {
-                this.set_current_subset(results[0]);
+                this.set_current_subset(results.models[0]);
```
rmacqueen: weird. How did this ever work??!

js/app/widgets/slidingPanelOverlay.js
```diff
@@ -28,6 +28,7 @@ const SlidingPanelOverlay = new Knowledge.Class({
     },
 
     _init: function (props) {
+        this._transition_duration = 250;
```
rmacqueen: maybe worth moving the `250` into a constant since it is used twice in this file

js/app/courseHistoryStore.js
ptomato: Indeed, my question exactly - I was wondering if you knew :-)

If this is actually never reached, then maybe we should just remove it instead?

rmacqueen: @ptomato https://github.com/endlessm/eos-knowledge-lib/commit/944f983b93814a99f9b164651c06767deb7004d9

Looks like at some point we moved from `results` being the array of models itself, to it being an object with the `models` property on it. I guess we didn't catch it because only capacitate app hits this code path and we didn't end up releasing capacitate.

In any case, your fix is correct :-)

rmacqueen:
CR PASS

PR #1355, opened 2017-02-09
T15257

mattdangerw:
https://phabricator.endlessm.com/T15257

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -11,6 +11,7 @@
 #include <eos-shard/eos-shard-record.h>
 #include <eos-shard/eos-shard-dictionary.h>
 #include <eos-shard/eos-shard-shard-file.h>
+#include <string.h>
 
 GQuark
 eknc_domain_error_quark (void)
@@ -847,6 +848,55 @@ eknc_domain_query_finish (EkncDomain *self,
                        "upper-bound", state->upper_bound,
                        "models", state->models,
                        NULL);
+}
+
+/**
+ * eknc_domain_read_uri:
+ * @self: domain
+ * @uri: the ekn uri to read
+ * @bytes: (out) (transfer full) (allow-none): return location for the contents GBytes
+ * @mime_type: (out) (transfer full) (allow-none): return location for the content mime type
+ * @error: #GError for error reporting
+ *
+ * Reads the contents of a ekn uri and returns a GBytes of the contents and the
+ * contents mime type, if the ekn uri contents was found.
+ *
+ * Returns: true if the uri was successfully searched for, false if an error occurred
+ */
+gboolean
+eknc_domain_read_uri (EkncDomain *self,
+                      const gchar *uri,
+                      GBytes **bytes,
+                      const gchar **mime_type,
+                      GError **error)
+{
+  g_return_val_if_fail(uri && g_str_has_prefix (uri, "ekn://"), FALSE);
+  g_auto(GStrv) tokens = g_strsplit (uri + strlen("ekn://"), "/", -1);
+  g_return_val_if_fail(tokens && tokens[0] && tokens[1], FALSE);
+
+  g_autoptr(EosShardRecord) record = NULL;
+  /* iterate over all shards until we find a matching record */
+  for (GSList *l = self->shards; l && !record; l = g_slist_next (l))
+    record = eos_shard_shard_file_find_record_by_hex_name (l->data, tokens[1]);
+
+  if (!record)
+    return TRUE;
+
+  g_autoptr(EosShardBlob) blob = eos_shard_blob_ref (record->data);
+  /* Use resource, if present */
+  if (tokens[2])
+    blob = eos_shard_record_lookup_blob (record, tokens[2]);
+
+  if (!blob)
+    return TRUE;
+
+  if (mime_type)
+    *mime_type = g_strdup (eos_shard_blob_get_content_type (blob));
+
+  if (bytes)
+    *bytes = eos_shard_blob_load_contents (blob, error);
+
+  return TRUE;
 }
 
 /**
--- a/ekncontent/ekncontent/eknc-xapian-bridge.c
+++ b/ekncontent/ekncontent/eknc-xapian-bridge.c
@@ -316,6 +316,8 @@ send_json_ld_request (EkncXapianBridge *self,
                       SoupURI *uri,
                       GTask *task)
 {
+  g_autofree gchar * uri_string = soup_uri_to_string (uri, FALSE);
+  g_debug ("Sending request to xapian-bridge at %s\n", uri_string);
   g_autoptr(SoupMessage) request = soup_message_new_from_uri (SOUP_METHOD_GET, uri);
 
   RequestState *state = g_task_get_task_data (task);
--- a/ekncontent/autogen.sh
+++ b/ekncontent/autogen.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Bootstrap script for Endless OS Knowledge Apps
+# Run this script on a clean source checkout to get ready for building.
+
+test -n "$srcdir" || srcdir=`dirname "$0"`
+test -n "$srcdir" || srcdir=.
+olddir=`pwd`
+
+cd $srcdir
+test -f configure.ac || {
+    echo "You must run this script in the top-level checkout directory"
+    exit 1
+}
+
+# NOCONFIGURE is used by gnome-common
+if test -z "$NOCONFIGURE"; then
+    echo "This script will run ./configure automatically. If you wish to pass "
+    echo "any arguments to it, please specify them on the $0 "
+    echo "command line. To disable this behavior, have NOCONFIGURE=1 in your "
+    echo "environment."
+fi
+
+# Run the actual tools to prepare the clean checkout
+gtkdocize || exit $?
+autoreconf -fi || exit $?
+
+cd "$olddir"
+test -n "$NOCONFIGURE" || "$srcdir/configure" "$@"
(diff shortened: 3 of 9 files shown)


mattdangerw:
Still need to fixup the mainloop issues, but opened https://phabricator.endlessm.com/T15880 for that.

Ready for review!

ptomato:
Not a full review, I still need to read up on v8 API. But a few comments:
- We discussed having this in a separate repo and I remember you had a reason not to, but I can't remember what it was. However, I think keeping it in eos-knowledge-lib is going to make setting up automated testing a lot more complicated.
- Would suggest squashing "initial bindings", "remove *", and "remove GNodeJS namespace" into an initial commit.
- Is `eknc_domain_read_uri()` only useful for the bindings? Seems like it could just as well be done in `domain.cc` since you have an extra C++ function that wraps it anyway.

mattdangerw:
> We discussed having this in a separate repo and I remember you had a reason not to, but I can't remember what it was. However, I think keeping it in eos-knowledge-lib is going to make setting up automated testing a lot more complicated.

Changing things like the engine API might require changing these bindings. My thought process was it would be easier to change of all of that in one single pull request than coordinating across repos.

For me ideally the ekncontent library is split out from eos-knowledge-lib, and the gjs specific code (lightweight overrides) and nodejs specific code both live in tree and are automatically tested in make check.

Then someone who wants to change something for eos-knowledge-lib in ekncontent, but is not worried about the nodejs bindings, could do their work and notice that they broke make check and fix things up on issue. Better then everything going along happily until a couple weeks later someone working with the node bindings raising a stink.

Also nice, parallel and consistent to newbies that gjs and nodejs specific code on top of the library live in the same place.

I do see the point that getting nodejs automated testing on the same box as the main library testing would be a pain, but do think it should be doable. Either in a flatpak sandbox on our ci, or with something like travis. IMO better to figure that out than require ekncontent library developers to hold an extra check in their head of testing an external repo.

So I *think* I'd rather split ekncontent off (before or after issue), than split the node bindings off. If our content library were more mature and ready to worry about API versioning and ABI compatibility, I'd be more inclined to switch. But I don't think we want to be there for a while. Better to be able to fearlessly change our content library and format without getting sucked into compatibility woes.

Sorry kinda rambly comment, but hopefully gets my thoughts across?

> Would suggest squashing "initial bindings", "remove *", and "remove GNodeJS namespace" into an initial commit.

Disagree here. That was quite deliberate, and I think its useful to have a recording of what we are changing from the code brought in from `node-gtk`. Its not like the extra commits are getting in the way of anything. Strictly more information.

> Is eknc_domain_read_uri() only useful for the bindings? Seems like it could just as well be done in domain.cc since you have an extra C++ function that wraps it anyway.

Yeah could do, that would mean the nodejs bindings would need to link the shard library, but that's not really a problem. Perhaps better would be to leave it and port https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/articleHTMLRenderer.js#L135 to it. The implementation in `read_uri` should be a little more efficient than what we are doing in `read_stream_sync` right now.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +850,56 @@ eknc_domain_query_finish (EkncDomain *self,
...
+ * Returns: (transfer full): the gbytes
+ */
+GBytes *
+eknc_domain_read_uri (EkncDomain *self,
+                      const gchar *uri,
+                      const gchar **mime_type,
+                      GError **error)
+{
+  if (!uri || !g_str_has_prefix (uri, "ekn:"))
+    return NULL;
```
ptomato: Use `g_return_val_if_fail()` since this seems like a programmer error?

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +850,56 @@ eknc_domain_query_finish (EkncDomain *self,
...
+
+  g_auto(GStrv) tokens = g_strsplit (uri + 6, "/", -1);
```
ptomato: For clarity use `strlen("ekn://")`, pretty sure any compiler will optimize it back to 6

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +850,56 @@ eknc_domain_query_finish (EkncDomain *self,
...
+  if (!tokens || !tokens[0] || !tokens[1])
+    return NULL;
```
ptomato: Another programmer error?

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +850,56 @@ eknc_domain_query_finish (EkncDomain *self,
...
+      if (!record)
+        continue;
+
+      EosShardBlob *blob = record->data;
+      /* Use resource, if present */
+      if (tokens[2])
+        blob = eos_shard_record_lookup_blob (record, tokens[2]);
+
+      if (!blob)
+        continue;
```
ptomato: This will go to the next iteration of the loop which will immediately end because `record != NULL`. That has the effect of making this statement equivalent to `return NULL`.

ekncontent/ekncontent/eknc-xapian-bridge.c
```diff
@@ -316,6 +316,7 @@ send_json_ld_request (EkncXapianBridge *self,
                       SoupURI *uri,
                       GTask *task)
 {
+  g_debug ("Sending request to xapian-bridge at %s\n", soup_uri_to_string (uri, FALSE));
```
ptomato: The URI string is leaked.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +850,56 @@ eknc_domain_query_finish (EkncDomain *self,
...
+
+      if (mime_type)
+        *mime_type = eos_shard_blob_get_content_type (blob);
+
+      return eos_shard_blob_load_contents (blob, error);
+    }
+
+  return NULL;
```
ptomato: You can fall off the end of the loop and `return NULL` without having set `error`. If null is a valid return value then you should have a `(nullable)` annotation in the doc comment, but additionally that poses a problem regarding the usual API pattern in GLib.

ekncontent/Makefile.am
```diff
@@ -244,7 +244,7 @@ typelibdir = $(libdir)/girepository-1.0
 typelib_DATA = $(INTROSPECTION_GIRS:.gir=.typelib)
 
 overridesdir = $(datadir)/gjs-1.0/overrides
-overrides_DATA = overrides/EosKnowledgeContent.js
```
ptomato: Contingent on whether the node bindings move to another repo or not. (If they do, then I'd prefer to keep `overrides` since that mirrors the directory where they have to be installed.)

mattdangerw: I'll just kill this commit I think

ekncontent/ekncontent/eknc-domain.c
mattdangerw: So this is all copied from our vfs, but yeah I'll clean it up.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +851,55 @@ eknc_domain_query_finish (EkncDomain *self,
...
+                      GError **error)
+{
+  g_return_val_if_fail(uri && g_str_has_prefix (uri, "ekn://"), FALSE);
+  g_auto(GStrv) tokens = g_strsplit (uri + strlen("ekn://"), "/", -1);
+  g_return_val_if_fail(tokens && tokens[0] && tokens[1], FALSE);
+
+  EosShardRecord *record = NULL;
+  /* iterate over all shards until we find a matching record */
+  for (GSList *l = self->shards; l && !record; l = g_slist_next (l))
+    record = eos_shard_shard_file_find_record_by_hex_name (l->data, tokens[1]);
```
ptomato: `record` is leaked, if found

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +851,55 @@ eknc_domain_query_finish (EkncDomain *self,
...
+
+  if (!record)
+    return TRUE;
+
+  EosShardBlob *blob = record->data;
+  /* Use resource, if present */
+  if (tokens[2])
+    blob = eos_shard_record_lookup_blob (record, tokens[2]);
```
ptomato: `blob` is leaked if this statement is executed (though not if `blob == record->data` still)

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +851,55 @@ eknc_domain_query_finish (EkncDomain *self,
...
+
+  if (!blob)
+    return TRUE;
+
+  if (mime_type)
+    *mime_type = eos_shard_blob_get_content_type (blob);
```
ptomato: I think you have to `g_strdup()` this and remove `const` from `mime_type` in the function signature, it's transfer none so otherwise it won't outlive the `blob` or the `record`, and you have it annotated as transfer full.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -850,6 +851,55 @@ eknc_domain_query_finish (EkncDomain *self,
 }
 
 /**
+ * eknc_domain_read_uri:
+ * @self: domain
+ * @uri: the ekn uri to read
+ * @bytes: (out) (transfer none) (allow-none): return location for the contents GBytes
```
ptomato: I think this has to be `(transfer full)`

ekncontent/ekncontent/eknc-domain.c
mattdangerw: Whoops meant to mark the bytes transfer full and not this, but yeah I think both transfer full makes sense

mattdangerw:
Good catch, tests should be fixed.

PR #1359, opened 2017-02-17
EknMediaBin: fixed object leaks

xjuan:
Fixed video_widget and playbin object leaks.

https://phabricator.endlessm.com/T15393

### DIFF
--- a/lib/eosknowledgeprivate/ekn-media-bin.c
+++ b/lib/eosknowledgeprivate/ekn-media-bin.c
@@ -76,6 +76,7 @@ typedef struct
   /* We place extra flags here so the get squashed with the boolean properties */
   gboolean title_user_set:1;            /* True if the user set title property */
   gboolean description_user_set:1;      /* True if the user set description property */
+  gboolean dump_dot_file:1;             /* True if GST_DEBUG_DUMP_DOT_DIR is set */
   gboolean ignore_adjustment_changes:1;
 
   /* Internal Widgets */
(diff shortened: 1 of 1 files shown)


lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1535,6 +1550,8 @@ ekn_media_bin_handle_msg_state_changed (EknMediaBin *self, GstMessage *msg)
              gst_element_state_get_name (old_state),
              gst_element_state_get_name (new_state));
 
+  ekn_media_bin_dump_dot (self, old_state, new_state);
```
ptomato: Don't you want to put some ifdef or env var guard around this so that you're not computing a filename on every state change?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -1796,17 +1823,19 @@ static void
 ekn_media_bin_init_playbin (EknMediaBin *self)
 {
   EknMediaBinPrivate *priv = EMB_PRIVATE (self);
+  GstElement *play;
```
ptomato: nitpick: Why create the local variable when you could assign directly to `priv->play`?

lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -622,10 +623,37 @@ ekn_media_bin_init_video_sink (EknMediaBin *self)
...
+  /* Stop Playback to give gst a chance to cleanup its mess */
+  if (priv->play)
+    gst_element_set_state (priv->play, GST_STATE_NULL);
+
+  /* Stop bus watch */
+  if (priv->bus)
+    {
+      gst_bus_set_flushing (priv->bus, TRUE);
+      gst_bus_remove_watch (priv->bus);
+      g_clear_pointer (&priv->bus, g_object_unref);
```
ptomato: Can be `g_clear_object()`

xjuan: I meant to use gst_object_unref() instead of regular g_object_unref()
In any case I will replace it with gst_object_replace()

lib/eosknowledgeprivate/ekn-media-bin.c
xjuan: Fixed

PR #1361, opened 2017-02-23
T15239

katelk28:
Part of design QA for the FNDE apps. Specifically:
- Library-List preset
- Escola preset

There is another [PR](https://github.com/endlessm/endless-content-factory/pull/579) for the same Phabricator ticket in endless-content-factory for the Portal do Professor article page SCSS/CSS.

https://phabricator.endlessm.com/T15239

### DIFF
--- a/data/css/escola.scss
+++ b/data/css/escola.scss
@@ -8,6 +8,8 @@ $highlight-color: #f5d888  !default;
 
 $background-light-color: #2b2b2c !default;
 $background-dark-color: #141418 !default;
+$background-dark-color-hover: #1e1e22 !default;
+$card-context-color: #b3b3b3 !default;
 
 $title-font: Raleway !default;
 $context-font: Raleway !default;
@@ -20,13 +22,21 @@ $display-font: 'Raleway Black' !default;
   background-image: none;
 }
 
+.DecorationThemeableImage {
+  margin: 0px 30px;
+}
+
+.NavigationSearchBox {
+  font-family: $context-font;
+}
+
 .home-page {
   .landing {
     .background {
       background-image: url($primary-background-image);
       background-repeat: no-repeat;
-      background-size: 100%;
-      min-height: 346px;
+      background-size: cover;
+      min-height: 400px;
     }
 
     .BannerDynamic {
@@ -45,11 +55,13 @@ $display-font: 'Raleway Black' !default;
       color: white;
       font-size: 30px;
       font-family: $title-font;
+      font-weight: 500;
+      margin-top: -0.90em;
     }
   }
 
   .ContentGroupContentGroup {
-    padding: 20px;
+    padding: 40px;
   }
 }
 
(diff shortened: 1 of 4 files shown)


data/css/escola.scss
```diff
@@ -203,3 +254,52 @@ where it needs to be */
...
+    .landing {
+      .background {
+        min-height: 250px;
+      }
+
+      .BannerDynamic {
+        &__logo {
+          min-height: 120px;
+          min-width: 400px;
+          font-size: 60px;
```
tchx84: `font-size` doesn't really affect the size of the DynamicLogo,the size you see is only the result of the min-height and min-width you use there.

data/css/escola.scss
```diff
@@ -191,10 +237,15 @@ where it needs to be */
...
-      color: #b3b3b3;
+      color: $card-context-color;
     }
     &__thumbnail {
-      background-color: $base-color;
+      background-color: $background-dark-color;
+    }
+    &.width-d.height-b &__contentFrame,
+    &.width-e.height-b &__contentFrame,
+    &.width-e.height-c &__contentFrame {
```
tchx84: Watch out with the `&` here! `&` means "everything that comes higher in the selector hierarchy", so what you get from:

```
 &.width-e.height-c &__contentFrame {
``` 

is actually:

```
.ArrangementGrid .CardThumb.width-e.height-c .ArrangementGrid .CardThumb__contentFrame {
      padding: 20px;
}
```

Notice the `.ArrangementGrid` in the middle, therefore this selector won't work.

data/css/escola.scss
```diff
@@ -122,16 +132,38 @@ $display-font: 'Raleway Black' !default;
...
+    font-size: 1.25em;
+    color: white;
+    padding: 0em 1.25em 0em 1.25em;
+    font-style: normal;
+  }
+  &__subtitle {
+    margin-top: 1.0em;
+    padding: 0;
+  }
+}
```
tchx84: I think there is something off in composite mode with the no results page:

virtualbox_endlessos_23_02_2017_16_21_56

data/css/escola.scss
```diff
@@ -203,3 +254,52 @@ where it needs to be */
...
+  .set-page,
+  .search-page {
+    .ContentGroupContentGroup,
+    .BannerSearch {
+      &__title {
+        font-size: 30px;
+      }
+    }
+  }
+}
```
tchx84: Just asking here, but was there any plans to improve the look of the search box popup?

virtualbox_endlessos_23_02_2017_16_19_42

data/css/escola.scss
katelk28: You're right -- am fixing the visuals a bit.

xjuan: virtualbox_endless dev_28_02_2017_14_18_40

Looks good now!

xjuan:
CR PAss

PR #1362, opened 2017-02-27
T15634

ebassi:
https://phabricator.endlessm.com/T15634

### DIFF
--- a/js/app/modules/card/legacyPolaroid.js
+++ b/js/app/modules/card/legacyPolaroid.js
@@ -122,17 +122,26 @@ const LegacyPolaroid = new Module.Class({
 
     vfunc_get_preferred_width: function () {
         // FIXME: figure out better way to respect childs size request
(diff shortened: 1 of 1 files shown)


data/css/modules/card/_legacyPolaroid.scss
```diff
@@ -7,7 +7,7 @@
         padding: 4px 0px;
         color: $accent-dark-color;
         font-family: $title-font, sans-serif;
-        font-size: 0.77em;
+        font-size: 0.67em;
```
ptomato: Instead of affecting the font size for the majority of apps that work fine, I'd prefer changing this font size only for the four affected apps that use the Bangladesh font, in their overrides.scss in eos-subscriptions-apps.

ebassi: Fair enough, though of course this issue may still manifest itself on other apps while we add more localizations.

PR #1364, opened 2017-03-02
introspect: Support absolute file paths

ptomato:
The "introspect" tool previously supported module names as its argument
(e.g. "Filter.Unread"). This adds a --file argument that takes a path
which will be imported instead.

https://phabricator.endlessm.com/T15141


ebassi:
Looks nice.

PR #1365, opened 2017-03-02
T15785

ebassi:
https://phabricator.endlessm.com/T15785

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -85,9 +85,9 @@ AC_REQUIRE_AUX_FILE([tap-driver.sh])
 AC_PATH_PROG([NATURALDOCS], [naturaldocs], [notfound])
 AC_ARG_VAR([NATURALDOCS], [Path to naturaldocs])
 AC_SUBST([GLIB_COMPILE_RESOURCES], [`$PKG_CONFIG --variable glib_compile_resources gio-2.0`])
-AC_PATH_PROG([SCSS], [scss], [notfound])
-AS_IF([test x$SCSS = xnotfound], [AC_MSG_ERROR([SCSS compiler not found.
-Try installing the "sass" Ruby gem or its equivalent package.])])
+AC_PATH_PROG([SCSS], [sassc], [notfound])
+AS_IF([test x$SCSS = xnotfound], [AC_MSG_ERROR([Sass compiler not found.
+Try installing "sassc" or its equivalent package.])])
 AC_ARG_VAR([SCSS], [Path to SCSS compiler])
 
 AC_CACHE_SAVE
--- a/configure.ac
+++ b/configure.ac
@@ -89,10 +89,6 @@ AC_PATH_PROG([SCSS], [scss], [notfound])
 AS_IF([test x$SCSS = xnotfound], [AC_MSG_ERROR([SCSS compiler not found.
 Try installing the "sass" Ruby gem or its equivalent package.])])
 AC_ARG_VAR([SCSS], [Path to SCSS compiler])
-AC_MSG_CHECKING([whether SCSS compiler supports compass])
-AS_IF([scss --compass], [AC_MSG_RESULT([yes])],
-    [AC_MSG_ERROR([SCSS compiler does not have Compass support.
-Try installing the "compass" Ruby gem or its equivalent package.])])
 
 AC_CACHE_SAVE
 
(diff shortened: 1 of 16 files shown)


ebassi:
This not thoroughly tested, and it's open mostly as a way to do a preliminary code review.

data/templates/css/wikimedia.scss
```diff
@@ -1,5 +1,4 @@
 @import "project-setup";
-@import "compass";
```
mattdangerw: You'll need to go a bit farther to ditch compass from these style sheets.

`@include global-reset;` I think is compass

`@include box-sizing(border-box);` is a silly mixin for cross browser vendor prefixing we can just replace with `box-sizing: border-box;`

Then `image-height($file)` and `image-width($file)` and `image-url($file)` will all need to be replaced with actual image widths and heights for our assets and `url(../some/path/)` calls.

data/css/_project-setup.scss
```diff
@@ -6,6 +6,3 @@
 
 $images-path: 'resource:///com/endlessm/knowledge/data/images/';
```
mattdangerw: I wonder if we can ditch this project setup entirely with the switch away from compass, I suspect we can.

tchx84: Aside from Matt's comments, we also have ruby-sass/compass bits in https://github.com/endlessm/eos-subscriptions-apps

data/templates/css/wikimedia.scss
tchx84: There is also `@include pie-clearfix;` from [compass](http://compass-style.org/reference/compass/utilities/general/clearfix/) but not exactly sure what it does.

tchx84: Done with the diff, very few cases for image-height/width, therefore I agree with @mattdangerw here lets just replace it with actual values.

ebassi: @tchx84 `pie-clearfix` is a cross-browser fix for extending the bottom of the nesting element: http://compass-style.org/reference/compass/utilities/general/clearfix/

configure.ac
```diff
@@ -85,9 +85,9 @@ AC_REQUIRE_AUX_FILE([tap-driver.sh])
 AC_PATH_PROG([NATURALDOCS], [naturaldocs], [notfound])
 AC_ARG_VAR([NATURALDOCS], [Path to naturaldocs])
 AC_SUBST([GLIB_COMPILE_RESOURCES], [`$PKG_CONFIG --variable glib_compile_resources gio-2.0`])
-AC_PATH_PROG([SCSS], [scss], [notfound])
-AS_IF([test x$SCSS = xnotfound], [AC_MSG_ERROR([SCSS compiler not found.
-Try installing the "sass" Ruby gem or its equivalent package.])])
+AC_PATH_PROG([SCSS], [sassc], [notfound])
+AS_IF([test x$SCSS = xnotfound], [AC_MSG_ERROR([Sass compiler not found.
+Try installing the "sassc" its equivalent package.])])
```
ptomato: Nitpick: `Try installing "sassc" or its equivalent package` (remove `the`, add `or`)

configure.ac
```diff
@@ -85,9 +85,9 @@ AC_REQUIRE_AUX_FILE([tap-driver.sh])
 AC_PATH_PROG([NATURALDOCS], [naturaldocs], [notfound])
 AC_ARG_VAR([NATURALDOCS], [Path to naturaldocs])
 AC_SUBST([GLIB_COMPILE_RESOURCES], [`$PKG_CONFIG --variable glib_compile_resources gio-2.0`])
-AC_PATH_PROG([SCSS], [scss], [notfound])
-AS_IF([test x$SCSS = xnotfound], [AC_MSG_ERROR([SCSS compiler not found.
-Try installing the "sass" Ruby gem or its equivalent package.])])
+AC_PATH_PROG([SCSS], [sassc], [notfound])
```
ptomato: For development tools, it's likely that we'll continue using apt packages at least for a while, until it becomes more convenient to develop within the sandbox. How about making this into `[sassc scss]` so that both compilers are supported for the time being? If not, then we should make sure to make an apt package for `sassc`.

ebassi: The problem is that the Ruby-Sass command line tool does not have the same arguments as `sassc`.

We could have two different variables and then a conditional to decide which rule to use.

ebassi: Debian has `sassc` in Stretch (current testing): https://packages.debian.org/stretch/sassc

ebassi:
> @ptomato (3:49 PM)
> @ebassi go right ahead and click the button, I thought I already approved those

PR #1370, opened 2017-03-08
Add base theme to picard as well

rmacqueen:
So that we reset Adwaita styling

https://phabricator.endlessm.com/T15903

### DIFF
--- a/tools/picard.js
+++ b/tools/picard.js
@@ -82,7 +82,8 @@ const CSS = '\
     to { box-shadow: 0px 0px 7px 3px alpha(#729fcf, 0.4),\
                      inset 0px 0px 10px alpha(#729fcf, 0.8); }\
 }\
-.hint { animation: glow 1s infinite alternate; }';
+.hint { animation: glow 1s infinite alternate; }\
+EosWindow { background-color: lightgray; }'
 
 let widgets = {};
 
--- a/tools/picard.js
+++ b/tools/picard.js
@@ -302,6 +302,8 @@ function build_ui () {
         default_width: RESOLUTIONS[2][0],
         default_height: RESOLUTIONS[2][1],
     });
+    // Add the .WindowSimple class so we get styling on the scrollbars
+    widgets.window.get_style_context().add_class('WindowSimple');
     widgets.reset_theme_box = new ResetThemeBox();
     widgets.window.set_titlebar(widgets.titlebar);
     add_remove.add(widgets.add_box);
--- a/tools/picard.js
+++ b/tools/picard.js
@@ -4,6 +4,7 @@ const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const GObject = imports.gi.GObject;
 const Gtk = imports.gi.Gtk;
+const Lang = imports.lang;
 
 const Card = imports.app.interfaces.card;
 const Knowledge = imports.app.knowledge;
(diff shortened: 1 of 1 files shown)


mattdangerw:
Won't quite work, as our rest only applies to eos window
https://github.com/endlessm/eos-sdk/blob/master/data/css/endless-widgets.css#L3

You can see it properly apply with `Gtk.Widget.set_css_name.call(Gtk.Window, 'EosWindow');` in picard, but then things get really obviously messed up in the titlebar, where we rely on adwaita.

rmacqueen:
works now adding in your commit! thanks

tchx84:
I noticed two things, 1) a different background color (not necessarily an issue, but since is too dark maybe we won't notice other details like shadows, etc) and 2) the scrollbar slider is missing (maybe too thin?)

virtualbox_endlessos_09_03_2017_17_12_50

mattdangerw:
> a different background color

That's the default background color for windows in our sdk theme. We should set a new background color specifically for picard if that one is undesired.

mattdangerw:
> the scrollbar slider is missing (maybe too thin?)

I guess our scrollbar theme is hidden in a `.WindowSimple` selector. We could either move it out, or add a `.WindowSimple` style class to our picard window.

ptomato:
This seems a bit convoluted - wouldn't it work just to add our own, non eos-specific reset.css in Picard?

mattdangerw:
> This seems a bit convoluted - wouldn't it work just to add our own, non eos-specific reset.css in Picard?

Yeah something like
```
.picard-contents,
.picard-contents :not(spinner):not(decoration) {
    all: unset;
}
```
loaded in yet another style sheet *before* the endless style sheet include would work.

But at the same time it would definitely be more fragile. Any changes to how we handle our theme in `eos-sdk` would break this, and I very much doubt that future developer changing sdk theme would remember "oh i need to update picard as well."

I think its best to try a keep things coupled. And if we want to sink more time in a do this in a nicer way, our efforts would be best spent making `eos-sdk` a bit more flexible. e.g. a function to load our theme on a non endless window.

tools/picard.js
```diff
@@ -247,6 +247,15 @@ function get_module_menu () {
     return menu;
 }
 
+// Our sdk theme only resets widgets inside a window will css name EosWindow. We
+// will take advantage of that to make an adwaita themed topbar with endless
+// themed widgets inside.
+const ResetThemeBox = new Knowledge.Class({
```
ptomato: No need to use `Knowledge.Class` here and in fact it's easier to use `Gtk.WidgetClass`'s `CssName` property:
```js
const ResetThemeBox = new Lang.Class({
    Name: 'ResetThemeBox',
    Extends: Gtk.Box,
    CssName: 'EosWindow',
});
```
Also make sure the names match up.

tools/picard.js
```diff
@@ -290,14 +299,16 @@ function build_ui () {
         default_width: RESOLUTIONS[2][0],
         default_height: RESOLUTIONS[2][1],
     });
+    widgets.reset_theme_box = new ResetThemeBox();
```
ptomato: Since you don't need the reset-theme-box anywhere else, you can just keep it local to this function.

tools/picard.js
```diff
@@ -299,6 +299,8 @@ function build_ui () {
         default_width: RESOLUTIONS[2][0],
         default_height: RESOLUTIONS[2][1],
     });
+    // Add the .WindowSimple class so we get styling on the scrollbars
+    widgets.window.get_style_context().add_class('WindowSimple');
```
ptomato: Consider moving the scrollbar styling into EosWindow, instead of doing this? I don't know if that's feasible or desired though

rmacqueen: Feasible yes, but do we want that? I thought those scrollbars were pretty specific to knowledge-lib. Not sure we want them on say, photo app?

tools/picard.js
rmacqueen: done

tools/picard.js
tchx84: Not sure if there is a better way of doing this, but I do think we should keep this specific to e-k-l, and not cause undesired changes on other apps.

PR #1372, opened 2017-03-10
T16085

katelk28:
Picard: DummyReadingHistoryModel GObject, ScrolledWindow expand

- DummyReadingHistoryModel wasn't extending from GObject, which was
  messing up Card.List being shown
- Gtk.ScrolledWindow didn't have an expand: true option, which prevented
  Card.List be able to expand within various arrangements

https://phabricator.endlessm.com/T16085


rmacqueen:
Good catch!

PR #1376, opened 2017-03-14
Update JHBuild dependencies

ptomato:
Remove compass and dbusmock, as they are no longer needed. Replace the
check for the SCSS compiler with the compiler binary name from libsass.

https://phabricator.endlessm.com/T16140


rmacqueen:
CR PASS

PR #1382, opened 2017-03-17
T16193

rmacqueen:
https://phabricator.endlessm.com/T16193


js/app/modules/card/thumb.js
```diff
@@ -25,23 +32,30 @@ const Thumb = new Module.Class({
...
-        this._text_layout.add(this._context_widget);
+        this._context_frame.add(this.create_context_widget_from_model());
 
         this.show_all();
 
         Utils.set_hand_cursor_on_widget(this);
+        this._card_type = this._get_card_type();
+    },
+
+    // FIXME: This should examine the card's thumbnail to determine which card
```
ptomato: Yep, this will need to be implemented before merging :smiley:

rmacqueen:
> So if it's something wrong with my version of Emeus, then it's probably caused by the same thing.

See my comment on other issue. I think you need to be on @tchx84's branch of emeus until it gets merged

js/app/modules/card/defaultFamily.js
```diff
@@ -192,21 +291,33 @@ const Thumb = new Module.Class({
     vfunc_size_allocate: function (alloc) {
         this._layout.clear_constraints();
```
tchx84: just a reminder: don't _always_ remove the constraints, move all this logic to a separate method and replace constraints only when really necessary... otherwise the performance cost is really high because the solver has a to start from scratch every time (instead of reusing previous calculations).

tchx84: Actually, talking to @ebassi here (and mixed with what we already spoke), the most efficient way would be to 1) create ALL constraints at _init (once) and then 2) enable/disable groups of constraints (the API allow this).

rmacqueen: so I spent pretty much all day after presentation on this. I thought it would be straightforward but it's actually non-trivial.

>  1) create ALL constraints at _init (once) and then 2) enable/disable groups of constraints (the API allow this).

To do this you essentially have to simulate every possible configuration of the card at init and generate constraints for them, e.g. thumb card, horizontal alignment with synopsis showing; thumb card, horizontal alignment with synopsis not showing; thumb card, vertical alignment, context widget showing; etc etc.

It gets really gnarly trying to maintain all of those constraints and keep track of which one belongs to which situation. There's probably a clean-ish way to do it, but I wasn't able to find it.

Will try some more tomorrow

ebassi: I agree with @rmacqueen; I'm looking at implementing some additional API in Emeus to have "constraint groups" to have semantic aggregates of constraints, and especially to ensure that switching between different layouts is not slowed down by the removal/addition of all constraints.

PR #1384, opened 2017-03-25
T16344

tchx84:
https://phabricator.endlessm.com/T16344

### DIFF
--- a/js/app/downloader.js
+++ b/js/app/downloader.js
@@ -10,6 +10,8 @@ const Soup = imports.gi.Soup;
 const Config = imports.app.config;
 const Utils = imports.app.utils;
 
+const CHUNK_SIZE = 1024 * 256;
+
 const FileDownloadRequest = new Lang.Class({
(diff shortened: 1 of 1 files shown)


tchx84:
@ptomato @xjuan do this makes sense?

js/app/downloader.js
```diff
@@ -10,6 +10,8 @@ const Soup = imports.gi.Soup;
 const Config = imports.app.config;
 const Utils = imports.app.utils;
 
+const CHUNK_SIZE = 1024 * 128;
```
ebassi: I don't want to bikeshed on the number, but I'd probably either downsize this to 64 or up it to 256 kB, with a preference for the latter.

ebassi:
Nice catch!

ebassi:
`Utils.CHUNK_SIZE` existed until https://phabricator.endlessm.com/T15257 was fixed.

ebassi:
CR pass.

PR #1388, opened 2017-03-31
Encyclopedia dynamic logo

ptomato:
https://phabricator.endlessm.com/T16259


data/css/encyclopedia.scss
```diff
@@ -3,6 +3,7 @@
 $body-font: 'Lato' !default;
 $display-font: 'Lato' !default;
 $title-font: 'Lato' !default;
+$logo-font: 'Libre Baskerville' !default;
```
tchx84: Just a reminder that we need this to be included in eos-theme (or as a dependency).

PR #1389, opened 2017-04-04
T14982

tchx84:
https://phabricator.endlessm.com/T14982

### DIFF
--- a/js/app/widgets/dynamicLogo.js
+++ b/js/app/widgets/dynamicLogo.js
@@ -4,6 +4,7 @@
 
 const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
 const Gio = imports.gi.Gio;
+const GLib = imports.gi.GLib;
 const Gtk = imports.gi.Gtk;
 const GObject = imports.gi.GObject;
 const Rsvg = imports.gi.Rsvg;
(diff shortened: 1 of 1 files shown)


js/app/widgets/dynamicLogo.js
```diff
@@ -135,6 +152,10 @@ const DynamicLogo = new Knowledge.Class({
     },
 
     _update_custom_style: function () {
+        this._max_width = EosKnowledgePrivate.style_context_get_custom_int(this.get_style_context(), 'max-width');
+        this._max_height = EosKnowledgePrivate.style_context_get_custom_int(this.get_style_context(), 'max-height');
+        this.queue_resize();
```
ptomato: This queues a resize anytime the style is updated, even if max-width and max-height didn't change.

js/app/widgets/dynamicLogo.js
```diff
@@ -62,6 +63,22 @@ const DynamicLogo = new Knowledge.Class({
 
     StyleProperties: {
         /**
+         * Property: max-width
+         *
+         * -EknDynamicLogo-max-width:
+         *  * maximum width in pixels.
+         */
+        'max-width': GObject.ParamSpec.int('max-width', 'max-width', 'max-width',
```
ptomato: I share @ebassi's trepidation of adding more style properties, since it will make it that much harder to move to GTK 3.9x. I wonder if there are any plans to parse the standard CSS max-width and max-height properties upstream, even if most or all widgets do not honor them.

js/app/widgets/dynamicLogo.js
```diff
@@ -135,6 +152,12 @@ const DynamicLogo = new Knowledge.Class({
     },
 
     _update_custom_style: function () {
+        let width = EosKnowledgePrivate.style_context_get_custom_int(this.get_style_context(), 'max-width');
```
ptomato: Factor out `let context = this.get_style_context();`, this function will be called quite often, potentially on every mouse movement

PR #1390, opened 2017-04-05
app: Add D-Bus endpoint for Grand Central queries

smspillaz:
https://phabricator.endlessm.com/T16409

### DIFF
--- a/js/app/application.js
+++ b/js/app/application.js
@@ -43,6 +43,15 @@ const KnowledgeSearchIface = '\
   </interface> \
 </node>';
 
+const GrandCentralContentIface = '\
+<node> \
+  <interface name="com.endlessm.GrandCentralContent"> \
+    <method name="ArticleCardDescriptions"> \
+      <arg type="aa{ss}" name="Result" direction="out" /> \
+    </method> \
+  </interface> \
+</node>';
+
 const CREDITS_URI = 'resource:///app/credits.json';
 const APP_JSON_URI = 'resource:///app/app.json';
 const APP_YAML_URI = 'resource:///app/app.yaml';
@@ -51,6 +60,15 @@ const OVERRIDES_SCSS_URI = 'resource:///app/overrides.scss';
 
 const AUTOBAHN_COMMAND = 'autobahn -I ' + Config.YAML_PRESET_DIR + ' ';
 const SCSS_COMMAND = 'sassc -a -I ' + Config.TOP_THEME_DIR + ' ';
+
+function grandCentralObjectPath(applicationId) {
+    return (
+        '/com/endlessm/GrandCentral/ContentProviderV1/' +
+        applicationId.replace(/([^A-Za-z0-9])/g, function(m) {
+            return '_' + m.charCodeAt(0).toString(16);
+        })
+    );
+}
 
 /**
  * Class: Application
@@ -73,6 +91,7 @@ const Application = new Knowledge.Class({
         this.parent(props);
         this._controller = null;
         this._knowledge_search_impl = Gio.DBusExportedObject.wrapJSObject(KnowledgeSearchIface, this);
+        this._grand_central_content_impl = Gio.DBusExportedObject.wrapJSObject(GrandCentralContentIface, this);
         this.image_attribution_file = Gio.File.new_for_uri(CREDITS_URI);
 
         Eknc.Engine.get_default().default_app_id = this.application_id;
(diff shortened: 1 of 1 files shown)


ptomato:
Does this have to be a separate interface from KnowledgeSearch?

smspillaz:
@ptomato Yes. Later on we'll be adding more fine-grained queries which aren't just free text. The interface needs to support them.

js/app/application.js
```diff
@@ -43,6 +43,19 @@ const KnowledgeSearchIface = '\
...
+const GrandCentralContentIface = '\
+<node> \
+  <interface name="com.endlessm.GrandCentralContent"> \
+    <method name="ContentQuery"> \
+      <arg type="s" name="Query" direction="in" /> \
+      <arg type="s" name="Result" direction="out" /> \
+    </method> \
+    <method name="QueryNotifications"> \
+      <arg name="notifications" direction="out" type="aa{ss}" /> \
+    </method> \
```
cosimoc: I don't think this API should have anything related to notifications. Notifications already have their own DBus API and grand central should tap into that.

js/app/application.js
```diff
@@ -227,6 +253,25 @@ const Application = new Knowledge.Class({
...
+        return [];
+    },
+
+    ContentQueryAsync: function (query, invocation) {
+        let engine = Eknc.Engine.get_default();
+        engine.query_promise(Eknc.QueryObject.new_from_props({
+            tags_match_all: ['EknArticleObject'],
+            limit: 5
+        })).then(results =>
+            invocation.return_value(new GLib.Variant('(s)', [JSON.stringify(results.models.map(m => ({
```
cosimoc: I really don't like marshalling all of this through JSON. I would return variants directly, like the search provider interface does.

js/app/application.js
```diff
@@ -43,6 +43,19 @@ const KnowledgeSearchIface = '\
   </interface> \
 </node>';
 
+const GrandCentralContentIface = '\
+<node> \
+  <interface name="com.endlessm.GrandCentralContent"> \
+    <method name="ContentQuery"> \
```
cosimoc: Are you sure this particular interface should have a query parameter? As we discussed last week, I think that we should use different DBus interfaces for different content capabilities, and the most basic one should probably take no arguments.

ptomato:
> Later on we'll be adding more fine-grained queries which aren't just free text. The interface needs to support them.

Well yes, but nothing's stopping you from adding methods to KnowledgeSearch to accommodate that, right?
I ask because you might run into the same problem that made it necessary to create KnowledgeSearch in the first place: 20 knowledge apps all starting up their own search providers on DBus at the same time was a performance killer, and I expect the same thing to happen if 20 apps all start up their grand central providers at the same time too.

Another question, is "grand central" going to be a user- or developer-facing name for this feature? As far as I know it's internal jargon right now, but this pull request makes it part of the API that third-party developers will have to deal with.

js/app/application.js
smspillaz: Yeah I think that's a fair point. I'll give it a more specific name and then if we want to add more functionality we can do so without necessarily introducing API breaks.

js/app/application.js
smspillaz: I think that's fine, though it is slightly more prone to API breakage if we want to pack more things into the return value later.

js/app/application.js
smspillaz: Ah yes, I had this in there in the past, since I wanted to query the shell itself for historical notifications instead of handling new ones. Also, the code used to handle all the interactions for notification is ... complicated and it seemed easier to just re-use them. But nonetheless, notifications are not in scope right now, so I'll get rid of this.

smspillaz:
> I ask because you might run into the same problem that made it necessary to create KnowledgeSearch in the first place: 20 knowledge apps all starting up their own search providers on DBus at the same time was a performance killer, and I expect the same thing to happen if 20 apps all start up their grand central providers at the same time too.

Hmm, I'm not sure I understand this point. The names will exist on the bus but the apps will only be started up in the background when GC wants to refresh its content.

.... that said, I do wonder if it makes more sense to have *one* service which has access to all the content across the apps, so that we're only booting once instance of that and not 20 instances of Gjs running knowledge apps. If we did want to go down that road, how would we implement something like that.

> Another question, is "grand central" going to be a user- or developer-facing name for this feature? As far as I know it's internal jargon right now, but this pull request makes it part of the API that third-party developers will have to deal with.

Its definitely developer facing and meant to be "hidden" from the user. At least for third party developers of modular apps, it shouldn't matter at all to them since everything is handled through knowledge-lib.

smspillaz:
Closing this PR in favour of https://github.com/endlessm/eos-knowledge-services/pull/6

PR #1394, opened 2017-04-07
Custom web overrides CSS

ptomato:
https://phabricator.endlessm.com/T16145

### DIFF
--- a/js/app/application.js
+++ b/js/app/application.js
@@ -93,7 +93,7 @@ const Application = new Knowledge.Class({
                              'Path to a overrides scss or css file to theme the application', null);
         this.add_main_option('app-json-path', 'J'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.FILENAME,
                              'Path to a yaml or json file to use as a preset', null);
-        this.add_main_option('dummy-content', 'C'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
+        this.add_main_option('dummy-content', 'm'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.NONE,
                              'Whether to use dummy content from Moltres', null);
         this.add_main_option('content-path', 'p'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.FILENAME,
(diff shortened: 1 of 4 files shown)


ptomato:
Thanks for catching that.

I've come to an implementation that I think is better now, because in all cases it is applied on top of the existing CSS rather than resetting it.

tchx84:
@ptomato this looks a lot better, one doubt: as it is, this won't work for HTML articles that are server-template'd (e.g. publmetro app). Shouldn't we add a default renderer just to add our custom css? To cover those as well?

data/templates/legacy-article.mst
```diff
@@ -1,6 +1,9 @@
 {{#css-files}}
 <link rel="stylesheet" type="text/css" href="resource:///com/endlessm/knowledge/data/templates/css/{{{.}}}"/>
 {{/css-files}}
+{{#custom-css-files}}
+<link rel="stylesheet" type="text/css" href="{{{.}}}"/>
+{{/custom-css-files}}
```
tchx84: Is this really necessary? it looks to me like adding to the wrapper is enough (?), and will simplify this patch.

ptomato: Yes, you are right! Good catch. I think I overlooked this solution before because I originally wanted the custom file to replace, not augment.

js/app/application.js
```diff
@@ -91,9 +91,12 @@ const Application = new Knowledge.Class({
                              'Path to a different gresource to use with the application', null);
         this.add_main_option('theme-overrides-path', 'O'.charCodeAt(), GLib.OptionFlags.NONE, GLib.OptionArg.FILENAME,
                              'Path to a overrides scss or css file to theme the application', null);
+        this.add_main_option('web-overrides-path', 'w'.charCodeAt(),
+            GLib.OptionFlags.NONE, GLib.OptionArg.FILENAME,
```
ebassi: The indentation is a bit meh; everything else aligns to the first argument.

ebassi: The option flags are also on the same line as `add_main_option()` for the other command line arguments.

ptomato:
Merging with the style fix, as per discussion in Slack.

PR #1395, opened 2017-04-11
T16557 (new selection for static pages)

tchx84:
https://phabricator.endlessm.com/T16557

### DIFF
--- a/js/app/modules/selection/staticArticles.js
+++ b/js/app/modules/selection/staticArticles.js
@@ -1,0 +1,34 @@
+/* exported StaticArticles */
+
+// Copyright 2017 Endless Mobile, Inc.
+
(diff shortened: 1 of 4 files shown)


js/app/modules/selection/staticArticles.js
```diff
@@ -0,0 +1,34 @@
...
+
+const Module = imports.app.interfaces.module;
+const Xapian = imports.app.modules.selection.xapian;
+
+/**
+ * Class: StaticArticles
+ *
+ * Retrieves articles tagged with EknStaticTag
+ */
+const StaticArticles = new Module.Class({
```
ptomato: We really need https://phabricator.endlessm.com/T13175 so we can stop the proliferation of selections, and have this just be an instance of Filter.Tag or something like that.

PR #1398, opened 2017-04-14
control: Make eos-knowledge-0-bin arch-dependent

ptomato:
It writes an arch-dependent library path into its scripts, so it can't be
arch-independent.

https://phabricator.endlessm.com/T14833


tchx84:
**CR PASS**

PR #1399, opened 2017-04-17
Documentation with Hotdoc

ptomato:
Removes the Naturaldocs documentation and generates the documentation using Hotdoc and the hotdoc-modular-framework extension.

https://phabricator.endlessm.com/T14833


Makefile.am
```diff
@@ -410,6 +410,30 @@ EXTRA_DIST += \
...
+	docs/api/hotdoc.json \
+	docs/api/index.md \
+	docs/api/sitemap.txt \
+	docs/framework/hotdoc.json \
+	docs/framework/index.md \
+	docs/framework/sitemap.txt \
+	docs/framework/tools.md \
+	$(NULL)
+
+HOTDOC_PROJECTS = docs/framework
```
tchx84: where is this being used? Any reason why not generating docs/api?

tchx84:
@ptomato 

I tried with 

```
./autogen.sh --prefix=$HOME/install --enable-documentation
```

but then `make distcheck` fails with:

```
ERROR: files left in build directory after distclean:
./tools/introspect-local
Makefile:2537: recipe for target 'distcleancheck' failed
make[1]: *** [distcleancheck] Error 1
make[1]: Leaving directory '/home/martin/checkout/eos-knowledge-lib/eos-knowledge-0.4.0/_build'
Makefile:2464: recipe for target 'distcheck' failed
make: *** [distcheck] Error 1

```

Makefile.am
ptomato: This is used inside `HOTDOC_MAKEFILE`.

I skipped generating `docs/api` so far, since we have almost nothing there and we're not going to publish it for the time being.

ptomato:
Distcheck should be fixed now.

tchx84:
@ptomato something odd, should the docs be generated even without --enable-documentation option? I am pretty sure it didn't the last time i tried, but now, even if I don't specify it, it keeps generating it.

```
./autogen.sh --prefix=$HOME/install && make install # this generates the doc
```

is this correct?

PR #1400, opened 2017-04-18
T16557

tchx84:
https://phabricator.endlessm.com/T16557

### DIFF
--- a/data/css/blog.scss
+++ b/data/css/blog.scss
@@ -1,0 +1,126 @@
+@import "base";
+
+$highlighted-background-color: darken($background-dark-color, 5%) !default;
+
+.LayoutSidebar {
+  background-color: $background-dark-color;
+  -EknLayout_Sidebar-sidebar-width-small: 252px;
+  -EknLayout_Sidebar-sidebar-width-large: 252px;
+
+  .sidebar {
+    .main-menu {
+      margin: 58px 0px 0px 0px;
+    }
+
+    .CardTitle {
+      padding: 8px 0px 8px 40px;
+
+      &__title {
+        padding: 0px;
+        color: $primary-medium-color;
+      }
+
+      &.highlighted,
+      &.highlighted:hover {
+        background-color: $highlighted-background-color;
+
+        .CardTitle__title {
+          color: $primary-dark-color;
+        }
+      }
+
+      &:hover {
+        background-color: transparent;
+
+        .CardTitle__title {
+          color: $accent-light-color;
+        }
+      }
+    }
+  }
+}
+
+.PagerSimple {
+  background-color: $background-light-color;
+}
+
+.main-arrangement {
+  margin-bottom: 20px;
+}
+
+.CardDefaultFamily {
+  min-height: 400px;
+
+  &.width-f {
+    .CardDefaultFamily__title {
+      font-size: 32px;
+    }
+
+    .CardDefaultFamily__synopsis {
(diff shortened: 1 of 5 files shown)


data/preset/blog.yaml
```diff
@@ -0,0 +1,181 @@
...
+                                - 0
+                                - 2
+                        selection:
+                          type: Selection.AllHighlights
+                          slots:
+                            filter: 'Filter.Sets(invert: true)'
+                            order: 'Order.PublishedDate(ascending: false)'
+
+root:
+  type: Controller.Buffet
```
ptomato: Consider `Controller.Mesh`? As far as I can tell from looking at the code, the only thing `Controller.Buffet` has that `Controller.Mesh` doesn't, is support for a context bar, which we don't use.

data/preset/blog.yaml
```diff
@@ -0,0 +1,181 @@
...
+                  slots:
+                    card: Card.List
+                selection:
+                  type: Selection.Search
+                  id: search-results
+                no-results:
+                  shortdef: 'Layout.Box(orientation: vertical)'
+                  slots:
+                    contents:
+                    - type: ContentGroup.NoResultsMessage
```
ptomato: It occurred to me much later after you asked this morning, that you could just stick `ContentGroup.StaticText` here if you wanted to change the text.

data/css/blog.scss
```diff
@@ -0,0 +1,111 @@
+@import "project-setup";
```
ptomato: Anything from this file we can put in the default theme? This contains the things you added to "make it pretty" but if there's anything we can add that makes the default theme prettier by default, then we should do that.

data/css/blog.scss
```diff
@@ -0,0 +1,111 @@
...
+  background-color: $background-dark-color;
+  -EknLayout_Sidebar-sidebar-width-small: 252px;
+  -EknLayout_Sidebar-sidebar-width-large: 252px;
+
+  .sidebar {
+    .main-menu {
+      margin: 66px 0px 0px 0px;
+    }
+
+    .CardTitle {
```
ptomato: Seems like you should be able to deduplicate `.CardTitle` with some nested rules here.

data/preset/blog.yaml
tchx84: I feel very silly now, thanks :)

data/preset/blog.yaml
tchx84: hmm, we do use the navigation card at the bottom for now, that only works with Buffet controller.

data/css/blog.scss
tchx84: I suspect we don't even need that file (at all) anymore, ill remove from here for now.

tchx84:
@ptomato fixed your comments plus other improvements from latest designs

data/preset/blog.yaml
ptomato: I opened an issue to get rid of the cases where one thing is not supported by one controller.

tchx84:
> I noticed that https://phabricator.endlessm.com/T13175 is a real problem when scrolling down all the way in the sorted "feed". We may have to consider this as a higher priority.

@ptomato I can work on that next, if you think we should.

tchx84:
@ptomato addressed all your comments and created an issue for the default theme styling refactors https://phabricator.endlessm.com/T16689.

PR #1401, opened 2017-04-18
T16565

tchx84:
https://phabricator.endlessm.com/T16565

### DIFF
--- a/data/widgets/card/list.ui
+++ b/data/widgets/card/list.ui
@@ -92,6 +92,19 @@
                 </style>
               </object>
             </child>
+            <child>
+              <object class="GtkBox" id="context-box">
+                <property name="no-show-all">True</property>
+                <property name="can_focus">False</property>
+                <property name="expand">True</property>
+                <property name="halign">start</property>
+                <property name="valign">end</property>
+                <style>
+                  <class name="CardList__context"/>
+                  <class name="Card__context"/>
+                </style>
+              </object>
+            </child>
           </object>
         </child>
         <child>
--- a/js/app/modules/card/list.js
+++ b/js/app/modules/card/list.js
@@ -30,11 +30,20 @@ const List = new Module.Class({
             'Show synopsis', 'Whether to show the synopsis label',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             true),
+        /**
+         * Property: show-context
+         * Whether to show the context label.
+         */
+        'show-context': GObject.ParamSpec.boolean('show-context',
+            'Show context', 'Whether to show the context label',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            false),
     },
 
     Template: 'resource:///com/endlessm/knowledge/data/widgets/card/list.ui',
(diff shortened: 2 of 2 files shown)


data/widgets/card/list.ui
```diff
@@ -92,6 +92,18 @@
                 </style>
               </object>
             </child>
+            <child>
+              <object class="GtkFrame" id="context-frame">
```
xjuan: Any particular reason why to use a frame and not a box?
theses days all widgets can be styled, also this widget should be no-show-all

js/app/modules/card/list.js
```diff
@@ -30,11 +30,20 @@ const List = new Module.Class({
             'Show synopsis', 'Whether to show the synopsis label',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             true),
+        /**
+         * Property: show-context
+         * Whether to show the context label.
+         */
+        'show-context': GObject.ParamSpec.boolean('show-context',
+            'Show context', 'Whether to show the context label',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
```
xjuan: I think it might be useful to be able to change this at runtime, dont you?

js/app/modules/card/list.js
```diff
@@ -54,6 +63,12 @@ const List = new Module.Class({
         if (this.model.duration) {
             this.set_duration_label(this._context_label, this.model.duration);
         }
+
+        if (this.show_context) {
+            this._context_widget = this.create_context_widget_from_model();
+            this._context_frame.add(this._context_widget);
```
xjuan: you are not showing the widget, so you rely on the API user to call show_all() for this to work

data/widgets/card/list.ui
tchx84: No particular reason really, except that this is what we do for the others cards e.g. https://github.com/endlessm/eos-knowledge-lib/blob/master/data/widgets/card/defaultFamily.ui#L31

So I am just duplicating it for consistency.

js/app/modules/card/list.js
tchx84: I am not really sure we can do that, since in at least one case, this is determined by designers during the design of the template/app.

js/app/modules/card/list.js
tchx84: This actually rely on https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/moduleFactory.js#L121

js/app/modules/card/list.js
ptomato: At least for now, we make all properties construct-only if you can set them in the YAML

data/widgets/card/list.ui
xjuan: fair enough, we just need to keep in mind that the looks of a GtkFrame might change, and if all we want is a placeholder, its usually better to use a box or a grid

js/app/modules/card/list.js
xjuan: no problem, we can remove the construct only flag when the time comes

xjuan:
CR Pass

PR #1402, opened 2017-04-19
T16666

tchx84:
https://phabricator.endlessm.com/T16666


tchx84:
This removes the old (now unnecessary) compass setup file http://compass-style.org/help/documentation/sass-based-configuration-options/

PR #1410, opened 2017-05-01
T16788

katelk28:
This allows Card.Title's title to wrap by adding a proxy property called "max-title-lines" to Card.Title, so setting the property on Card.Title will change the card's internal title_label's "lines" property.
Default value of max-title-lines = 1, for most uses of Card.Title.

Also added the max-title-lines property to Banner.Set's Card.Title in library-list & Preset B's set-page.

https://phabricator.endlessm.com/T16788

### DIFF
--- a/js/app/modules/card/title.js
+++ b/js/app/modules/card/title.js
@@ -1,6 +1,7 @@
 // Copyright 2014 Endless Mobile, Inc.
 
 const Gdk = imports.gi.Gdk;
+const GLib = imports.gi.GLib;
 const GObject = imports.gi.GObject;
 const Gtk = imports.gi.Gtk;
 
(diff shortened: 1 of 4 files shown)


js/app/modules/card/title.js
```diff
@@ -41,6 +42,22 @@ const Title = new Module.Class({
             'Decorate on highlight', 'Whether to draw a custom decoration when the card is highlighted',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             false),
+
+        /*
+         * Propety: lines
```
katelk28: Thanks, good to know. And my bad on the name change -- I originally just called it "lines" but figured it might make more sense to call it max-title-lines, and forgot to change it in the documentation. 

What do you think about that, actually? Maybe it should just be "lines" to be consistent with other places we use it?

katelk28: @ptomato And to your overall comment, yes, I tested all other templates to check, and the default value of 1 line is working as intended

ptomato: I thought about that, but I think at least having `title` in the name is good since it refers to the title. This card only has a title, but if we wanted to have a similar property on the other cards, it would be good if it were called the same thing. Otherwise, up to you.

PR #1411, opened 2017-05-03
T16869

tchx84:
https://phabricator.endlessm.com/T16869

### DIFF
--- a/data/templates/article-wrapper.mst
+++ b/data/templates/article-wrapper.mst
@@ -15,6 +15,9 @@
         <script type="text/javascript" src="resource:///com/endlessm/knowledge/data/templates/js/{{{.}}}"></script>
         {{/javascript-files}}
 
+        <script type="text/javascript">
+            window.contentMetadata = {{{content-metadata}}};
+        </script>
         {{{content}}}
         <script type="text/javascript">
             window.crosslink_init({{{crosslink-data}}});
--- a/data/templates/news-article.mst
+++ b/data/templates/news-article.mst
@@ -1,37 +1,55 @@
-{{#extra-header-information}}
 <div class="extra-header">
-    <div class="context"><p>{{{context}}}</p></div>
+    <div class="context">
+        <a class="eos-show-link" id="context-link"></a>
+    </div>
     <div class="extra-header-right">
-        {{#date-published}}
-        <p class="date-published">{{.}}</p>
-        {{/date-published}}
-        {{#source-link}}
+        <p class="date-published" id="date-published"></p>
         <p class="source-link">
-            {{{.}}}
-            {{#author}}
-            &mdash; {{.}}
-            {{/author}}
+            <a class="eos-show-link" id="source-link">Prensalibre.com</a>
+            &mdash; <span id="authors"></span>
         </p>
-        {{/source-link}}
     </div>
 </div>
-{{/extra-header-information}}
+
 <section id="inside-content" class="section">
     <div>{{{body-html}}}</div>
     <hr>
-    {{#disclaimer}}
     <div class="disclaimer">
-        <p>{{{.}}}</p>
+        <p>
+        Leer más en <a class="eos-show-link" id="disclaimer-link">Prensalibre.com</a><br>
+        <a class="eos-modal-link" href="#modal"> Aviso Legal y Política de Propiedad Intelectual </a>
+        </p>
     </div>
-    {{/disclaimer}}
     <div class="eos-modal" id="modal">
         <a class="eos-modal-close-background" href="#close"></a>
         <div>
-            <a class="eos-modal-close-button" href="#close"> &times </a>
-            <p align="justify">{{disclaimer-window}}</p>
+            <a class="eos-modal-close-button" href="#close"> &times; </a>
+            <p align="justify">La presente aplicación incluye reportes de noticias de interés que han sido obtenidos automáticamente de sus respectivas fuentes por medio del formato RSS que cada una de ellas ha hecho públicamente disponibles. En cada una se hace constar el titular de los derechos de autor del contenido, incluyendo pero no limitado a las imágenes en ellos incluidas. Las opiniones e información expresadas en las distintas notas, artículos, anuncios o cartas son responsabilidad exclusiva de sus autores. Endless no se hace responsable de la veracidad, del contenido o calidad de la información, ni de los productos, servicios, ofertas, promociones u otras que puedan estar ahí incluidas. El contenido es proporcionado acá de manera gratuita para uso no comercial. El contenido de los mismos y la información ahí expresada es responsabilidad exclusiva de sus autores. No asumimos responsabilidad alguna por el contenido de los mismos ni por cualquier acción u omisión tomada con base en esa información. El contenido puede incluir vínculos a páginas web de terceros que no se encuentran bajo nuestro control. No nos hacemos responsables por el contenido de esas páginas ni la inclusión de la información o de los vínculos constituye aceptación de las mismas, de sus autores o de los medios de comunicación que las han hecho públicas. Nos reservamos el derecho de modificar, suspender o descontinuar cualquier aspecto de esta aplicación y la información acá contenida en cualquier momento.</p>
         </div>
     </div>
 </section>
+
+<script>
+let metadata = window.contentMetadata;
+
+let set = metadata.sets.filter(set => set.featured)[0];
+let set_link = document.getElementById('context-link');
+set_link.href = set.id;
+set_link.innerHTML = set.title.toLowerCase();
+
+let published = document.getElementById('date-published');
+published.innerHTML = new Date(metadata.published).toLocaleDateString();
+
+let source_link = document.getElementById('source-link');
+source_link.href = metadata.original_uri;
+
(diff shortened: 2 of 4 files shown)


ptomato:
This seems quite inflexible to me. You can only render context in the specific way we plan to do it for the blog articles.

Also you have to basically inject HTML into (unstyled?) `<p>` containers, which will of course lead to things like `</p><ul><li style="color:red;">context1</li><li style="color:blue;">context2</li></ul><p>` being cooked up in the code :smile:

I think a good starting point for making this more flexible would be to consider how to absorb the Prensa Libre-specific HTML addition into this API.

tchx84:
@ptomato 

> This seems quite inflexible to me. You can only render context in the specific way we plan to do it for the blog articles.

Just to clarify, are you referring strictly to the implementation or the design of this context "addon" to the articles? 

In terms of implementation, for this first approach I just wanted something simple. If we want to make this all-flexible then we would need to provide API to:

1. Specify where this context should be shown, something like "top, bottom, both, none"
2. Tell, for each individual place, when we want to filter these sets and how, something like, "all, featured-only, non-featured-only, whatever else".

I think this would lead to a kind of complex API, possibly having to pass a many properties from the YAML to Card.KnowledgeDocument, through other objects to the ArticleHTMLRenderer. But maybe there is a better approach that doesn't require keeping conventions between the content and the HTML renderer.

Do you have another idea in mind? I mean, as-in a completely different approach?

> Also you have to basically inject HTML into (unstyled?) <p> containers, which will of course lead to things like </p><ul><li style="color:red;">context1</li><li style="color:blue;">context2</li></ul><p> being cooked up in the code :smile:

Well, I originally though we would just need to add the styling as part of the content CSS, but if that doesn't work, we could provide a default css for the context, like in a wrapper CSS.

> I think a good starting point for making this more flexible would be to consider how to absorb the Prensa Libre-specific HTML addition into this API.

We could do this, but only for the context.

But keep in mind this prensa libre custom stuff not only provides the context but also provides the publishing, author and licensing information (something we already provide as part of the context in other news app). If we want to remove all this custom stuff (which I think is a good idea) we need to do those other changes both in design and ingestion.

tchx84:
@ptomato to test this I prepared a "hello world" using basin:

1. download and extract [files.tar.gz](https://github.com/endlessm/eos-knowledge-lib/files/987214/files.tar.gz)
2. create the shard using `basin-helper files .`
3. to test it 'com.endlessm.prensa_libre.es_GT --content-path=/path/to/content'

js/app/articleHTMLRenderer.js
```diff
@@ -266,6 +266,36 @@ const ArticleHTMLRenderer = new Knowledge.Class({
...
+        let metadata = {
+            id: model.ekn_id,
+            title: model.title,
+            published: model.published,
+            authors: model.authors,
+            license: model.license,
+            source: model.source,
+            source_name: model.source_name,
+            source_uri: model.original_uri,
+            sets: [],
```
ptomato: Why the different property name here?

data/templates/article-wrapper.mst
```diff
@@ -15,6 +15,7 @@
         <script type="text/javascript" src="resource:///com/endlessm/knowledge/data/templates/js/{{{.}}}"></script>
         {{/javascript-files}}
 
+        <script type="application/json" id="content-metadata">{{{content-metadata}}}</script>
```
ptomato: You could consider making this a `<script type="text/javascript">` and prefixing the JSON object with something like `window.contentMetadata = `. That would make it much easier to access it from browser-side JS, but a disadvantage is that it would pollute the global namespace.

js/app/articleHTMLRenderer.js
tchx84: because we store/provide sets (with its properties) and not just simply plain tags, so I think we should make that differentiation.

data/templates/article-wrapper.mst
tchx84: @ptomato you mean something like

```
<script type="text/javascript">
window.contentMetadata = JSON.parse({{{content-metadata}}});
</script>
```

So they just need to window.contentMetadata ? (and not having to JSON.parse each time?)

js/app/articleHTMLRenderer.js
ptomato: This comment got on the wrong line - I meant `source_uri` vs `original_uri`

data/templates/article-wrapper.mst
ptomato: Yes, indeed.

ptomato: Actually you don't need the `JSON.parse` in this case - you are just generating code directly.

tchx84:
@ptomato new example (still without thumbnails) but it using window.contentMetadata now.
[files.tar.gz](https://github.com/endlessm/eos-knowledge-lib/files/987693/files.tar.gz)

js/app/articleHTMLRenderer.js
tchx84: oh! arg, I didn't notice! let me push the fix.

tests/js/app/testArticleHTMLRenderer.js
```diff
@@ -74,8 +74,8 @@ describe('Article HTML Renderer', function () {
         let html_no_title = renderer.render(wikibooks_model);
         renderer.show_title = true;
         let html_with_title = renderer.render(wikibooks_model);
-        expect(html_with_title).toMatch('Wikibooks title');
-        expect(html_no_title).not.toMatch('Wikibooks title');
+        expect(html_with_title).toMatch('<h1>Wikibooks title</h1>');
```
ptomato: Does this really need to be changed? `<h1>` is presentation information so I'd recommend keeping it out of here if possible.

data/templates/news-article.mst
```diff
@@ -1,37 +1,55 @@
...
         </div>
     </div>
 </section>
+
+<script>
+let metadata = window.contentMetadata;
+
+let set = metadata.sets.filter(set => set.featured)[0];
```
ptomato: I assume you checked the arrow function works in browser JS? I'm pretty sure it does, though I never tried.

data/templates/news-article.mst
```diff
@@ -1,37 +1,55 @@
...
+        <p>
+        Leer más en <a class="eos-show-link" id="disclaimer-link">Prensalibre.com</a><br>
+        <a class="eos-modal-link" href="#modal"> Aviso Legal y Política de Propiedad Intelectual </a>
+        </p>
     </div>
-    {{/disclaimer}}
     <div class="eos-modal" id="modal">
         <a class="eos-modal-close-background" href="#close"></a>
         <div>
             <a class="eos-modal-close-button" href="#close"> &times </a>
```
ptomato: today I learned that the semicolon is optional on character entities for historical reasons :stuck_out_tongue_closed_eyes:

data/templates/news-article.mst
```diff
@@ -1,37 +1,55 @@
...
-{{/extra-header-information}}
+
 <section id="inside-content" class="section">
     <div>{{{body-html}}}</div>
     <hr>
-    {{#disclaimer}}
     <div class="disclaimer">
-        <p>{{{.}}}</p>
+        <p>
+        Leer más en <a class="eos-show-link" id="disclaimer-link">Prensalibre.com</a><br>
```
ptomato: Yes, it makes sense to me to remove this from internationalization!

Can you remove gettext from articleHTMLRenderer now, and remove it from POTFILES.in? Or are there still other gettext strings left?

data/templates/news-article.mst
```diff
@@ -1,37 +1,55 @@
...
-    <div class="context"><p>{{{context}}}</p></div>
+    <div class="context">
+        <a class="eos-show-link" id="context-link"></a>
+    </div>
     <div class="extra-header-right">
-        {{#date-published}}
-        <p class="date-published">{{.}}</p>
-        {{/date-published}}
-        {{#source-link}}
+        <p class="date-published" id='date-published'></p>
```
ptomato: nitpick: be consistent with quotes

tests/js/app/testArticleHTMLRenderer.js
tchx84: We now inject the metadata into the article, this means that the title string will always be present in the content, therefore, this test needs to differentiate between "just matching the string" vs "matching the string when is presented as the title", that is why adding that presentation detail is necessary.

data/templates/news-article.mst
tchx84: We can't, we still have other strings there like _("Copy"), among others related to legacy stuff.

data/templates/news-article.mst
tchx84: Well, I just added it, to be consistent with the `&mdash;`.

data/templates/news-article.mst
tchx84: It has too, otherwise this wouldn't have worked at all :) but yeah, I tested it.

tests/js/app/testArticleHTMLRenderer.js
ptomato: OK then, not ideal but I see the reasoning

PR #1413, opened 2017-05-03
T16845

tchx84:
https://phabricator.endlessm.com/T16845


ptomato:
I can't easily test this since `Selection.StaticArticles` isn't in the runtime yet, but looks good.

PR #1417, opened 2017-05-09
T16765

smspillaz:
http://phabricator.endlessm.com/T16765

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -743,8 +743,8 @@ query_state_free (gpointer data)
   g_slice_free (QueryState, state);
 }
 
-typedef void (*ScatterTaskCountdownCompleteCallback)(gpointer user_data);
-typedef void (*ScatterTaskCountdownErrorCallback)(GError *error, gpointer user_data);
+typedef void (*ScatterTaskCountdownCompleteCallback) (gpointer user_data);
+typedef void (*ScatterTaskCountdownErrorCallback) (GError *error, gpointer user_data);
 
 typedef struct {
   guint remaining;
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -759,7 +759,7 @@ scatter_task_countdown_new (guint remaining,
                             ScatterTaskCountdownErrorCallback error_callback,
                             gpointer data)
 {
-  ScatterTaskCountdown *countdown = g_new0(ScatterTaskCountdown, 1);
+  ScatterTaskCountdown *countdown = g_new0 (ScatterTaskCountdown, 1);
   countdown->remaining = remaining;
   countdown->callback = callback;
   countdown->error_callback = error_callback;
@@ -804,7 +804,7 @@ object_response_data_new (ScatterTaskCountdown *countdown,
                           GPtrArray *models,
                           guint index)
 {
-  ObjectResponseData *data = g_new0(ObjectResponseData, 1);
+  ObjectResponseData *data = g_new0 (ObjectResponseData, 1);
   data->countdown = countdown;
   data->domain = domain;
   data->models = models;
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -984,7 +984,7 @@ ptr_array_to_list_deep (GPtrArray *array)
 
   for (guint i = 0; i < array->len; ++i)
     {
-      list = g_list_append (list, g_object_ref(g_ptr_array_index(array, i)));
+      list = g_list_append (list, g_object_ref (g_ptr_array_index(array, i)));
     }
 
   return list;
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -743,26 +743,23 @@ query_state_free (gpointer data)
   g_slice_free (QueryState, state);
 }
 
-typedef void (*ScatterTaskCountdownCompleteCallback) (gpointer user_data);
-typedef void (*ScatterTaskCountdownErrorCallback) (GError *error, gpointer user_data);
+typedef void (*ScatterTaskCountdownCompleteCallback) (GError *error, gpointer user_data);
 
 typedef struct {
   guint remaining;
   ScatterTaskCountdownCompleteCallback callback;
-  ScatterTaskCountdownErrorCallback error_callback;
   gpointer data;
+  GError *error;
 } ScatterTaskCountdown;
 
 static ScatterTaskCountdown *
 scatter_task_countdown_new (guint remaining,
                             ScatterTaskCountdownCompleteCallback callback,
-                            ScatterTaskCountdownErrorCallback error_callback,
                             gpointer data)
 {
   ScatterTaskCountdown *countdown = g_new0 (ScatterTaskCountdown, 1);
   countdown->remaining = remaining;
   countdown->callback = callback;
-  countdown->error_callback = error_callback;
   countdown->data = data;
 
   return countdown;
@@ -775,15 +772,15 @@ scatter_task_countdown_subtask_completed (ScatterTaskCountdown **countdown_ptr,
   ScatterTaskCountdown *countdown = *countdown_ptr;
 
   /* If we received an error we should still continue to countdown, but
-   * allow the creator of the countdown to handle the error */
-  if (error && countdown->error_callback)
-    {
-      (*countdown->error_callback) (error, countdown->data);
-    }
+   * store the error so that the caller can deal with it */
+  if (!countdown->error)
+    countdown->error = error;
+  else
+    g_clear_error (&error);
 
   if (--countdown->remaining == 0)
     {
-      (*countdown->callback) (countdown->data);
+      (*countdown->callback) (error, countdown->data);
       g_free (countdown);
       *countdown_ptr = NULL;
     }
@@ -813,29 +810,15 @@ object_response_data_new (ScatterTaskCountdown *countdown,
 }
 
 static void
-on_received_all_model_objects (gpointer user_data)
+on_received_all_model_objects (GError *error, gpointer user_data)
 {
-  GTask *task = user_data;
+  g_autoptr(GTask) task = user_data;
 
-  /* We might have already been marked as completed earlier due to an
-   * error state. Don't return if that is the case */
-  if (g_task_get_completed (task))
-    return;
-
-  g_task_return_boolean (task, TRUE);
-  g_object_unref (task);
-}
-
-static void
-on_receive_model_error (GError *error, gpointer user_data)
-{
-  GTask *task = user_data;
-
-  if (error != NULL)
-    {
-      g_task_return_error (task, error);
-      return;
-    }
+  /* Received an error, propogate it to the caller */
+  if (error)
+    g_task_return_error (task, error);
+  else
+    g_task_return_boolean (task, TRUE);
 }
 
 static void
@@ -915,7 +898,6 @@ on_xapian_query_response (GObject *source,
    * we are done receiving all model objects */
   ScatterTaskCountdown *countdown = scatter_task_countdown_new (state->total_models,
                                                                 on_received_all_model_objects,
-                                                                on_receive_model_error,
                                                                 g_object_ref (task));
 
   for (guint i = 0; i < json_array_get_length (results_array); i++)
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -812,7 +812,7 @@ object_response_data_new (ScatterTaskCountdown *countdown,
   return data;
 }
 
-void
+static void
 on_received_all_model_objects (gpointer user_data)
 {
   GTask *task = user_data;
@@ -826,7 +826,7 @@ on_received_all_model_objects (gpointer user_data)
   g_object_unref (task);
 }
 
-void
+static void
 on_receive_model_error (GError *error, gpointer user_data)
 {
   GTask *task = user_data;
@@ -838,7 +838,7 @@ on_receive_model_error (GError *error, gpointer user_data)
     }
 }
 
-void
+static void
 on_object_response (GObject *source,
                     GAsyncResult *result,
                     gpointer user_data)
@@ -853,7 +853,7 @@ on_object_response (GObject *source,
   scatter_task_countdown_subtask_completed (&object_response_data->countdown, error);
 }
 
-void
+static void
 maybe_unref_object (gpointer data)
 {
   if (data)
@@ -862,7 +862,7 @@ maybe_unref_object (gpointer data)
     }
 }
 
-void
+static void
 on_xapian_query_response (GObject *source,
                           GAsyncResult *result,
                           gpointer user_data)
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -978,9 +978,7 @@ ptr_array_to_list_deep (GPtrArray *array)
 
   /* If there's no array just return an empty list */
   if (!array)
-    {
-      return list;
-    }
+    return NULL;
 
   for (guint i = 0; i < array->len; ++i)
     {
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -773,14 +773,12 @@ scatter_task_countdown_subtask_completed (ScatterTaskCountdown **countdown_ptr,
 
   /* If we received an error we should still continue to countdown, but
    * store the error so that the caller can deal with it */
-  if (!countdown->error)
-    countdown->error = error;
-  else
-    g_clear_error (&error);
+  if (!countdown->error && error)
+    countdown->error = g_error_copy (error);
 
   if (--countdown->remaining == 0)
     {
-      (*countdown->callback) (error, countdown->data);
+      (*countdown->callback) (countdown->error, countdown->data);
       g_free (countdown);
       *countdown_ptr = NULL;
     }
@@ -828,8 +826,11 @@ on_object_response (GObject *source,
 {
   g_autofree ObjectResponseData *object_response_data = user_data;
   EkncDomain *domain = object_response_data->domain;
-  GError *error = NULL;
+  g_autoptr(GError) error = NULL;
 
+  /* We take ownership of the returned error here and
+   * scatter_task_countdown_completed may make a copy of it if
+   * it needs to */
   EkncContentObjectModel *model = eknc_domain_get_object_finish (domain, result, &error);
 
   *(&g_ptr_array_index (object_response_data->models, object_response_data->index)) = model;
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -981,7 +981,7 @@ ptr_array_to_list_deep (GPtrArray *array)
     return NULL;
 
   for (guint i = 0; i < array->len; ++i)
-    list = g_list_append (list, g_object_ref (g_ptr_array_index(array, i)));
+    list = g_list_append (list, g_object_ref (g_ptr_array_index (array, i)));
 
   return list;
 }
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -857,9 +857,7 @@ static void
 maybe_unref_object (gpointer data)
 {
   if (data)
-    {
-      g_object_unref (G_OBJECT(data));
-    }
+    g_object_unref (data);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


ekncontent/ekncontent/eknc-domain.c
```diff
@@ -891,7 +938,7 @@ eknc_domain_query_finish (EkncDomain *self,
   QueryState *state = g_task_get_task_data (task);
   return g_object_new (EKNC_TYPE_QUERY_RESULTS,
                        "upper-bound", state->upper_bound,
-                       "models", state->models,
+                       "models", ptr_array_to_list_shallow (state->models),
```
smspillaz: Going from a GPtrArray to a GList like this is not ideal, though I'm not sure if EkncQueryResults' properties are considered to be "external API" that we can't break.

ptomato: No, we're not making any API guarantees other than "it will be really inconvenient if the way it's accessed from JS changes" :smile:

However, I'm not sure if gobject-introspection supports introspecting the type of a GPtrArray-valued property, whereas with a GList it does.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -730,44 +730,69 @@ eknc_domain_get_fixed_query_finish (EkncDomain *self,
...
   guint upper_bound;
+  guint num_received;
 } QueryState;
 
 static void
 query_state_free (gpointer data)
 {
   QueryState *state = data;
-  g_slist_free_full (state->models, g_object_unref);
+  g_clear_pointer (&state->models, g_ptr_array_unref);
```
ptomato: I think this leaks the objects in `state->models` in the case where they're not marshalled into the GList

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -858,13 +891,27 @@ eknc_domain_query (EkncDomain *self,
 
   GTask *task = g_task_new (self, cancellable, callback, user_data);
   QueryState *state = g_slice_new0 (QueryState);
+  state->models = NULL;
```
ptomato: Should not be necessary since the previous line already zeroes out the whole struct

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -830,8 +858,13 @@ on_xapian_query_response (GObject *source,
           return;
         }
 
+      /* Note that we pass another closure here instead of just the task
```
ptomato: You don't have to encapsulate the task in a closure data struct - use `g_task_set_task_data` instead.

ekncontent/ekncontent/eknc-domain.c
smspillaz: I think that's right. The memory ownership semantics around this point got a little confusing - I initially tried having the pointer array reduce the refcount on the models when it was destroyed though that caused it to free memory that was being used elsewhere. Would it make more sense to take a ref on each of the objects when marshalling them into the GList and then use `g_ptr_array_new_full` above?

ekncontent/ekncontent/eknc-domain.c
smspillaz: That's true - I think I was just being paranoid here but this can be removed.

ekncontent/ekncontent/eknc-domain.c
smspillaz: We're already setting the task-level data to `QueryState`, whereas these callbacks are sort of per-object. What I wanted to do was create a closure for each callback to indicate which index was in use, which I don't think I would be able to do on the `QueryState` object itself since that is shared between all callback invocations.

ptomato: Hmm, I see that now, but I'm still picking up a code smell from this. Maybe create a new task for these operations then, with its own task data that's not shared with `QueryState`?

smspillaz: We would still need to share some data with query state though - in particular, the number of models which have been fetched asynchronously.

smspillaz: Okay, I've updated this so as to separate out the part which handles the "countdown" from the part that actually returns the task. All the countdown knows about is an opaque gpointer and a callback that gets called when all the child tasks are completed. The caller passes in a callback and the GTask and returns it accordingly.

ekncontent/ekncontent/eknc-domain.c
smspillaz: I've done this, adding a function to unref any models which had a ref taken on them by the GPtrArray

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -891,7 +1022,7 @@ eknc_domain_query_finish (EkncDomain *self,
   QueryState *state = g_task_get_task_data (task);
   return g_object_new (EKNC_TYPE_QUERY_RESULTS,
                        "upper-bound", state->upper_bound,
-                       "models", state->models,
+                       "models", ptr_array_to_list_deep (state->models),
```
ptomato: Did you check if we can have a GPtrArray-valued property in gobject-introspection?

ekncontent/ekncontent/eknc-domain.c
ptomato: I'm *sure* there must be a better way to do this, and I'd even bet there is an example of it somewhere in GNOME code. In particular I don't like the separate success/error callbacks, as that duplicates `g_task_return_pointer()`/`g_task_return_error()`.

It seems to me the only data you need in `on_object_response()` is the `EkncDomain` (which would be the task's "source object") and the models array plus an index into it.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -730,45 +730,144 @@ eknc_domain_get_fixed_query_finish (EkncDomain *self,
 typedef struct
 {
   // List of EkncContentModel items
-  GSList *models;
+  GPtrArray *models;
   guint total_models;
   guint upper_bound;
+  guint num_received;
```
ptomato: This is no longer needed.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -830,8 +937,19 @@ on_xapian_query_response (GObject *source,
           return;
         }
 
+      /* Note that we pass another closure here instead of just the task
+       * containing the index of the response. This is done to preserve
+       * ordering since the callback order is not guaranteed.
+       *
+       * We need to take a ref on the task here so that it is preserved until
```
ptomato: The code that this comment refers to has moved elsewhere.

ekncontent/ekncontent/eknc-domain.c
smspillaz: Not yet - I kinda wanted to avoid the yak-shavey impact changing the API here might have. But I can look into it briefly.

smspillaz: Hmm, it doesn't seem to immediately work. GObject-Introspection doesn't complain and I'm able to pass in a GPtrArray into the property just fine. However, reading it out just gets an empty array (it seems like `get_property` is not even called).

I can move this into a separate patch and we can inspect as a low-priority thing there I suppose.

ptomato: No need to move it into a separate patch — this is apparently the only API that works with gobject-introspection :-)

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
 query_state_free (gpointer data)
 {
   QueryState *state = data;
-  g_slist_free_full (state->models, g_object_unref);
+  g_clear_pointer (&state->models, g_ptr_array_unref);
   g_slice_free (QueryState, state);
 }
 
+typedef void (*ScatterTaskCountdownCompleteCallback)(gpointer user_data);
+typedef void (*ScatterTaskCountdownErrorCallback)(GError *error, gpointer user_data);
```
cosimoc: Missing space before the argument list (both occurrences)

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
...
+  gpointer data;
+} ScatterTaskCountdown;
+
+static ScatterTaskCountdown *
+scatter_task_countdown_new (guint remaining,
+                            ScatterTaskCountdownCompleteCallback callback,
+                            ScatterTaskCountdownErrorCallback error_callback,
+                            gpointer data)
+{
+  ScatterTaskCountdown *countdown = g_new0(ScatterTaskCountdown, 1);
```
cosimoc: Missing space before paren

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
...
+scatter_task_countdown_subtask_completed (ScatterTaskCountdown **countdown_ptr,
+                                          GError *error)
+{
+  ScatterTaskCountdown *countdown = *countdown_ptr;
+
+  /* If we received an error we should still continue to countdown, but
+   * allow the creator of the countdown to handle the error */
+  if (error && countdown->error_callback)
+    {
+      (*countdown->error_callback)(error, countdown->data);
```
cosimoc: Missing space before argument list

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
...
+    }
+
+  if (--countdown->remaining == 0)
+    {
+      (*countdown->callback)(countdown->data);
```
cosimoc: Missing space before argument list

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
...
+  guint index;
+} ObjectResponseData;
+
+static ObjectResponseData *
+object_response_data_new (ScatterTaskCountdown *countdown,
+                          EkncDomain *domain,
+                          GPtrArray *models,
+                          guint index)
+{
+  ObjectResponseData *data = g_new0(ObjectResponseData, 1);
```
cosimoc: Missing space before paren

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
...
+  return data;
+}
+
+static void
+object_response_data_free (ObjectResponseData *data)
+{
+  g_free (data);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(ObjectResponseData, object_response_data_free);
```
cosimoc: Remove terminal semicolon

cosimoc: On the other hand, since the cleanup function is just `g_free()` you could just use `g_autofree` instead of the autoptr stuff?

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,36 +739,134 @@ static void
...
 
-  if (g_slist_length (state->models) == state->total_models)
+void
+maybe_unref_object (gpointer data)
+{
+  if (data)
     {
-      g_task_return_boolean (task, TRUE);
-      return;
+      g_object_unref (G_OBJECT(data));
```
cosimoc: This function should at the very least be `static`, but can't you avoid adding `NULL` to the array in the first place?
Also, you don't need to cast `data` to `G_OBJECT()` -- `g_object_unref()` takes a `void *`

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -865,6 +981,25 @@ eknc_domain_query (EkncDomain *self,
...
+
+  /* If there's no array just return an empty list */
+  if (!array)
+    {
+      return list;
+    }
+
+  for (guint i = 0; i < array->len; ++i)
+    {
+      list = g_list_append (list, g_object_ref(g_ptr_array_index(array, i)));
```
cosimoc: Missing space before paren

ekncontent/ekncontent/eknc-domain.c
smspillaz: Hmm, `g_new0` is a macro and my understanding is that macro calls can't contain any spaces between the macro name and arguments. Or maybe that was definitions - I'll just check with the space.

ekncontent/ekncontent/eknc-domain.c
smspillaz: Ah, we do use spaces for funcpointer calls.

ekncontent/ekncontent/eknc-domain.c
ptomato: That's definitions only

smspillaz:
@cosimoc all comments addressed

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -865,6 +971,25 @@ eknc_domain_query (EkncDomain *self,
...
 }
 
+static GList *
+ptr_array_to_list_deep (GPtrArray *array)
+{
+  GList *list = NULL;
+
+  /* If there's no array just return an empty list */
+  if (!array)
+    {
```
cosimoc: Remove curly braces for single-line if

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -865,6 +971,25 @@ eknc_domain_query (EkncDomain *self,
...
+      return list;
```
cosimoc: Just `return NULL` here

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -865,6 +971,25 @@ eknc_domain_query (EkncDomain *self,
...
+    }
+
+  for (guint i = 0; i < array->len; ++i)
+    {
```
cosimoc: Remove curly braces for single-line if

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -865,6 +971,25 @@ eknc_domain_query (EkncDomain *self,
...
+      list = g_list_append (list, g_object_ref (g_ptr_array_index(array, i)));
```
cosimoc: Missing space before paren

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -820,6 +908,16 @@ on_xapian_query_response (GObject *source,
       return;
     }
 
+  state->models = g_ptr_array_new_full (state->total_models, maybe_unref_object);
+  g_ptr_array_set_size (state->models, state->total_models);
```
cosimoc: There should be no reason to call this here -- even more so because the previous line already sets the size of the array to `state->total_models`

ekncontent/ekncontent/eknc-domain.c
cosimoc: I agree with Philip here... Also why is it important that ordering is preserved? Even if it's important, could we not resort the models later?

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,128 @@ static void
...
+  *(&g_ptr_array_index (object_response_data->models, object_response_data->index)) = model;
+  scatter_task_countdown_subtask_completed (&object_response_data->countdown, error);
 }
 
-void
+static void
+maybe_unref_object (gpointer data)
+{
+  if (data)
+    g_object_unref (data);
```
cosimoc: If ordering is not important, or if we could just use `g_ptr_array_add()` in the callback because we e.g. resort the models later, this function could go away

smspillaz:
> I agree with Philip here... Also why is it important that ordering is preserved? Even if it's important, could we not resort the models later?

Hmm, GitHub won't let me reply to this inline.

Preserving the ordering here is definitely important - we did all that work on the Xapian side to order the models and then by asynchronously fetching the model contents we're throwing that work away. We could re-order, but there doesn't seem to be much point to doing that when you already have things in the correct order anyway.

I take the point that re-ordering for 5 or 6 models is definitely a negligible cost. That said, I'm thinking of the case where we could be re-ordering hundreds or thousands of results, at which point you're paying an `n log n` cost for essentially keeping things the way they are.

The only tradeoff here is a little bit of extra complexity to ensure that we put models back in place in the correct order. But I think the complexity is worth it from a correctness and performance point of view. In addition, I don't think the additional complexity is really that burdensome - we were doing something similar anyway when appending the results to the list in that we needed to count-down until we had received all the models or errors. The big difference here is that we now have to send the correct index in a wrapper closure, but that's basically about it.

ekncontent/ekncontent/eknc-domain.c
smspillaz: I'm not sure I understand this point - is that because (if we don't care about ordering) we wouldn't have `NULL` slots for failed models, and thus we could set `g_object_unref` to be the finaliser func?

ekncontent/ekncontent/eknc-domain.c
smspillaz: Not quite - https://developer.gnome.org/glib/stable/glib-Pointer-Arrays.html#g-ptr-array-new-full does not actually set the *size* of the model, all it does is preallocate memory. The size will still be zero.

ekncontent/ekncontent/eknc-domain.c
cosimoc: Yeah, that would be the reason

ekncontent/ekncontent/eknc-domain.c
cosimoc: Oh, I see...  I guess you want the size to match `state->total_models` because of how you use `g_ptr_array_index()` in the callback, so this makes sense.

cosimoc:
@smspillaz Of course, I missed the fact that the sorting order is the entire reason we're doing this!
If possible, I would still try to find a way to remove the `ScatterTaskCountdown` abstraction and just leave the extra wrapper closure for the index, but otherwise your explanation makes sense to me.

smspillaz:
@ptomato Okay, I've given merging the two callbacks a try and it seems to work fine.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,111 @@ static void
...
   EkncContentObjectModel *model = eknc_domain_get_object_finish (domain, result, &error);
-  if (error != NULL)
-    {
-      g_task_return_error (task, error);
-      return;
-    }
 
-  state->models = g_slist_append (state->models, model);
+  *(&g_ptr_array_index (object_response_data->models, object_response_data->index)) = model;
+  scatter_task_countdown_subtask_completed (&object_response_data->countdown, error);
```
cosimoc: I think the error ownership around this function could be improved; `scatter_task_countdown_subtask_completed()` should make a copy of the error when needed, and this function should declare the error with `g_autoptr()`

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,111 @@ static void
...
+  /* If we received an error we should still continue to countdown, but
+   * store the error so that the caller can deal with it */
+  if (!countdown->error)
+    countdown->error = error;
+  else
+    g_clear_error (&error);
+
+  if (--countdown->remaining == 0)
+    {
+      (*countdown->callback) (error, countdown->data);
```
cosimoc: `countdown->error` should be passed here, not `error`.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,111 @@ static void
...
+static void
+scatter_task_countdown_subtask_completed (ScatterTaskCountdown **countdown_ptr,
+                                          GError *error)
+{
+  ScatterTaskCountdown *countdown = *countdown_ptr;
+
+  /* If we received an error we should still continue to countdown, but
+   * store the error so that the caller can deal with it */
+  if (!countdown->error)
+    countdown->error = error;
```
cosimoc: The error should be copied here -- see comment below

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,111 @@ static void
...
+  else
+    g_clear_error (&error);
```
cosimoc: You won't need to do this here if you follow my suggestion about error ownership

ekncontent/ekncontent/eknc-domain.c
smspillaz: Yep, that's right. Thanks for catching

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,111 @@ static void
...
+                                          GError *in_error)
+{
+  ScatterTaskCountdown *countdown = *countdown_ptr;
+  g_autoptr(GError) error = in_error;
+
+
+  /* If we received an error we should still continue to countdown, but
+   * store the error so that the caller can deal with it */
+  if (!countdown->error && error)
+    countdown->error = g_error_copy(error);
```
cosimoc: Missing space before paren

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -739,40 +739,111 @@ static void
...
+
+  return countdown;
+}
+
+static void
+scatter_task_countdown_subtask_completed (ScatterTaskCountdown **countdown_ptr,
+                                          GError *in_error)
+{
+  ScatterTaskCountdown *countdown = *countdown_ptr;
+  g_autoptr(GError) error = in_error;
```
cosimoc: This looks a bit too complicated. The caller of this function owns `in_error` and it should be freed from there instead, since this function will make a copy when appropriate.

smspillaz:
@cosimoc Amended the last commit with your suggestions

PR #1421, opened 2017-05-13
Order/Filter refactor to modify Xapian queries

ptomato:
Large pull request, but it can be divided approximately into three parts. I would suggest reviewing it commit by commit, that will make the most sense.

- first 6 commits: bugfixes / refactors that I ran into; would be good to commit these to master as soon as they are approved.
- next 3 commits: the new functionality
- last 7 commits: refactors that attempt to make the best use of the new functionality

https://phabricator.endlessm.com/T13175


ptomato:
...and of course Github messes up the order of the commits in this view. My division into three parts assumes the order in which they actually appear on the branch.

js/app/utils.js
```diff
@@ -540,3 +540,25 @@ function ensure_directory (dir) {
...
+
+// Set-like operations for arrays: union, intersection
+
+function union(a, b) {
+    if (!a)
+        return b;
+    if (!b)
+        return a;
+    let union = new Set(a);
+    b.forEach(item => union.add(item));
```
tchx84: nit: why not
```
let a = new Set([1,2,3]);
let b = new Set([4,3,2]);
let union = new Set([...a, ...b]);
```

?

data/preset/blog.yaml
```diff
@@ -89,9 +89,12 @@ defines:
...
             card: *main-card
         selection:
-          type: Selection.RelatedArticles
+          type: Selection.Related
           slots:
-            filter: 'Filter.Sets(invert: true)'
+            filter:
+              type: Filter.Articles
+              slots:
+                sub-filter: Filter.Unread
```
tchx84: This looks nice!

js/app/modules/selection/featuredFirst.js
```diff
@@ -21,13 +30,11 @@ const AllHighlights = new Module.Class({
         if (query_index == 0) {
             return Eknc.QueryObject.new_from_props({
                 limit: limit,
-                tags_match_all: ['EknArticleObject', 'EknFeaturedTag'],
-                excluded_tags: [],
+                tags_match_all: ['EknFeaturedTag'],
```
tchx84: Isn't this querying for only Featured content? So is not really "first", but "only".

data/preset/blog.yaml
```diff
@@ -39,7 +39,13 @@ defines:
...
                 card:
                   type: Card.Title
-            selection: Selection.StaticArticles
+            selection:
+              type: Selection.All
+              slots:
+                filter:
+                  shortdef: 'Filter.Tagged(tag: EknStaticTag)'
+                  slots:
+                    sub-filter: Filter.Articles
```
tchx84: Also nice!

data/preset/buffet.yaml
```diff
@@ -118,7 +121,9 @@ defines:
                             - 0
                             - 1
                     selection:
-                      type: Selection.AllHighlights
+                      type: Selection.FeaturedFirst
```
tchx84: when testing travel app, I noticed something weird, when you open the app.. you see the content group containing this selection loading cards one by one quite slowly.

tchx84:
@ptomato like we mentioned during the standup, moving these 6 fixes to another PR would be great, so we can merge them faster.

js/app/modules/selection/featuredFirst.js
ptomato: No, because when this query returns nothing, the `query_index` is incremented, and the second query excluding `EknFeaturedTag` kicks in.

data/preset/buffet.yaml
ptomato: Yes, I see that too. It seems like it's querying 4 models, then 3, then 2, then 1 model, but only adding one each time. And it's not happening on master.

ptomato:
I haven't had time to track down the change in buffet behaviour, but otherwise this is ready for review again.

js/app/modules/filter/currentSet.js
```diff
@@ -48,4 +50,26 @@ const CurrentSet = new Module.Class({
 
         return belongs_to_set && !belongs_to_subset;
     },
+
+    // Filter override
+    // This Filter can modify the Xapian query, but it can't express its whole
+    // filtering logic in the Xapian query, so we still need to request more
+    // models in case some returned models must be dropped.
+    can_modify_xapian_query: function () {
```
tchx84: @ptomato couldn't this be implemented by just using the SetMap data to..

tags_match_all: [current_set_tag]
AND
excluded_tags:[list_of_subsets_of_current_set].

?

tchx84:
@ptomato finished my second round, I think is in good shape. Re. the last comment I did for Filter.CurrentSet and the previous regression seen in travel, do you think we should merge it as is and open other issues to solves those? or we fix them before?

ptomato:
Up to you, as the reviewer. I likely won't get around to improving it before we branch for eos3.2 so it may be advantageous to merge it now.

tchx84:
@ptomato yeah, makes sense, can you create these issue? Ill merge it now.

PR #1424, opened 2017-05-15
T17076

tchx84:
https://phabricator.endlessm.com/T17076

### DIFF
--- a/data/css/blog.scss
+++ b/data/css/blog.scss
@@ -77,6 +77,15 @@ $highlighted-background-color: darken($background-dark-color, 5%) !default;
   }
 }
 
+.ContentGroup--complementarycontent {
+  .ContentGroupContentGroup__title {
+    font-size: 46px;
+    font-weight: 900;
+    color: $primary-medium-color;
+    margin: 30px 0px 20px 0px;
+  }
+}
+
 .home-page {
   .main-arrangement {
     margin-top: 40px;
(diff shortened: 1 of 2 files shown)


data/css/blog.scss
```diff
@@ -122,8 +122,9 @@ $highlighted-background-color: darken($background-dark-color, 5%) !default;
 
 .article-page {
   .related-posts {
-    margin: 0px 50px 20px 50px;
-
+    background-color: $background-light-color;
+    margin-right: 15px;
+    padding: 0px 50px 20px 50px;
```
ptomato: Why the change from `margin` to `padding`? It would be good to give a rationale for a non-obvious change like that in the commit message.

tchx84: @ptomato let me try to sketch it here, hehe:

1) this change is about adding a background color to the Related Articles content goup.
2) previously, we had that margin applied to content group, so when adding a background color, the background would only extend to immediate boundaries of the content goup (looking like a floating box on top of the article background color).
3) so, to fix that I switch to padding so we would have the spacing around the content but extending its background on top of the articles background.
4) lastly, the final margin-right is simply to prevent the content group to hide the scrolled window slider.

does this make sense?

ptomato: OK, got it, could you please add it to the commit message then?

ptomato:
Per @katelk28's comment on https://phabricator.endlessm.com/T17076, it might be good to add a style class to the default theme for content groups like this one that pop up over the webview.

xjuan:
margin -> padding is because it needs to cover the article with the background color

CR pass

PR #1426, opened 2017-05-17
T16238

katelk28:
An attempt to more fully implement our default theme across the modular FW modules and templates, and detangle a bit of CSS.
- Modules: implement correct default theming, while untangling various module CSS and implementing various UI changes as well
- Template overrides: Refactored template CSS to reflect the above changes in the default theme
- App overrides: I'll be putting a PR in to eos-subscriptions-apps soon as well, which will have changes to various app overrides that have been using some colors and module overrides incorrectly, so these changes affect those as well

This is a huge PR with a few huge commits, but hopefully this will ultimately be really helpful for continued use of the framework!

https://phabricator.endlessm.com/T16238


data/css/modules/pager/_parallaxBackground.scss
```diff
@@ -33,4 +33,10 @@
     &--right {
         background-image: url($secondary-background-image);
     }
+
+    &--right {
+        .LayoutSidebar .sidebar {
```
katelk28: You're right, it probably doesn't. I moved it to the template level CSS.

js/app/widgets/lightbox.js
```diff
@@ -275,14 +274,14 @@ const LightboxContainer = new Knowledge.Class({
         this._previous_button = new Gtk.Button({
             halign: Gtk.Align.START,
             valign: Gtk.Align.START,
-            margin: this._ICON_MARGIN,
+            margin: 20,
```
katelk28: I purposefully moved it out of _ICON_MARGIN because they're different values; `this._previous_button` needs margins all around, and `this._close_button` needs different values for `margin-top` and `margin-left`.

Also, yeah, did try to move it to CSS, but something about GTK doesn't seem to like actual CSS margins as much as these explicit margins in the Gtk.Button definition; it would make the next & previous buttons go past the bottom of the screen. Not sure why, so decided it wasn't important enough to spend a ton of time on in this commit.

js/app/widgets/lightbox.js
```diff
@@ -275,14 +274,14 @@ const LightboxContainer = new Knowledge.Class({
...
             image: img_prev
         });
         Utils.set_hand_cursor_on_widget(this._previous_button);
         this._previous_button.connect('clicked', function () {
             this.emit('navigation-previous-clicked');
         }.bind(this));
-        this._previous_button.get_style_context().add_class(nav_button_class);
+        this._previous_button.get_style_context().add_class(Utils.get_element_style_class(Lightbox, 'navigationButtonPrevious'));
```
katelk28: @ptomato what's the 'same comment' you're talking about? 
and good thought. good thought. changed it to `navigationPrevious` and `navigationNext`.

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -92,6 +93,7 @@ const ContentGroup = new Module.Class({
 
                 // Style class needs to go on the button itself, so that we get :hover states and the like.
                 title_button.get_style_context().add_class(Utils.get_element_style_class(ContentGroup, 'title'));
+                title_button.get_style_context().add_class(Utils.get_element_style_class(ContentGroup, 'title_clickableStyle'));
```
katelk28: yeah I was tihnking about this. you're right. I'll separate them into 2 different styles. `ContentGroup__title` and `ContentGroup__title--clickable`.

js/app/modules/card/media.js
```diff
@@ -58,12 +58,12 @@ const Media = new Module.Class({
...
         this.set_label_or_hide(this._attribution_label, attributions.map((s) => {
             return s.split('\n')[0];
-        }).join(' - ').toUpperCase());
+        }).join(' - '));
         this._attribution_button.visible = this._attribution_label.visible;
         this._attribution_button.connect('clicked', this._show_license_in_external_viewer.bind(this));
-
-        this._separator.visible = this._attribution_label.visible &&
-                                  this._caption_label.visible;
+        if (this._caption_label.visible && this._attribution_label.visible) {
```
katelk28: @ptomato  I agree with your comment about BEM naming, but this solution seems odd to me. to my knowledge, in CSS, you can't give 2 different classes certain styling depending on if their _container contains_ those 2 classes. I think, rather, you can give 2 classes certain styling depending on if its container acutally has those 2 classes applied to _it_... I think that's what you're suggesting anyway, to add `--caption` and `--attribution` to the parent module, but that's kind of doing the same thing that I suggested, right? But just adding 2 new classes to the container rather than one (which i named `CaptionAndAttrVisible` or something. 

I agree that should be done in CSS, but I didn't think it was possible.

data/css/_colors.scss
```diff
@@ -7,10 +7,3 @@ $accent-dark-color: #cc532b !default;
 
 $background-light-color: #e9e9e3 !default;
 $background-dark-color: #ddddd5 !default;
-
-$white-60-color: transparentize(white, 0.4) !default;
```
katelk28: @ptomato I don't know, I see your point but still think we should get rid of them; I don't think something repeated enough times necessarily means a variable should be made out of it. The fact that these variables represent static values pretty much defeats the purpose of it being a variable at all. Also, 2 things -- in our CSS, we flip all the time between using the variables or using the actual values, and then when we don't use the variables, sometimes we might want another transparency value. I don't think we should have variables for these.

But I do agree that maybe having a `$overlay` color system might be interesting, if we use it enough. But right now, we've only been using these colors to define different transparencies of text colors, and sometimes background colors; we don't use overlays that much. But that seems like a better use case for variables if we did want that in the future.

js/app/modules/card/media.js
ptomato: I'm not sure I understand what you mean by the container of a class. The selector `.bemclass.bemclass--modifier1.bemclass--modifier2` will select an element with both modifiers applied to it at once. It is indeed similar to what you were suggesting, the differences are splitting it into 2 modifiers and using the `--` BEM convention for modifiers.

data/css/_colors.scss
ptomato: OK. like I said, I made the same case to @tchx84 as you are making now, so check with him to see if there's nothing that I'm missing.

js/app/modules/card/media.js
katelk28: Okay, I see what you mean; the only point of my comment was that it would be adding 2 classes to the module instead of 1, which seemed odd, but yeah, I agree that when looking at CSS It would look more intuitive. I'll change it!

data/css/modules/card/_media.scss
```diff
@@ -2,23 +2,40 @@
...
+    &__caption,
+    &__attributionText {
+        padding: 25px;
+    }
+}
+
+// When both the caption & attribution are present in the lightbox,
+// edit padding to accommodate both
+.CardMedia--caption.CardMedia--attribution {
+    .CardMedia {
```
ptomato: This doesn't seem correct; this would select a `Card.Media` module _inside_ another `Card.Media` module with caption and attribution. It looks like it should be `.CardMedia.CardMedia--caption.CardMedia--attribution` or else nested the other way around.

data/css/modules/contentGroup/_contentGroup.scss
```diff
@@ -1,68 +1,54 @@
...
 
-        &:hover {
-            color: $accent-dark-color;
-
-            .ThemeableImage {
-                -gtk-icon-source: $arrow-recolored-icon;
-            }
-        }
+    &__title,
+    &__title--clickable,
```
ptomato: If the BEM naming is correct, this isn't necessary; everything that has `.ContentGroupContentGroup__title--clickable` will also already have `.ContentGroupContentGroup__title`.

js/app/modules/card/media.js
ptomato: The logic is not quite right here - this now assigns `--caption` and `--attribution` if and only if both are present. In order to get the benefit of having 2 classes, it should assign `--caption` if a caption is present, and `--attribution` if an attribution is present

data/css/modules/card/_media.scss
katelk28: So what's happening (and what we discussed earlier on this thread, tell me if I interpreted your suggestion wrongly) is this:
- The `Card.Media` module has a class called `.CardMedia`, and when both attribution text & caption text are present, that module is given 2 new classes called `.CardMedia--attribution` and `.CardMedia--caption` in the JS file.
- This CSS is saying that whenever the outer `Card.Media` module has both of these classes applied to it, it should style its inner widgets (attribution text, with class `.CardMedia__attributionText`, and caption text, with class `.CardMedia__caption`) with different paddings than they usually have.

This works because the attribution text & the caption text are inside of the `Card.Media` module.

data/css/modules/contentGroup/_contentGroup.scss
katelk28: So I initlally had it apply both `.ContentGroupContentGroup__title` and `.ContentGroupContentGroup__title--clickable`, but an earlier comment from you said it was repetitive, so now the only class that's applied is `ContentGropuContentGroup__title--clicakble`.

data/css/modules/card/_media.scss
ptomato: You described exactly how I thought it _should_ work, and you are also right that it _does_ work that way, here's the CSS that it compiles down to:
```css
.CardMedia--caption.CardMedia--attribution .CardMedia__caption {
  padding-bottom: 7px; }

.CardMedia--caption.CardMedia--attribution .CardMedia__attributionText {
  padding-top: 7px; }
```
I didn't understand why until I considered how I would have written it. I got confused by the extra level of SCSS nesting looking like it should correspond to an extra level of nesting in the widget tree. Anyway, go ahead with this.

data/css/modules/contentGroup/_contentGroup.scss
ptomato: Nothing worse than a code review where you get conflicting comments in consecutive reviews! :man_facepalming: 
I looked back at the other comment to see what I meant. My intention with the comment was to get rid of `__title--clickable` and keep only `__title`, but I missed the fact that in the surrounding code there is also a case where the title is _not_ clickable and only `__title` is applied. Please disregard the earlier comment, sorry about that.

js/app/modules/card/media.js
```diff
@@ -58,12 +59,15 @@ const Media = new Module.Class({
...
-        }).join(' - ').toUpperCase());
+        }).join(' - '));
         this._attribution_button.visible = this._attribution_label.visible;
         this._attribution_button.connect('clicked', this._show_license_in_external_viewer.bind(this));
-
-        this._separator.visible = this._attribution_label.visible &&
-                                  this._caption_label.visible;
+        /* When both the caption & attribution are present in the lightbox,
+           add these classes so their spacing can be designed correctly. */
+        if (this._caption_label.visible && this._attribution_label.visible) {
```
ptomato: Posting again because my previous comment got eaten by an outdated diff:
The logic is not quite right here - this now assigns `--caption` and `--attribution` only if both are present. In order to get the benefit of having 2 classes, it should assign `--caption` if a caption is present, and `--attribution` if an attribution is present.

data/css/_colors.scss
katelk28: This should be fine, he siad he only needed them for the blog template earlier becuase that's how the default theme worked earlier

data/css/modules/card/_media.scss
katelk28: great!

data/css/modules/contentGroup/_contentGroup.scss
katelk28: hahaha no worries, glad to get it figured out. I agree with this solution.

js/app/modules/card/media.js
katelk28: overlooked this. good catch.

js/app/modules/contentGroup/contentGroup.js
```diff
@@ -91,7 +92,7 @@ const ContentGroup = new Module.Class({
                 });
 
                 // Style class needs to go on the button itself, so that we get :hover states and the like.
-                title_button.get_style_context().add_class(Utils.get_element_style_class(ContentGroup, 'title'));
+                title_button.get_style_context().add_class(Utils.get_element_style_class(ContentGroup, 'title--clickable'));
```
katelk28: fixed

ptomato:
Go ahead and merge if you agree with my commit.

PR #1430, opened 2017-05-24
presets: Switch to Banner.Dynamic in A, B, Buffet

ptomato:
These presets will use Banner.Dynamic instead of Banner.App going
forward, in order to keep them easily internationalizable.

https://phabricator.endlessm.com/T16725


xjuan:
CR pass

PR #1433, opened 2017-05-27
Kill & Replace Card.LegacyPost and Card.LegacyPolaroid with Card.DefaultFamily

katelk28:
Permanently deleting `Card.LegacyPolaroid` and `Card.LegacyPost`, and replacing instances of it with `Card.DefaultFamily` (and its surrouning `Arrangement` with `Arrangement.SquareGuys`, except for template A and library-list which use `Arrangement.Overflow`).

Also refactored bits of template files to reflect these changes.

https://phabricator.endlessm.com/T15933


data/preset/A.yaml
```diff
@@ -37,6 +35,10 @@ defines:
           shortdef: 'Arrangement.Overflow(orientation: horizontal)'
           slots:
             card: *card
+            properties:
```
xjuan: you need to add a type in order to add properties
```
card:
  type: *card
  properties:
```
Anyways I pushed a fix for so you can take a look at it and squash it with this commit

katelk28: wow, missed that in A, B, and library-list. not sure how I missed that. thanks!

xjuan:
CR pass

PR #1434, opened 2017-05-29
Add a ContentObjectModel for WordOfTheDay

erikos:
https://phabricator.endlessm.com/T17276

### DIFF
--- a/ekncontent/Makefile.am
+++ b/ekncontent/Makefile.am
@@ -74,6 +74,7 @@ ekncontent_private_installed_headers = \
 	ekncontent/eknc-set-object-model.h \
 	ekncontent/eknc-utils.h \
 	ekncontent/eknc-video-object-model.h \
+	ekncontent/eknc-word-object-model.h \
 	ekncontent/eknc-xapian-bridge.h \
 	$(NULL)
 ekncontent_sources = \
@@ -94,6 +95,7 @@ ekncontent_sources = \
 	ekncontent/eknc-utils-private.h \
 	ekncontent/eknc-utils.c \
 	ekncontent/eknc-video-object-model.c \
+	ekncontent/eknc-word-object-model.c \
 	ekncontent/eknc-xapian-bridge.c \
 	$(NULL)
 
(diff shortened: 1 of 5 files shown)


smspillaz:
Looks good to me generally. We probably want one for Quotes too. Is this currently being constructed anywhere?

erikos:
Thanks Sam for taking a look, yes, I will do Quotes next, just wanted to see if this is where we are heading.

ptomato:
I thought eos-discovery-feed wasn't intended to depend on this library - wouldn't you have to put this somewhere else?

erikos:
Added a ContentObjectModel for QuoteOfTheDay. 

We only use the Evergreen specific EkncContentObjectModel in the eos-knowledge-service to retrieve the content from the model, and that data is then forwarded to eos-discovery-feed. No dependency will be introduced for the e-d-f onto the knowledge-lib.

ptomato:
I see. Would it be possible to keep this in eos-knowledge-services among the other evergreen-related code? Putting it here exposes it as API in our developer platform, and I don't think I agree with that.

smspillaz:
@ptomato I think I'd agree with the sentiment here. The implementation is what potentially becomes tricky. The main problem is that we need to expose new fields specific to the WoTD/QoTD cards ("definition", "pronunciation", "type" etc) at the shard level. Those new fields have to propagate up the stack somehow.

We could retain the approach that we were using before and stuff them into `discoveryFeedContent`. I remember at least just before this we were field-stuffing into somewhat irrelevant fields like `synopsis`. I can't quite remember why that was, maybe @erikos can add more detail.

erikos:
Follow up for eos-knowledge-services: https://github.com/endlessm/eos-knowledge-services/pull/14

Follow up for eos-discovery-feed: https://github.com/endlessm/eos-discovery-feed/pull/23/commits/0f3087d5cae95f27d265d7d97ead8942e3052dc1

erikos:
Closing as we decided to go with another approach.

PR #1436, opened 2017-05-30
Cards: remove bottom right arrow

xjuan:
Remove bottom right arrow from deck, defaultFamily and list cards

https://phabricator.endlessm.com/T17305


ptomato:
You can remove all the `.arrow` CSS from the theme(s) as well!

PR #1440, opened 2017-06-01
T17212

tchx84:
https://phabricator.endlessm.com/T17212

### DIFF
--- a/js/app/application.js
+++ b/js/app/application.js
@@ -20,6 +20,7 @@ const Knowledge = imports.app.knowledge;
 const ModuleFactory = imports.app.moduleFactory;
 const MoltresEngine = imports.app.moltresEngine;
 const PromiseWrapper = imports.app.promiseWrapper;
+const SetMap = imports.app.setMap;
 
 let _ = Gettext.dgettext.bind(null, Config.GETTEXT_PACKAGE);
 
@@ -215,6 +216,7 @@ const Application = new Knowledge.Class({
     vfunc_startup: function () {
         this.parent();
         this._check_for_content();
+        this._initialize_set_map();
     },
 
     LoadItem: function (ekn_id, query, timestamp) {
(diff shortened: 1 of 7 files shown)


ptomato:
Looks like this needs a rebase, there are a lot of commits in here that were already merged

tchx84:
@ptomato done!

js/app/application.js
```diff
@@ -245,8 +246,23 @@ const Application = new Knowledge.Class({
         });
     },
 
+    initialize_set_map: function () {
```
ptomato: Make this private, it's not used outside of this class anymore.

js/app/application.js
```diff
@@ -245,8 +246,23 @@ const Application = new Knowledge.Class({
...
+            SetMap.init_map_with_models(results.models);
+        })
+        .catch(function (error) {
+            logError(error, 'Failed to load sets from database');
+        });
+    },
+
     _ensure_controller: function () {
         if (this._controller === null) {
+            this.initialize_set_map();
```
ptomato: I think this has to be done synchronously. Or else this whole call stack needs to be made asynchronous. Right now, there's a race between setting up the set map, and creating modules that want to use the set map.

tchx84: right, as it is, this is just trying to make the initialization earlier. I am a bit unsure how to address this properly..

a) re. making it sync, it looks all our API in ekncontent is async, so I am not sure how we can make that call sync without modifying the API.

b) re. making this whole block async, still not sure how that would work, since we expecting the controller creation tor be sync (e.g. by ensure_controller).

Any suggestions?

tchx84: Looks like our current Gjs version doesn't include await yet :( https://ptomato.wordpress.com/2017/04/28/gjs-whats-next/

tchx84: @ptomato we could the init call right after we setup the default domain, on _check_for_content, but, I don't think that guarantees anything (or at least not as I understand it).

ptomato:
OK, good enough for now.

PR #1446, opened 2017-06-06
Allow DEBUG_COMMAND to be specified directly

ojwb:
Don't explicitly set DEBUG_COMMAND to an empty value in the wrapper
scripts, to allow specifying tools other than gdb, e.g.
DEBUG_COMMAND=valgrind


ptomato:
Thanks!

PR #1447, opened 2017-06-06
Refactoring some News CSS & YAML

katelk28:
Some extra CSS removal from the News template that is not necessary on the template level.
- Deleted CSS having to do with text-transform, color, and some margin on content groups that
  should inherit from other content group styling
- Moved .ContentGroup--outercontainer upward in code for readability
- Combined CSS for .ContentGroup--complementarycontent and .LayoutResponsiveMargins--complementarycontent
- Changed trigger text to "More news from ...."

And, some default theme design QA on the Layout.TopMenu and its new dropdown menu for news template

https://phabricator.endlessm.com/T17424

### DIFF
--- a/data/preset/news.yaml
+++ b/data/preset/news.yaml
@@ -42,7 +42,7 @@ defines:
     contents:
     - type: ContentGroup.DynamicTitle
       properties:
-        format-string: !translate "see more from %s"
+        format-string: !translate "See more from %s"
         ellipsize: end
         hexpand: true
     - shortdef: 'Decoration.ThemeableImage(valign: end, halign: end)'
--- a/js/app/modules/arrangement/sideBySide.js
+++ b/js/app/modules/arrangement/sideBySide.js
@@ -12,7 +12,7 @@ const Knowledge = imports.app.knowledge;
 const Module = imports.app.interfaces.module;
 const Utils = imports.app.utils;
 
-const SPACING_MAX = 50;
+const SPACING_MAX = 0;
 
 /**
  * Class: SideBySide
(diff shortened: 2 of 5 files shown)


data/preset/news.yaml
```diff
@@ -42,7 +42,7 @@ defines:
     contents:
     - type: ContentGroup.DynamicTitle
       properties:
-        format-string: !translate "see more from %s"
+        format-string: !translate "More news from %s"
```
ptomato: Drive-by comment: normally we should avoid changing translatable strings after the feature freeze, since it invalidates translations that we may have already paid to have done. I don't know if that's the case now, but make sure to check before merging this.

katelk28: okay, good point. I won't worry about that for this release, I'll change it back, but I'd like to at least capitalize the "S" in "see more"

ptomato: That will still invalidate the translation, but we can work around that by manually re-entering the translations with the appropriate capitalization in Transifex.

js/app/modules/arrangement/sideBySide.js
```diff
@@ -12,7 +12,7 @@ const Knowledge = imports.app.knowledge;
 const Module = imports.app.interfaces.module;
 const Utils = imports.app.utils;
 
-const SPACING_MAX = 50;
+const SPACING_MAX = 0;
```
xjuan: This disables the responsive spacing between categories, did the design changed since I implemented the changes in the topbar?

katelk28: Yes, it's supposed to disable the spacing between the categories; now, the `CardTitle` modules are each given their own spacing. We don't need that responsive spacing anymore because of the new work you did on the dropdown menu; it's more important for the highlighted state of `CardTitle` to look more seamless

xjuan: OK, I am asking because a good chunk of the work I did on the topbar was to have that responsive spacing between cards. I wish I knew beforehand :D
So, I think the ideal thing would be to replace that constant with a property with a default of 0, let me know if you want me to do it

js/app/modules/arrangement/sideBySide.js
```diff
@@ -26,6 +24,16 @@ const SideBySide = new Module.Class({
     Extends: Endless.CustomContainer,
     Implements: [Arrangement.Arrangement],
 
+    Properties: {
+        /**
+         * Property: max-spacing
+         * The maximun spacing between cards
+         */
+        'max-spacing': GObject.ParamSpec.uint('max-spacing', 'Max Spacing', 'Maximun Spacing',
```
ptomato: typo: Maximun -> Maximum

xjuan:
Kate's commits CR pass

PR #1450, opened 2017-06-06
Probe for features supported by xapian-bridge in use

ojwb:
The results of the probe aren't yet used.

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -949,7 +949,9 @@ on_xapian_test_response (GObject *source,
   EkncDomain *domain = g_task_get_source_object (task);
   GCancellable *cancellable = g_task_get_cancellable (task);
 
-  eknc_xapian_bridge_test_finish (bridge, result);
+  g_autoptr(GError) error = NULL;
+  eknc_xapian_bridge_test_finish (bridge, result, &error);
+  /* Ignore failures - older xapian-bridge doesn't support /test. */
 
   EkncQueryObject *query = g_task_get_task_data (task);
   send_query_to_xapian_bridge (domain, bridge, query, cancellable,
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -924,6 +924,28 @@ on_xapian_query_response (GObject *source,
     }
 }
 
+void
+on_xapian_test_response (GObject *source,
+                         GAsyncResult *result,
+                         gpointer user_data)
+{
+  EkncXapianBridge *bridge = EKNC_XAPIAN_BRIDGE (source);
+  g_autoptr(GTask) task = user_data;
+  EkncDomain *domain = g_task_get_source_object (task);
+  GCancellable *cancellable = g_task_get_cancellable (task);
+
+  eknc_xapian_bridge_test_finish (bridge, result);
+
+  EkncQueryObject *query = g_task_get_task_data (task);
+
+  QueryState *state = g_slice_new0 (QueryState);
+  g_task_set_task_data (task, state, query_state_free);
+
+  g_autoptr(GHashTable) params = eknc_get_domain_query_params (domain);
+  eknc_xapian_bridge_query (bridge, query, params, cancellable,
+                            on_xapian_query_response, g_steal_pointer (&task));
+}
+
 /**
  * eknc_domain_query:
  * @self: the domain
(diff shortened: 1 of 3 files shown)


ekncontent/ekncontent/eknc-xapian-bridge.c
```diff
@@ -330,6 +350,72 @@ send_json_ld_request (EkncXapianBridge *self,
...
+
+/**
+ * eknc_xapian_bridge_test_finish:
+ * @self: the xapian bridge object
+ * @result: the #GAsyncResult that was provided to the callback.
+ *
+ * Finish a eknc_xapian_bridge_test call.
+ */
+void
+eknc_xapian_bridge_test_finish (EkncXapianBridge *self,
```
ptomato: The usual pattern for a `_finish` function in GIO-style async programming would be to include a `GError **error` parameter at the end, and have a `gboolean` return value. If the operation failed, it would return `FALSE` and set `*error`. (even though that's not likely yet, it would be good to keep the API consistent)

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -924,6 +924,28 @@ on_xapian_query_response (GObject *source,
...
+  EkncXapianBridge *bridge = EKNC_XAPIAN_BRIDGE (source);
+  g_autoptr(GTask) task = user_data;
+  EkncDomain *domain = g_task_get_source_object (task);
+  GCancellable *cancellable = g_task_get_cancellable (task);
+
+  eknc_xapian_bridge_test_finish (bridge, result);
+
+  EkncQueryObject *query = g_task_get_task_data (task);
+
+  QueryState *state = g_slice_new0 (QueryState);
```
cosimoc: I would want to refactor this code into a function that gets called by both here and the code path when the feature test is not needed below.

ekncontent/ekncontent/eknc-xapian-bridge.c
ojwb: @cosimoc explicitly said that failure should be handled as if this was an old-style xapian-bridge.

I can add the parameter and return value, but they'd be unused and always `TRUE` respectively...

cosimoc: I think it would be appropriate for this function to set an error and return `FALSE` when it can't talk to xapian-bridge, or there's an error during the communication. That of course would leave `has_default_op` set to `FALSE`, so it would be handled as an old-style xapian-bridge from the caller.

ptomato: I also think it would be OK even if they were never used - the pattern is more or less expected by bindings to other languages.

ojwb:
I pushed fixes yesterday for all points raised (in case github failed to notify you of that).

cosimoc:
Looks good to me, I will wait for @ptomato to give his approval before merge.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -924,6 +924,40 @@ on_xapian_query_response (GObject *source,
...
+                         gpointer user_data)
+{
+  EkncXapianBridge *bridge = EKNC_XAPIAN_BRIDGE (source);
+  g_autoptr(GTask) task = user_data;
+  EkncDomain *domain = g_task_get_source_object (task);
+  GCancellable *cancellable = g_task_get_cancellable (task);
+
+  g_autoptr(GError) error = NULL;
+  eknc_xapian_bridge_test_finish (bridge, result, &error);
+  /* Ignore failures - older xapian-bridge doesn't support /test. */
```
ptomato: This will leak `error` if there is one; use `g_clear_error(&error)`.

In addition, you might want to log a message if the error is something other than the expected case where xapian-bridge is old. However, that's not a blocker for this pull request, especially since any later query will just error out if that happens.

ojwb: Will change, but I really don't understand why the `g_autoptr` doesn't take care of cleaning up the object.

ojwb: Hmm - so https://developer.gnome.org/glib/stable/glib-Error-Reporting.html#g-clear-error says that `g_clear_error(&error);` boils down to `if (error) g_error_free(error)`.

But that's the same cleanup action defined for use by `g_autoptr()`:

```
/usr/include/glib-2.0/glib/glib-autocleanups.h:G_DEFINE_AUTOPTR_CLEANUP_FUNC(GError, g_error_free)
```

So why/how does this leak?

The codebase already widely uses `g_autoptr()`, so we're already assuming the use of a compiler which supports it.

ptomato: My mistake.

PR #1452, opened 2017-06-09
Query parsing changes

ojwb:
This makes use of new functionality in xapian-bridge (for which I've not yet opened a PR for as it changes code in the same places as another PR I've already submitted which is still awaiting review).  However, this should fall back to how things were before if used with an older xapian-bridge, so there's no inherent need to wait to merge this.


ekncontent/ekncontent/eknc-xapian-bridge.c
```diff
@@ -27,7 +27,7 @@ struct _EkncXapianBridge
   GObject parent_instance;
 
   gboolean feature_test_done;
-  gboolean has_default_op, has_flags, has_filter;
+  gboolean has_default_op, has_filter, has_flags;
```
ptomato: This order reversal looks like it was by mistake?

ekncontent/ekncontent/eknc-xapian-bridge.c
```diff
@@ -421,10 +447,10 @@ eknc_xapian_bridge_test_finish (EkncXapianBridge *self,
         {
           if (g_strcmp0 (s, "query-param-defaultOp") == 0)
             self->has_default_op = TRUE;
-          else if (g_strcmp0 (s, "query-param-flags") == 0)
-            self->has_flags = TRUE;
           else if (g_strcmp0 (s, "query-param-filter") == 0)
             self->has_filter = TRUE;
+          else if (g_strcmp0 (s, "query-param-flags") == 0)
```
ptomato: This move also looks like it might be by mistake

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -547,6 +549,19 @@ get_title_clause (EkncQueryObject *self, gchar **terms)
 }
 
 static gchar *
+get_title_clause2 (EkncQueryObject *self, gchar **terms)
+{
+  guint length = g_strv_length (terms);
+  g_auto(GStrv) title_terms = g_new (gchar *, length + 1);
```
ojwb: I just copied a pattern already widely used (in `get_exact_title_clause()`, `get_title_clause()`, `get_body_clause()`, `get_tags_clause()`, and `get_ids_clause()`).

I've pushed another commit which changes all of these to use the pattern you suggest.

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -607,6 +622,53 @@ get_query_clause (EkncQueryObject *self)
 }
 
 static gchar *
+get_query_clause2 (EkncQueryObject *self)
```
ojwb: The code duplication isn't ideal, but my thinking was that `get_query_clause()` is likely to be left alone now, while the code in `get_query_clause2()` is likely to evolve further, and so diverge more from `get_query_clause()`. Keeping the code paths for an old bridge and new bridge separate makes it easier to evolve the code for a new bridge without risk of breaking handling of an old bridge.

ekncontent/ekncontent/eknc-xapian-bridge.c
ojwb: Sorry, yes - badly handled rebase there.  I've force-pushed a fix.

ekncontent/ekncontent/eknc-xapian-bridge.c
ojwb: Indeed - also fixed.

ekncontent/ekncontent/eknc-query-object.c
ptomato: Thanks, indeed the existing pattern was wrong.

ekncontent/ekncontent/eknc-query-object.c
ptomato: OK, got it.

PR #1455, opened 2017-06-14
Make sure that va_end() is called on early exit

ojwb:
The man page says "Each invocation of va_start() must be matched by
a corresponding invocation of va_end() in the same function."


ptomato:
Good catch, thanks!

PR #1456, opened 2017-06-14
Add corrected-query property

ojwb:
The three-way `eknc_query_object_new_from_object()` call seems clumsy.  I wondered about just calling it without any extra properties (and tweaking `eknc_query_object_new_from_object()` so it can be called with just `(source, NULL)`), and then setting whichever properties apply.  However setting a property seems to require messing around with `GVariant` which seems like it's going to end up with even clumsier code.

Any better ideas?

### DIFF
--- a/ekncontent/ekncontent/eknc-query-object.c
+++ b/ekncontent/ekncontent/eknc-query-object.c
@@ -51,6 +51,7 @@ struct _EkncQueryObject
 
   gchar *app_id;
   gchar *query;
+  gchar *corrected_query;
   gchar *stopword_free_query;
   gchar *literal_query;
   gchar *query_parser_string;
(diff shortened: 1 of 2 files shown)


ptomato:
I don't mind the three invocations of `eknc_query_object_new_from_object()` — we set all the properties to be `CONSTRUCT_ONLY` so that query objects are immutable, and that's the price we pay. You could add a `eknc_query_object_new_from_object_and_params(EkncQueryObject *obj, unsigned n_params, GParameter **params)` if you really wanted to avoid it, but that seems more trouble than it's worth.

ojwb:
Ah, I hadn't considered the immutability.

If you're OK with the three invocations, l think that's probably best overall then.  If we end up with more than two parameters, we can reconsider.

PR #1457, opened 2017-06-14
T17434

tchx84:
https://phabricator.endlessm.com/T17434

### DIFF
--- a/lib/web-extensions/scrollplugin.c
+++ b/lib/web-extensions/scrollplugin.c
@@ -70,6 +70,9 @@ on_document_loaded (WebKitWebPage  *page,
                                               G_CALLBACK (signal_scroll),
                                               FALSE,
                                               data);
+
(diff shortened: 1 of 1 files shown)


lib/web-extensions/scrollplugin.c
```diff
@@ -70,6 +70,9 @@ on_document_loaded (WebKitWebPage  *page,
                                               G_CALLBACK (signal_scroll),
                                               FALSE,
                                               data);
+
+  // Always let the client know about the scroll status
+  signal_scroll (WEBKIT_DOM_EVENT_TARGET (document), NULL, data);
```
ptomato: The `WEBKIT_DOM_EVENT_TARGET` cast is not necessary, I think, since the function itself takes a `WebKitDOMDocument*`

tchx84:
@ptomato thanks, done!

PR #1463, opened 2017-06-15
News CSS: set font weight on dropdown menu cards

xjuan:
https://phabricator.endlessm.com/T17559


katelk28:
Good catch, lgtm. CR Pass

PR #1464, opened 2017-06-19
T17013

tchx84:
https://phabricator.endlessm.com/T17013

### DIFF
--- a/js/app/downloader.js
+++ b/js/app/downloader.js
@@ -285,6 +285,10 @@ const SubscriptionDownloader = new Lang.Class({
                 this._app.release();
                 this._ongoing_downloads.delete(subscription_id);
                 return updated;
+            })
+            .catch((error) => {
+                this._app.release();
+                throw error;
             });
         });
     },


js/app/downloader.js
```diff
@@ -270,7 +270,8 @@ const SubscriptionDownloader = new Lang.Class({
             let manifest_uri = Soup.URI.new(uri);
 
             let request = new FileDownloadRequest(manifest_uri, new_manifest_file);
-            return this._file_downloader.download_file_promise(request, cancellable).catch(function (error) {
+            return this._file_downloader.download_file_promise(request, cancellable).catch((error) => {
+                this._app.release();
```
ptomato: It seems like you should make a separate `.catch` clause and tack it on to the very end. If I understand the code correctly, then it seems like this will not release the app if an error occurs in the following `.then` clause.

Maybe something like this for the final clause:
```js
.then(updated => {
    this._app.release();
    this._ongoing_downloads.delete(subscription_id);
    return updated;
}, error => {
    logError(error);
    this._app.release();
});
```

PR #1466, opened 2017-06-20
Add TravisCI script

ebassi:
We want to use Travis to perform CI on the Endless apps SDK modules; in
order to achieve that, we use pre-built Docker images with the required
build dependencies.


ptomato:
We have CI on our internal Jenkins instance now. Eventually it would be good to move it to a publicly available service like TravisCI, but that should be a port of our existing jobs.

PR #1473, opened 2017-06-23
navigation: Only show the back button if we have > 1 page on stack

smspillaz:
Previously we were only hiding it when we were on the homepage,
though that assumed that the homepage was the first page on the
stack. If we open an article from a search provider, then we're
going to end up with that being the first page. Also, the user
should be able to press 'back' if they went to the homepage by
clicking on the home button.

https://phabricator.endlessm.com/T17573

### DIFF
--- a/js/app/modules/layout/navigation.js
+++ b/js/app/modules/layout/navigation.js
@@ -40,7 +40,7 @@ const Navigation = new Module.Class({
     },
 
     _on_history_changed: function () {
-        let item = HistoryStore.get_default().get_current_item();
-        this.back_visible = (item.page_type != Pages.HOME);
+        // The back button should only be visible if we can go back
+        this.back_visible = HistoryStore.get_default().can_go_back();
     },
 });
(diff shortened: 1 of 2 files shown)


erikos:
That looks good, though I leave to Philip to decide.

ptomato:
We discussed on Slack that this is touching the wrong code. That said, Sam said it fixed the problem, so I'm not sure what's going on there.

tchx84:
When I open an article in **master** I see:

virtualbox_endlessos_28_06_2017_14_13_59

When I open an article in this branch I see:
virtualbox_endlessos_28_06_2017_14_20_39

In both cases the top-bar buttons behave correctly. In master, the in-page back button is shown and takes you to the home page, while in this branch that button is disabled.

@ptomato said in slack:
> I think that behaviour that we already have is correct
> I asked Sam specifically yesterday evening in #dev which button he was talking about, he said the one in the top bar

I guess I move this back to Dev, but if there nothing to change lets just close it.

smspillaz:
Yeah, we've decided that the current behaviour is correct, so I'm closing this PR. Sorry for the confusion!

PR #1475, opened 2017-06-23
Picard: add card model type combo box

xjuan:
Add a way to choose which type of dummy ObjectModel to use when creating cards

https://phabricator.endlessm.com/T17891


tools/picard.js
```diff
@@ -377,7 +385,8 @@ function connect_signals () {
     SetMap.init_map_with_models(sets);
 
     widgets.add_box.connect('clicked', () => {
-        let model = Eknc.ArticleObjectModel.new_from_props({
+        let object_model = widgets.object_model_combo_box.get_active_text();
```
tchx84: @xjuan would it be too hard to update the current setup/view when another object model type is selected?

xjuan: I guess not, but I want to be able to have different object models at the same time

PR #1476, opened 2017-06-26
Card base CSS: use eos-play-symbolic icon

xjuan:
Add hover state for play icon

https://phabricator.endlessm.com/T17897


eos-knowledge.gresource.xml
```diff
@@ -12,11 +12,11 @@
     <file compressed="true">data/css/preset_a.css</file>
     <file compressed="true">data/css/preset_b.css</file>
     <file compressed="true">data/css/keybindings.css</file>
+    <file>data/icons/scalable/actions/eos-play-symbolic.svg</file>
```
ptomato: optional: use `xml-stripblanks` here

data/css/modules/card/_base.scss
```diff
@@ -38,14 +38,23 @@
       min-width: 96px;
       min-height: 96px;
       border-radius: 50%;
+      color: rgba(white, .6);
       background-color: rgba(#1e1e1e, .8);
```
ptomato: There is no spec for this, but Kate suggested using this very similar one: https://endlessm.invisionapp.com/share/9GC3BQEHQ#/screens/190153077_Video_Player_States_r3-1
That would mean that the background color should be adjusted slightly.

xjuan: I used the same color as in the video player, but now that I double check that design I see its says it should be black with .6 opacity

PR #1477, opened 2017-06-26
T17870

katelk28:
Video/audio overlay size responsiveness for Card.List & Card.default.

https://phabricator.endlessm.com/T17870

### DIFF
--- a/data/css/modules/card/_default.scss
+++ b/data/css/modules/card/_default.scss
@@ -209,6 +209,19 @@
                 padding: 20px;
             }
         }
+
+        // Video/audio icon sizing
+        &.height-b.width-c,
+        &.height-b.width-b,
+        &.height-b.width-a,
+        &.height-a {
+            .CardDefault__media_overlay.audio,
+            .CardDefault__media_overlay.video {
+                min-width: 60px;
+                min-height: 60px;
+                background-size: 18px 18px;
+            }
+        }
     }
 
     &.CardPost {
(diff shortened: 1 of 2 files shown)


data/css/modules/card/_default.scss
```diff
@@ -209,6 +209,19 @@
                 padding: 20px;
             }
         }
+
+        // Video/audio icon sizing
```
ptomato: I don't see this in the specs that you linked in the Phabricator issue; the icons are all large-sized on the polaroid samples shown there. Even so, shouldn't it be `&.height-b, &.height-a` without regard to width, as in the post card below?

data/css/modules/card/_list.scss
```diff
@@ -87,4 +87,16 @@
     .composite &.width-e &__synopsis {
         font-size: 18.8px;  /* 16px spec * 88% because spec was at 110% scale not 125% */
     }
+
+    // Video/audio icon sizing
+    &.width-f,
```
ptomato: I don't see this in the specs either, but even so shouldn't it be the small size all the way down to `.width-a`?

data/css/modules/card/_default.scss
xjuan: Here are the specs
https://endlessm.invisionapp.com/d/main#/console/8756792/238130473/preview

The only different size is in the smaller default postcard (Card Height B)

katelk28: The specs were a bit unclear as to what exact size the video overlays should take. I updated the specs with the specific sizes. 
For the Polaroid card, some of them do need widths, because not every card in that height wants the larger play button. For the Post card, yeah, it's only specified by height.

data/css/modules/card/_list.scss
katelk28: You are right. I didn't think it was necessary since the list card won't really ever go down past `&.width-d`, but that should be specified anyway.

xjuan:
CR pass

PR #1480, opened 2017-06-27
Creating gallery template and CSS

katelk28:
New template 'Gallery', and template CSS.
Added to Makefile, gresource, and Kactus

https://phabricator.endlessm.com/T17334

### DIFF
--- a/data/preset/gallery.yaml
+++ b/data/preset/gallery.yaml
@@ -1,0 +1,556 @@
+defines:
+
+# Recurring components
+- &main-card
+  type: Card.Default
+  properties:
+    excluded_types:
+      - 0
+      - 1
+
+- &main-arrangement
+  type: Arrangement.Constrained
+  slots:
+    card:
+      <<: *main-card
+
+- &main-menu
+  type: Layout.ResponsiveMargins
+  slots:
+    content:
+      type: Navigation.TopMenu
+      slots:
+        banner: 'Banner.Dynamic(mode: full, show-subtitle: false)'
+        menu:
+          shortdef: ContentGroup.ContentGroup
+          slots:
+            arrangement:
+              shortdef: 'Arrangement.SideBySide(expand: true)'
+              slots:
+                card:
+                  type: Card.Title
+                  properties:
+                    halign: fill
+                    vexpand: true
+            selection:
+              type: Selection.AllSets
+              slots:
+                filter: Filter.Featured
+
+- &selection-home
+  type: Selection.FeaturedFirst
+
+- &selection-set
+  type: Selection.ContentForSet
+
+# Pages
+
+# FIXME: The repeating ContentGroup structure should be replaced with the
+# infinite scrolling arrangement once it's made (T16360).
+- &home-page
+  type: Layout.TopMenu
+  slots:
+    top-menu:
+      <<: *main-menu
+    content:
+      shortdef: 'Layout.Scrolling(expand: true)'
+      slots:
+        content:
+          type: Layout.ResponsiveMargins
+          slots:
+            content:
+              shortdef: 'Layout.Box(orientation: vertical)'
+              slots:
+                contents:
+                  - type: ContentGroup.ContentGroup
+                    slots:
+                      arrangement:
+                        <<: *main-arrangement
+                      selection:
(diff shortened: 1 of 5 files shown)


data/preset/gallery.yaml
```diff
@@ -0,0 +1,555 @@
...
+        content:
+          type: Layout.ResponsiveMargins
+          slots:
+            content:
+              shortdef: 'Layout.Box(orientation: vertical)'
+              slots:
+                contents:
+                  - type: ContentGroup.ContentGroup
+                    slots:
+                      arrangement:
```
tchx84: bit: trailing whitespace

katelk28: @tchx84 fixed!

data/preset/gallery.yaml
```diff
@@ -0,0 +1,555 @@
...
+                                            slots:
+                                              sub-filter:
+                                                type: Filter.OtherGroup
+                                                references:
+                                                  other: CG5
+                                                slots:
+                                                  sub-filter:
+                                                    type: Filter.OtherGroup
+                                                    references:
+                                                      other: CG6
```
tchx84: note for the future: we really need the infinite scrolling version of Arrangement.Constrained

data/preset/gallery.yaml
```diff
@@ -0,0 +1,555 @@
...
+                                    type: Filter.OtherGroup
+                                    references:
+                                      other: CG2
+                  - type: ContentGroup.ContentGroup
+                    slots:
+                      arrangement:
+                        <<: *main-arrangement
+                      selection:
+                        <<: *selection-home
+                        id: CG4
```
tchx84: I think I spotted one issue, AFAIK, the module ids should be unique! So, if you id a module "CG1", every module (regardless of where they are in the app) that references that id would be getting the same one.. but here you are assigning the same id to multiple modules, and that could cause a lot of unexpected behaviors.

data/preset/gallery.yaml
```diff
@@ -0,0 +1,555 @@
...
+                    slots:
+                      card:
+                        type: Card.Title
+                  - type: ContentGroup.ContentGroup
+                    slots:
+                      arrangement:
+                        <<: *main-arrangement
+                      selection:
+                        <<: *selection-set
+                        id: CG1
```
tchx84: continuing with my previous comment, "id: CG1" is defined twice, but should probably be unique.

data/preset/gallery.yaml
katelk28: Very, very much agree. Left a `FIXME` comment in my fixes.

data/preset/gallery.yaml
```diff
@@ -0,0 +1,560 @@
...
+
+# FIXME: The repeating ContentGroup structure should be replaced with the
+# infinite scrolling arrangement once it's made (T16360).
+- &set-page
+  type: Layout.TopMenu
+  slots:
+    top-menu:
+      <<: *main-menu
+    content:
+      shortdef: 'Layout.InfiniteScrolling(expand: true)'
```
tchx84: If the amoutn of card we display is fixed (because Constrained uses fixed space plus we have a fixed number of ContentGriups) then we should use Layout.Scrolling. Any particular reason for not doing that?

katelk28: You're right. There's no reason, probably just was not as familiar with the real differences between `Layout.Scrolling` and `Layout.InfiniteScrolling` when I wrote this yaml file. Changing now.

I'll leave it be `InfiniteScrolling` for the search page, though, since that does use an infinite scrolling arrangement (doesn't use `Arrangement.Constrained`.

data/preset/gallery.yaml
```diff
@@ -0,0 +1,560 @@
...
+# infinite scrolling arrangement once it's made (T16360).
+- &home-page
+  type: Layout.TopMenu
+  slots:
+    top-menu:
+      <<: *main-menu
+    content:
+      shortdef: 'Layout.Scrolling(expand: true)'
+      references:
+        lazy-load: highlighted
```
tchx84: areferences: lazy-load, is not a valid slot for Scrolling (and is not needed)

PR #1481, opened 2017-06-27
application: Don't initialize icon theme too early

ptomato:
In the application's constructor, GTK has not yet been initialized.

https://phabricator.endlessm.com/T17897


smspillaz:
I was running into this too, and this fixes the problem.

tchx84:
**CR Pass**

PR #1483, opened 2017-07-03
application: Fix content download dialog

tchx84:
Domain.get_subscription_entries no longer exists, we
forgot to update this call to the new ekncontent API.

https://phabricator.endlessm.com/T18050


xjuan:
CR Pass

PR #1485, opened 2017-07-06
T18071

tchx84:
https://phabricator.endlessm.com/T18071


ekncontent/tests/ekncontent/testDomain.js
```diff
@@ -63,6 +63,13 @@ describe('Domain', function () {
             });
         });
 
+        it('throws for an ID that is not valid', function (done) {
+            domain.get_object("ekn:///02463d24cb5690af2c8e898736ea8c80e0e77077/", null, function (domain, result) {
```
ptomato: Better to make a different, "more invalid" example of an invalid ID, since we were talking about possibly accepting the trailing slash.

ekncontent/ekncontent/eknc-domain.h
```diff
@@ -34,6 +35,7 @@ typedef enum {
   EKNC_DOMAIN_ERROR_BAD_RESULTS,
   EKNC_DOMAIN_ERROR_UNSUPPORTED_VERSION,
   EKNC_DOMAIN_ERROR_PATH_NOT_FOUND,
+  EKNC_DOMAIN_ERROR_ID_NOT_VALID,
```
ptomato: This is technically an API addition, so backporting to SDK1 would be problematic. @ebassi what do you think?

PR #1486, opened 2017-07-06
T18066

tchx84:
https://phabricator.endlessm.com/T18066


xjuan:
CR pass
@tchx84 will file a new task to create a new filter to include sets in the results

PR #1488, opened 2017-07-12
T18071 (second approach)

tchx84:
https://phabricator.endlessm.com/T18071

### DIFF
--- a/js/app/modules/card/knowledgeDocument.js
+++ b/js/app/modules/card/knowledgeDocument.js
@@ -22,6 +22,7 @@ const Utils = imports.app.utils;
 
 const SPINNER_PAGE_NAME = 'spinner';
 const CONTENT_PAGE_NAME = 'content';
+const SCROLLED_PAST_PREFIX = 'scrolled-past-';
 
 /**
  * Class: KnowledgeDocument
@@ -224,6 +225,11 @@ const KnowledgeDocument = new Module.Class({
         this.content_view.run_javascript(script, null, null);
     },
 
+    _find_section_index: function (sectionName) {
+        return this._mainArticleSections.findIndex(section =>
+            section.hasContent.split('#')[1] === sectionName);
+    },
+
     // Keep separate function to mock out in tests
     _create_webview: function () {
         return new EknWebview.EknWebview({
@@ -237,7 +243,7 @@ const KnowledgeDocument = new Module.Class({
     _get_webview: function () {
         let webview = this._create_webview();
 
-        webview.renderer.enable_scroll_manager = this.show_toc;
+        webview.renderer.enable_scroll_manager = true;
         webview.renderer.show_title = !this.show_toc;
         if (this.custom_css)
             webview.renderer.set_custom_css_files([this.custom_css]);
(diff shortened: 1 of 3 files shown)


js/app/modules/card/knowledgeDocument.js
```diff
@@ -224,6 +225,15 @@ const KnowledgeDocument = new Module.Class({
         this.content_view.run_javascript(script, null, null);
     },
 
+    _find_section_index: function (sectionName) {
+        for (let index in this._mainArticleSections) {
```
ptomato: Optional: I suggest letting JS do the work here :smile:
```js
return this._mainArticleSections.findIndex(section =>
    section.hasContent.split('#')[1] === sectionName);
```

js/app/modules/card/knowledgeDocument.js
```diff
@@ -237,7 +247,7 @@ const KnowledgeDocument = new Module.Class({
     _get_webview: function () {
         let webview = this._create_webview();
 
-        webview.renderer.enable_scroll_manager = this.show_toc;
+        webview.renderer.enable_scroll_manager = true;
```
tchx84: This approach uses our JS [scroll manager](https://github.com/endlessm/eos-knowledge-lib/blob/master/data/templates/js/scroll-manager.js#L33) helper (same as when you click on the TOC). Before, made no sense to have it enabled if the TOC wasn't display, but now it does, because you want this to work even if the TOC is not displayed.

js/app/modules/card/knowledgeDocument.js
```diff
@@ -293,10 +297,18 @@ const KnowledgeDocument = new Module.Class({
 
             let [baseURI, hash] = decision.request.uri.split('#');
 
-            if (baseURI === this.model.ekn_id) {
+            if (baseURI.startsWith(this.model.ekn_id)) {
```
tchx84: Forgot to mention, but I didn't find a good way of doing this that doesn't require an external js lib.

js/app/modules/card/knowledgeDocument.js
ptomato: But the scroll manager isn't needed to do the scrolling, right? The `scroll_to()` method only does the fancy animation while updating the TOC. It _should_ be that if you navigate to `the.current/page#anchor` then WebKit will just move to that anchor as if you had navigated only to `#anchor`. In fact I'm wondering why that's not the case already in the code.

tchx84: hmm, at least in the way this is implemented it depends on `scrollTo` which is defined in the manager, which is disabled if the TOC is not visible, so it simply won't work.

I thought the actual scroll was handled [here](https://github.com/endlessm/eos-knowledge-lib/blob/master/data/templates/js/scroll-manager.js#L39) in the manager.

ptomato: Oh. Maybe the answer is to check in the decision handler if the scroll manager is enabled. If it is not, then return false and let webkit's default decision handler navigate to `#anchor`?

tchx84: You are right, but doesn't work out of the box.

1. we still need the `.startWith` fix.
2. even if we return false, the EknWebView custom handler for 'ekn:' scheme must consider this case, or else this case falls into a [case](https://github.com/endlessm/eos-knowledge-lib/blob/master/js/app/widgets/eknWebview.js#L154) which I honestly can't understand what it is for, any idea?

Anyway, I did a dirty/quick hack just to check it works (and it does), but I need to make sure I can insole from that case I mentioned inside that handler so I don't break anything.

The only doubt I have re. this approach (via webkit) is that clicking on that link will teleport the user to the section, differently from the animated way of doing when clicking on the TOC, so te user will get different behaviors depending on where it clicks. 

if we try to unify these two behaviors we basically end up with same PR we have now.

Let me know what you think!

ptomato: The second case is a "named blob" which was something Jasper invented: https://github.com/endlessm/eos-knowledge-lib/commit/12e0b3900a37eff58acb8e55cd48a298c69107ba So, I think your check for the component starting with `#` should work fine.

Personally I like your WIP branch better, I prefer not having the overhead of the jquery hack firing on every scroll event, if we don't need it for updating the TOC. But maybe we can ask Joana to weigh in?

ptomato: To summarize our discussions from today, I am convinced now that we should use the scroll manager. When you click on the link, it's not clear that you're going to remain in the same page, and so the scroll animation helps make it clear that you are elsewhere in the same document.

tchx84:
@ptomato can you double check is in the right branch? (this PR branch is T18071-try2), I am testing it as well, and works fine. It also has a unit tests, which works as well...

Can you give me more details?

ptomato:
I isolated the problem down to running with mozjs52, since I was testing this rebased on top of my T18106 branch. Since it works fine for the current configuration, I'll merge it.

PR #1496, opened 2017-07-27
Update README for new directory structure

mwleeds:


ptomato:
Thanks. I thought for sure I had done this already, but then I realized it had only been updated on the website (http://endlessm.github.io/eos-knowledge-lib/contributing), not in the README. Instead of this change, we should probably remove the duplicated information from the README so that these two don't diverge.

mwleeds:
Ah, okay. Should we also link to the website in the README or in the GitHub description for the repo?

ptomato:
Sure. If you want to add that to this PR, that would be welcome — in any case it's good to merge as is.

PR #1497, opened 2017-08-01
Add support for content access metrics

mwleeds:
This is working and ready for CR. See T18516 for details.

### DIFF
--- a/eos-knowledge.gresource.xml
+++ b/eos-knowledge.gresource.xml
@@ -76,6 +76,7 @@
     <file>js/app/courseHistoryStore.js</file>
     <file>js/app/dispatcher.js</file>
     <file>js/app/dominantColor.js</file>
+    <file>js/app/entryPoints.js</file>
     <file>js/app/historyItem.js</file>
     <file>js/app/historyStore.js</file>
     <file>js/app/interfaces/arrangement.js</file>
--- a/js/app/application.js
+++ b/js/app/application.js
@@ -15,11 +15,14 @@ const System = imports.system;
 const Actions = imports.app.actions;
 const Config = imports.app.config;
 const Dispatcher = imports.app.dispatcher;
+const HistoryStore = imports.app.historyStore;
 const Knowledge = imports.app.knowledge;
 const ModuleFactory = imports.app.moduleFactory;
 const MoltresEngine = imports.app.moltresEngine;
+const Pages = imports.app.pages;
 const PromiseWrapper = imports.app.promiseWrapper;
 const SetMap = imports.app.setMap;
+const Utils = imports.app.utils;
 
 let _ = Gettext.dgettext.bind(null, Config.GETTEXT_PACKAGE);
 
@@ -308,6 +311,12 @@ var Application = new Knowledge.Class({
     },
 
     vfunc_shutdown: function () {
+        // Record a content access metric before shutting down
+        let history = HistoryStore.get_default();
+        let last_item = history.get_current_item();
+        if (last_item && last_item.model && last_item.page_type === Pages.ARTICLE)
+            Utils.stop_content_access_metric(last_item.model);
+
         Dispatcher.get_default().pause();
         this.parent();
     },
--- a/js/app/historyStore.js
+++ b/js/app/historyStore.js
@@ -9,6 +9,7 @@ const Lang = imports.lang;
 
 const Actions = imports.app.actions;
 const Dispatcher = imports.app.dispatcher;
+const EntryPoints = imports.app.entryPoints;
 const HistoryItem = imports.app.historyItem;
 const Pages = imports.app.pages;
 const ReadingHistoryModel = imports.app.readingHistoryModel;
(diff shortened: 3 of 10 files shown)


js/app/historyStore.js
```diff
@@ -119,6 +122,29 @@ const HistoryStore = new Lang.Class({
...
+                                               old_item.model.title,
+                                               old_item.model.content_type);
+
+        if (item.model && item.model instanceof Eknc.ArticleObjectModel &&
+            (!old_item || !old_item.model ||
+             old_item.model.ekn_id !== item.model.ekn_id)) {
+            Utils.record_content_access_metric(true,
+                                               item.entry_point ? item.entry_point : '',
+                                               item.model.ekn_id,
+                                               item.model.title,
```
mwleeds: We don't need to check if title or content_type are defined right? I couldn't find anything during testing that didn't have them, but I don't know if that's guaranteed.

js/app/historyStore.js
```diff
@@ -119,6 +119,27 @@ const HistoryStore = new Lang.Class({
 
     // Default signal handler
     on_changed: function () {
+        let old_item = this._direction == Direction.FORWARDS ?
```
ptomato: nitpick: use `===`

js/app/historyStore.js
ptomato: In theory they are guaranteed, but content is messy, so I would prefer to be defensive here.

js/app/utils.js
```diff
@@ -451,6 +451,19 @@ function record_search_metric (query) {
...
+// Should be mocked out during tests so that we don't actually send metrics
+function record_content_access_metric (open, ekn_id, title, content_type) {
+    let app_id = Gio.Application.get_default().application_id;
+    let recorder = EosMetrics.EventRecorder.get_default();
+    if (open)
+        recorder.record_start(CONTENT_ACCESS_METRIC_EVENT_ID, new GLib.Variant('s', ekn_id),
+            new GLib.Variant('(sss)', [app_id, title, content_type]));
+    else
+        recorder.record_stop(CONTENT_ACCESS_METRIC_EVENT_ID, new GLib.Variant('s', ekn_id),
+            new GLib.Variant('(sss)', [app_id, title, content_type]));
```
ptomato: You don't actually have to send the payload along with the stop metric a second time, only the key. So the second variant can be empty.

js/app/application.js
```diff
@@ -352,6 +355,15 @@ const Application = new Knowledge.Class({
     },
 
     vfunc_shutdown: function () {
+        // Record a content access metric before shutting down
```
ptomato: It would be better not to introduce all this coupling into application.js, though I'm not sure if it's really possible. Is there any way we could keep this inside HistoryStore?

js/app/historyStore.js
```diff
@@ -91,9 +92,11 @@ const HistoryStore = new Lang.Class({
         Dispatcher.get_default().register((payload) => {
             switch(payload.action_type) {
                 case Actions.HISTORY_BACK_CLICKED:
+                    this.get_previous_item().entry_point = EntryPoints.NAV_BUTTON_CLICKED;
```
ptomato: Hmm. I realized this changes things so that history items are no longer immutable. Is there a way to avoid that? I think there should be, since you don't need to know the entry point at the time you record the stop metric.

eos-knowledge.gresource.xml
```diff
@@ -76,6 +76,7 @@
     <file>js/app/dispatcher.js</file>
     <file>js/app/dominantColor.js</file>
     <file>js/app/downloader.js</file>
+    <file>js/app/entryPoints.js</file>
```
ptomato: Seems like the entryPoints.js file itself should be part of this commit, rather than the previous one.

mwleeds: Good catch

js/app/application.js
mwleeds: As far as I could tell it's the only way

js/app/historyStore.js
mwleeds: The same item can cause more than one set of metrics if you go back to it, and the second set of metrics will have a different entry point. So to preserve the immutability of history items I think we'd have to decouple the entry point data from the history item, which would get messy.

mwleeds:
@ptomato pushed some fixed commits (not sure if you get notifications from the review comments)

js/app/historyStore.js
ptomato: But I don't think you can have two open views of the same item at the same time, so just the key variant should be sufficient to deduplicate them. So really you only need the entry point info at the time you record the start metric, which is also exactly when you know the entry point info; so you don't need to save it anywhere. I still think it should be possible.

mwleeds: Are you suggesting that we make duplicate calls to the metrics daemon even though that causes warnings to print? Even if we don't care about that it makes the code messier because the code in `on_changed` dealing with the metrics can't easily be broken out into a function, because you need to use a different entry point just when the history buttons are clicked.

ptomato: I don't understand how that would make duplicate calls to the metrics daemon, since you can only be viewing two items at a time (one on a page, and one in a lightbox on top of the page.) Anyway, this might be a bit overinvolved for a discussion in review comments — I'll try tomorrow to poke at some code to illustrate what I mean. Hopefully that will either convince me that I'm talking nonsense, or else convince you that it will work for your use case :smile:

mwleeds: Right, you don't have to make duplicate calls, but the alternative is kind of messy. I pushed a commit to show what I mean, but I couldn't test it because of some gjs issues.

mwleeds:
(Repeating myself because GitHub hides comments on outdated diffs)
I pushed a commit showing how we might preserve `HistoryItem` immutability, but haven't tested it.

mwleeds:
I tested f52f316555744e17386bbbda09dd26f856112e02 and it works, so it's just a trade off between readability and mutability.

mwleeds:
Alright I pushed a version that I think is ready to merge. It's basically just a combination of my patches and @ptomato's patch to preserve immutability of `HistoryItem`s. I had to change a few things to get the tests working and to prevent it from recording metrics for sets (as opposed to articles).

ptomato:
Looks great, thanks! I'm going to squash the two commits together since tests should be added in the same commit as the functionality, but no need to go through another review round for that. I guess the next step is to figure out which apps need to be ported to SDK2 to take advantage of this feature!

PR #1498, opened 2017-08-01
T16482

tchx84:
https://phabricator.endlessm.com/T16482

### DIFF
--- a/js/app/moltresEngine.js
+++ b/js/app/moltresEngine.js
@@ -202,7 +202,7 @@ placerat varius non id dui.',
         data.title = VIDEO_TITLE; // Override title so we can see which ones are videos
         // Extra slashes are ignored during the resource lookup so we always get the same
         // file. However this does ensure we have unique ekn_ids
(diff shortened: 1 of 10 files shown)


js/app/moltresEngine.js
```diff
@@ -196,6 +205,16 @@ placerat varius non id dui.',
         return Eknc.SetObjectModel.new_from_props(data);
     },
 
+    _generate_audio_object: function (data) {
+        data.synopsis = this._SYNOPSIS;
+        data.content_type = 'audio/webm';
+        data.title = AUDIO_TITLE;
+        data.ekn_id = IMAGES_DIR + VIDEO_FILENAME + new Array(this._audio_count++).fill('/').join('');
```
ptomato: I don't understand this. Am I reading correctly that it generates `resource:///com/endlessm/knowledge/data/images/tools/give_that_man_a_knighthood.webm/////////////`?

In any case if that's intended, you can use `'/'.repeat(this._audio_count++)`.

tchx84: Yes, I was just following the same scheme we use for videos, but I can fix it there as well.

PR #1500, opened 2017-08-15
modules: Big rename of Views

ptomato:
Renames Card.Audio to View.Audio, Card.KnowledgeDocument to
View.Document, Card.Video to View.Video, Card.Media to View.Media,
ContentGroup.MediaLightbox's 'card' slot to 'view', and
Layout.ArticleStack's 'card' slot to 'article'.

https://phabricator.endlessm.com/T16467

### DIFF
--- a/js/app/modules/view/document.js
+++ b/js/app/modules/view/document.js
@@ -9,7 +9,6 @@ const Gtk = imports.gi.Gtk;
 const WebKit2 = imports.gi.WebKit2;
 
 const ArticleContent = imports.app.interfaces.articleContent;
-const Card = imports.app.interfaces.card;
 const EknWebview = imports.app.widgets.eknWebview;
 const HistoryStore = imports.app.historyStore;
 const InArticleSearch = imports.app.widgets.inArticleSearch;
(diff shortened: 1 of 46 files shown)


js/app/modules/view/document.js
```diff
@@ -19,28 +18,27 @@ const PDFView = imports.app.widgets.PDFView;
 const SlidingPanelOverlay = imports.app.widgets.slidingPanelOverlay;
 const TableOfContents = imports.app.widgets.tableOfContents;
 const Utils = imports.app.utils;
+const {View} = imports.app.interfaces.view;
```
ebassi: This is interesting; why the curly braces? Is it worth it leaving a comment?

js/app/modules/view/media.js
```diff
@@ -3,10 +3,10 @@ const Gdk = imports.gi.Gdk;
 const Gio = imports.gi.Gio;
 const Gtk = imports.gi.Gtk;
 
-const Card = imports.app.interfaces.card;
 const Module = imports.app.interfaces.module;
 const Previewer = imports.app.widgets.previewer;
 const Utils = imports.app.utils;
+const {View} = imports.app.interfaces.view;
```
ebassi: Same as above.

js/app/modules/view/video.js
```diff
@@ -4,9 +4,9 @@ const GObject = imports.gi.GObject;
 const Gtk = imports.gi.Gtk;
 
 const ArticleContent = imports.app.interfaces.articleContent;
-const Card = imports.app.interfaces.card;
 const EosKnowledgePrivate = imports.gi.EosKnowledgePrivate;
 const Module = imports.app.interfaces.module;
+const {View} = imports.app.interfaces.view;
```
ebassi: And here…

tests/js/app/modules/view/testAudio.js
```diff
@@ -4,44 +4,41 @@ const Gtk = imports.gi.Gtk;
 const Utils = imports.tests.utils;
 Utils.register_gresource();
 
-const Compliance = imports.tests.compliance;
+const {Audio} = imports.app.modules.view.audio;
```
ebassi: Same as `{View}`.

tests/js/app/modules/view/testDocument.js
```diff
@@ -8,8 +8,8 @@ const WebKit2 = imports.gi.WebKit2;
 const Utils = imports.tests.utils;
 Utils.register_gresource();
 
+const {Document} = imports.app.modules.view.document;
```
ebassi: Same as `{View}` above.

js/app/modules/view/document.js
ptomato: I hope to move to this syntax more widely in new code. so I don't know if it's worth leaving a comment in this particular instance; but it's [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). Basically means take the `View` member of the RHS and assign it to the `View` local variable. It avoids having to refer to `View.View` everywhere as we did with `Card.Card` for example, and adheres more closely to the upcoming `import {View} from "app/interfaces/view"` syntax for ES6 modules.

PR #1503, opened 2017-08-16
constrained: fix forEach() callback warning

xjuan:
Emeus.ConstraintLayout.add_constraint can not be used as a forEach()
callback because it only expect one argument

https://phabricator.endlessm.com/T18829

### DIFF
--- a/js/app/modules/arrangement/constrained.js
+++ b/js/app/modules/arrangement/constrained.js
@@ -109,7 +109,7 @@ var Constrained = new Module.Class({
             0, 0,
             widgets,
             {});
-        constraints.forEach(this.add_constraint, this);
+        constraints.forEach(c => this.add_constraint(c));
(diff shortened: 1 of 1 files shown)


xjuan:
This only fix the warning, the test still segfaults

js/app/modules/arrangement/constrained.js
```diff
@@ -109,7 +109,7 @@ var Constrained = new Module.Class({
             0, 0,
             widgets,
             {});
-        constraints.forEach(this.add_constraint, this);
+        constraints.forEach(c => { this.add_constraint (c); }, this);
```
ptomato: nitpick: Can be done without the extra `this` binding since the arrow function binds it already.
`constraints.forEach(c => this.add_constraint(c));`

PR #1506, opened 2017-08-23
build: Use $TMPDIR for $XDG_DATA_HOME in tests

ptomato:
This should ensure that g_get_user_data_dir() always returns something
that exists, even in a clean Jenkins environment.

https://phabricator.endlessm.com/T18932

### DIFF
--- a/ekncontent/tests/ekncontent/testDomain.js
+++ b/ekncontent/tests/ekncontent/testDomain.js
@@ -1,9 +1,16 @@
 const Eknc = imports.gi.EosKnowledgeContent;
+const Gio = imports.gi.Gio;
+const GLib = imports.gi.GLib;
 
 const InstanceOfMatcher = imports.tests.InstanceOfMatcher;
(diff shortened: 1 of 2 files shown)


ebassi:
Would be nice to delete the temporary directory to avoid leaving around artefacts; I'm also not entirely sure if this will break distcheck.

ptomato:
Hmm, there's not really a hook for Automake to delete the directory after the tests have run. Any suggestions?

ebassi:
The way we do this upstream in GLib is to use a fixture init function that creates the directory and calls `g_setenv()`; then the fixture teardown function deletes the directory. For instance, [like this commit](https://git.gnome.org/browse/glib/commit/?id=0f5b523fac8b2d9b9305419937e51b6c87e765be).

ptomato:
Good idea, here's an implementation.

ebassi:
Looks great.

PR #1507, opened 2017-08-23
historyStore: add SOCIAL_SHARE distpatcher action

xjuan:
https://phabricator.endlessm.com/T18521


js/app/historyStore.js
```diff
@@ -369,6 +375,32 @@ var HistoryStore = new Lang.Class({
...
+
+    can_share_current_item: function () {
+        let item = this.get_current_item();
+        return (item.model &&
+                item.model instanceof Eknc.ContentObjectModel &&
+                item.model.original_uri !== null &&
+                item.model.original_uri !== "");
+    },
+
+    social_share: function (network) {
```
tchx84: Not crucial, but did you have other ideas for this method name? e.g, share_* (since this is the main action)

js/app/historyStore.js
```diff
@@ -369,6 +375,32 @@ var HistoryStore = new Lang.Class({
...
+        if (!this.can_share_current_item())
+          return;
+        let item = this.get_current_item();
+
+	let uri = encodeURIComponent (item.model.original_uri);
```
tchx84: The indentation is messed up in this entire block (from this line and on).

js/app/historyStore.js
```diff
@@ -15,6 +16,8 @@ const Pages = imports.app.pages;
 const ReadingHistoryModel = imports.app.readingHistoryModel;
 const Utils = imports.app.utils;
 
+let _ = imports.gettext.dgettext.bind(null, imports.app.config.GETTEXT_PACKAGE);
```
tchx84: this file needs to be added to the POTFILES file.

js/app/modules/window/simple.js
```diff
@@ -114,8 +121,39 @@ var Simple = new Module.Class({
         this.get_child().show_all();
     },
 
+    _social_box_new: function () {
+	    let revealer = new Gtk.Revealer({
```
tchx84: More indentation issues.

js/app/historyStore.js
```diff
@@ -369,6 +375,32 @@ var HistoryStore = new Lang.Class({
...
+
+        if (network === 'facebook')
+	  Gtk.show_uri (null, 'https://www.facebook.com/dialog/share?app_id=407909575958642&display=popup' +
+                        '&href=' + uri + '&quote=' + msg, Gdk.CURRENT_TIME);
+        else if (network === 'twitter')
+	  Gtk.show_uri (null, 'https://twitter.com/intent/tweet?hashtags=SharedFromEndless' +
+	                '&original_referer=https%3A%2F%2Fendlessos.com&url=' + uri, Gdk.CURRENT_TIME);
+        else if (network === 'whatsapp')
+	  Gtk.show_uri (null, 'https://api.whatsapp.com/send?text=' + uri + '%20' + msg, Gdk.CURRENT_TIME);
+    },
```
tchx84: maybe adding a basic unit test, to make sure this doesn't break?

js/app/historyStore.js
xjuan: Yeah, I just started using Builder last week

js/app/historyStore.js
xjuan: testing the actual codepath would be too complicated, but I can add a simple test that would check the url we are using to share are working

js/app/historyStore.js
xjuan: you mean having one API for each social network?

js/app/actions.js
```diff
@@ -16,3 +16,4 @@ var DBUS_LOAD_QUERY_CALLED = 'dbus-load-query-called';
 var DBUS_LOAD_ITEM_CALLED = 'dbus-load-item-called';
 var SEARCH_BOX_FOCUSED = 'search-box-focused';
 var LIGHTBOX_CLOSED = 'lightbox-closed';
+var SHARE = 'share';
```
tchx84: the commit description should reflect this

js/app/historyStore.js
```diff
@@ -369,6 +381,55 @@ var HistoryStore = new Lang.Class({
...
+            return;
+        let item = this.get_current_item();
+
+        let original_uri = encodeURIComponent(item.model.original_uri);
+        let uri = null;
+
+        switch(network) {
+            case Network.FACEBOOK:
+                uri = 'https://www.facebook.com/dialog/share?app_id=407909575958642&display=popup&hashtag=%23SharedFromEndless&href='
+                      + original_uri;
```
tchx84: I tested this, and works fine. But can we leave that id publicly like this?

tests/js/app/testHistoryStore.js
```diff
@@ -161,4 +162,48 @@ describe('History Store', function () {
...
+            page_type: 'article',
+            model: Eknc.ArticleObjectModel.new_from_props({
+                title: 'Endless OS',
+                ekn_id: 'ekn://article1',
+            }),
+        });
+        expect(history_store.can_share()).toBe(false);
+
+        history_store.set_current_item_from_props({
+            page_type: 'article',
```
tchx84: there is a Pages "enum" we should use, but since this was already used like that in this code, consider this a nit, and if you want to fix it, just add an extra commit fixing all.

js/app/historyStore.js
xjuan: Yes, there is no other way to do it, besides its like the public id of "Endless Social" facebook app

PR #1509, opened 2017-08-30
tests: Ignore trailing space in polib output

ptomato:
I thought the regex '^# ?$' would cover both cases of the output for
polib < 1.0.7 and >= 1.0.7. However, it doesn't. For simplicity, just
ignore all trailing spaces.

https://phabricator.endlessm.com/T18981


ebassi:
Oh, nice; I always forget these options exist in `diff`.

PR #1510, opened 2017-09-01
ekncontent: Fix memory leaks

ptomato:
Picked up with LSan and Valgrind.

https://phabricator.endlessm.com/T18981

### DIFF
--- a/ekncontent/ekncontent/eknc-content-object-model.c
+++ b/ekncontent/ekncontent/eknc-content-object-model.c
@@ -252,6 +252,7 @@ eknc_content_object_model_constructed (GObject *object)
                                          (guchar *) &MOCKED_OBJECTS,
                                          sizeof (MOCKED_OBJECTS));
       MOCKED_OBJECTS++;
+      g_free (priv->ekn_id);
       priv->ekn_id = g_strconcat ("ekn:///", sum, NULL);
     }
 }
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -303,7 +303,8 @@ eknc_domain_process_subscription (EkncDomain *self,
     {
       JsonArray *json_dbs = json_object_get_array_member (json_manifest, "xapian_databases");
       g_autofree gchar *subscription_path = g_file_get_path (bundle_dir);
-      GList *dbs = json_array_get_elements (json_dbs), *l;
+      g_autoptr(GList) dbs = json_array_get_elements (json_dbs);
+      GList *l;
 
       for (l = dbs; l; l = g_list_next (l))
         {
@@ -413,8 +414,14 @@ eknc_domain_initable_init (GInitable *initable,
       /* Find out root relative path from manifest file */
       g_autoptr(GFile) p = g_file_get_parent (self->manifest_file);
       g_autoptr(GString) relative_path = g_string_new ("");
-      while ((p = g_file_get_parent (p)))
-        g_string_append (relative_path, "../");
+      while (TRUE)
+        {
+          g_autoptr(GFile) old_p = g_steal_pointer (&p);
+          p = g_file_get_parent (old_p);
+          if (p == NULL)
+            break;
+          g_string_append (relative_path, "../");
+        }
 
       GList *l;
       for (l = self->subscriptions; l; l = g_list_next (l))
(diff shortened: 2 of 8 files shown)


ekncontent/ekncontent/eknc-domain.c
```diff
@@ -438,8 +447,10 @@ eknc_domain_initable_init (GInitable *initable,
 
       json_object_set_array_member (json_root, "xapian_databases", json_subscriptions);
       json_node_set_object (json_root_node, json_root);
+      json_object_unref (json_root);
```
ebassi: Line 449 and 450 could be replaced with:

```cpp
json_node_take_object (json_root_node, json_root);
```

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -413,8 +415,15 @@ eknc_domain_initable_init (GInitable *initable,
       /* Find out root relative path from manifest file */
       g_autoptr(GFile) p = g_file_get_parent (self->manifest_file);
       g_autoptr(GString) relative_path = g_string_new ("");
-      while ((p = g_file_get_parent (p)))
-        g_string_append (relative_path, "../");
+      while (TRUE)
+        {
+          GFile *old_p = g_steal_pointer (&p);
```
ebassi: This could be:

```cpp
while (TRUE)
  {
    g_autoptr(GFile) old_p = g_steal_pointer (&p);
    p = g_file_get_parent (old_p);
    if (p == NULL)
      break;
  }
```

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -633,6 +644,7 @@ eknc_domain_get_object (EkncDomain *self,
 
   GInputStream *stream = eos_shard_blob_get_stream (record->metadata);
```
ebassi: Worth using `g_autoptr(GInputStream) stream = …` and avoid `g_object_unref()` altogether.

ekncontent/ekncontent/eknc-content-object-model.c
```diff
@@ -252,6 +252,7 @@ eknc_content_object_model_constructed (GObject *object)
                                          (guchar *) &MOCKED_OBJECTS,
                                          sizeof (MOCKED_OBJECTS));
       MOCKED_OBJECTS++;
+      g_clear_pointer(&priv->ekn_id, g_free);
```
ebassi: This is really not necessary:
 - `g_free()` is `NULL`-safe
 - you're changing the pointer right after that
 - `g_strconcat()` will return `NULL` on error

I've seen this pattern elsewhere, and I'd rather we stopped using `g_clear_pointer()` when it's entirely unnecessary.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -549,9 +556,10 @@ eknc_domain_get_shards (EkncDomain *self)
  * Attempts to determine if the given link corresponds to content within
  * this domain.
  *
- * Returns: (transfer full): Returns an EKN URI to that content if so, and NULL otherwise.
+ * Returns: (nullable): Returns an EKN URI to that content if so, and NULL
```
ebassi: The transfer annotation is still useful.

The appropriate way to mark `NULL` in gtk-doc is `%NULL`.

ekncontent/ekncontent/eknc-engine.c
```diff
@@ -156,9 +156,10 @@ eknc_engine_init (EkncEngine *self)
  * Attempts to determine if the given link corresponds to content within
  * the default domain.
  *
- * Returns: Returns an EKN URI to that content if so, and NULL otherwise.
+ * Returns: (nullable): Returns an EKN URI to that content if so, and NULL
```
ebassi: Same as above: `(transfer full)` and `%NULL`.

ekncontent/ekncontent/eknc-engine.c
```diff
@@ -178,9 +179,10 @@ eknc_engine_test_link (EkncEngine *self,
  * Attempts to determine if the given link corresponds to content within
  * the domain for the given application id.
  *
- * Returns: Returns an EKN URI to that content if so, and NULL otherwise.
+ * Returns: (nullable): Returns an EKN URI to that content if so, and NULL
```
ebassi: And, again: `(transfer full)` and `%NULL`.

ekncontent/ekncontent/eknc-domain.c
ptomato: I thought transfer full was the default for non-const types?

ekncontent/ekncontent/eknc-content-object-model.c
ptomato: OK, will change, but why not just always use `g_clear_pointer()` and eliminate the cognitive load?

ebassi: Well, mostly because it's pointless line noise, and `g_clear_pointer()` is not free, as it needs to do an atomic pointer swap, which is a memory barrier and a synchronisation point.

Clearing the pointer is only useful if you're nullifying it, not replacing it with something else immediately afterwards. For GObject we have `g_steal_object()`, but that's because it's easier to use as it does not have an additional argument for the free function. Maybe we could have a `g_steal_pointer()` in GLib.

ekncontent/ekncontent/eknc-domain.c
ebassi: Better explicit than implicit. It means the documentation will also include the transfer annotation.

PR #1511, opened 2017-09-06
Use Xapian directly

ebassi:
https://phabricator.endlessm.com/T18482


ekncontent/ekncontent/eknc-xapian-bridge.c
```diff
@@ -110,9 +113,11 @@ eknc_xapian_bridge_finalize (GObject *object)
 {
   EkncXapianBridge *self = EKNC_XAPIAN_BRIDGE (object);
 
+  g_clear_object (&self->database_manager);
+  g_clear_object (&self->session);
```
ptomato: Can you remove the soup session (and possibly the libsoup dependency) entirely as well?

ebassi: Yes, the goal is to drop that in later commits.

ebassi:
@ptomato Yes, that's part of the cleanup phase; the idea is to land the minimal changes needed to let the existing code keep working and then remove the various bits of indirection caused by the bridge design. I'm already working on that.

ekncontent/ekncontent/eknc-xapian-bridge.c
```diff
@@ -1,19 +1,10 @@
...
-{
-  return g_quark_from_static_string("eknc-xapian-bridge-error-quark");
-}
-
-#define XB_FIX_ENDPOINT "/fix"
-#define XB_QUERY_ENDPOINT "/query"
-#define XB_TEST_ENDPOINT "/test"
-
 /**
  * SECTION:xapian-bridge
```
ptomato: You have to remove this doc comment as well, and remove the `xapian-bridge` section from `ekncontent/docs/reference/ekncontent/eos-knowledge-content-docs.xml`.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -1093,12 +1083,12 @@ eknc_domain_read_uri (EkncDomain *self,
   return TRUE;
 }
 
-/**
+/*< private >
  * eknc_domain_get_impl:
```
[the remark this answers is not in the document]
ebassi: This function is private, so it should not be in the API reference at all

ekncontent/ekncontent/eknc-xapian-bridge.c
ebassi: The whole file is going away anyway.

ekncontent/ekncontent/eknc-database-manager-private.h
```diff
@@ -0,0 +1,68 @@
+/* Copyright 2014  Endless Mobile
```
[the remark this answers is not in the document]
ebassi: Yeah, the API reference bit needs some love…

ekncontent/ekncontent/eknc-engine.c
```diff
@@ -480,14 +474,23 @@ eknc_engine_get_domain_for_app (EkncEngine *self,
   g_return_val_if_fail (app_id && *app_id, NULL);
   g_return_val_if_fail (error == NULL || *error == NULL, NULL);
 
-  if (g_hash_table_contains (self->domains, app_id))
-    return g_hash_table_lookup (self->domains, app_id);
+  EkncDomain *domain = g_hash_table_lookup (self->domains, app_id);
```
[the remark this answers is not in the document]
ebassi: Good point.

ebassi:
EkncXapianBridge is no more; I'm a little bit more sceptical about dropping the database manager code. As it is now, it's fairly self-contained and it would need to be dumped into `EkncDomain` anyway.

ekncontent/ekncontent/eknc-engine.c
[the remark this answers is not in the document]
ebassi: Looking at this a bit better, we seem to have a lot of API that assumes you can get an `EkncEngine` instance (typically through `eknc_engine_get_default()`) and then use the `for_app` variants of the API. Should this API also go away?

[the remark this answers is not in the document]
ptomato: Still depends on whether we need to keep that API for the search provider.

I _think_ as far as e-k-l is concerned, Engine and Domain could just be merged altogether. The separation was necessary because Domain was the DB-version-specific implementation, because e-k-l needed to handle all DB versions that we ever published. But now the flatpak runtime takes care of that abstraction.

ekncontent/docs/reference/ekncontent/Makefile.am
```diff
@@ -1,73 +1,43 @@
...
-# Images to copy into HTML directory.
-# e.g. HTML_IMAGES=$(top_srcdir)/gtk/stock-icons/stock_about_24.png
-HTML_IMAGES=
+IGNORE_HFILES = \
+	eknc-enums.h \
+	eknc-content-object-model-private.h \
+	eknc-database-manager-private.h \
+	eknc-domain-private.h \
+	eknc-media-object-model-private.h \
+	eknc-utils-private.h
```
ptomato: Adding this introduced one more documentation failure - we need to remove `eknc_utils_id_get_hash` from the sections.txt file, since it's in a private header file.

ekncontent/ekncontent/eknc-engine.c
```diff
@@ -480,14 +473,23 @@ eknc_engine_get_domain_for_app (EkncEngine *self,
...
-  if (!(domain = eknc_domain_get_impl (app_id, NULL, self->xapian_bridge, NULL, error)))
+  if (domain != NULL)
+    return domain;
+
+  domain = eknc_domain_get_for_app_id (app_id, NULL, self->language, NULL, error);
+  if (domain == NULL)
     return NULL;
+
+  // Bind our own language property to the domain
+  g_object_bind_property (self, "language",
```
ptomato: This doesn't work (and isn't necessary) now that the property is construct-only.

ekncontent/ekncontent/eknc-engine.c
```diff
@@ -506,17 +508,23 @@ eknc_engine_add_domain_for_path (EkncEngine *self,
...
   if (g_hash_table_contains (self->domains, app_id))
     return;
 
-  domain = eknc_domain_get_impl (app_id, path, self->xapian_bridge, NULL, error);
+  EkncDomain *domain = eknc_domain_get_for_app_id (app_id, path, self->language, NULL, error);
+  if (domain == NULL)
+    return;
+
+  // Bind our own language property to the domain
+  g_object_bind_property (self, "language",
```
ptomato: Ditto

ptomato:
PS. To get `flapjack test` to work I rebased the branch on master

ekncontent/docs/reference/ekncontent/Makefile.am
ebassi: Fixed

ebassi:
Ended up with a further simplification of `EkncDatabaseManager` and `EkncDomain`, and now I've dropped the JSON ping-pong altogether.

ptomato:
I'm not sure about "Remove the 1:N database support from DatabaseManager". I think the number of databases per domain is _constant_ unlike it used to be, but not necessarily only _one_.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -1013,11 +1040,16 @@ ptr_array_to_list_deep (GPtrArray *array)
...
     return NULL;
 
-  for (guint i = 0; i < array->len; ++i)
-    list = g_list_append (list, g_object_ref (g_ptr_array_index (array, i)));
+  /* Walk the list backwards so we can use prepend() to get constant
+   * time insertion in the results list, instead of having to use
+   * append(), or to use prepend() and then reverse() before returning
+   * it at the end
+   */
+  for (guint i = array->len - 1; i >= 0; i--)
```
ptomato: Here's the crash that you are currently getting; if `array->len == 0` then you are counting backwards from 4294967296...

This worked for me:
```c
  for (guint i = array->len; i > 0; i--)
    list = g_list_prepend (list, g_object_ref (g_ptr_array_index (array, i - 1)));
```
At least this gets it back to the previous state where a window shows up, but no content.

ebassi: Duh. Fixed, and now I'm getting a lovely stack corruption I can debug :-)

ebassi:
@ptomato Ideally we could have multiple sub-databases associated to a single database, if we're querying a set of databases.

ptomato:
As far as I can tell the errors about converting GVariants aren't a database schema change; rather, the conversion is failing at [this point in json-glib](https://gitlab.gnome.org/GNOME/json-glib/blob/master/json-glib/json-gvariant.c#L909) with `errno == EAGAIN`. How bizarre.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -39,7 +40,10 @@ struct _EkncDomain
   gchar *path;
   GList *subscriptions;
 
-  EkncXapianBridge *xapian_bridge;
+  char *language;
```
ptomato: Left over from debugging?

ptomato:
I did check if the multiple-shard support was working properly; and Kapan Lagi still launches and runs correctly, so that seems good.

ekncontent/ekncontent/eknc-domain.c
ebassi: Not really; we pass the language from the engine to the domain at construction time because the domain has to know how to build the query parser — which implies known how to create a stemmer for the given language.

ptomato: Never mind! I read the diff wrong and thought this was a global variable.

ptomato:
Bad news; running the animals app on the existing SDK with xapian-bridge logging to a console, didn't show any errors. So the change _must_ be somewhere in here... could be that the problem was already there but the errors were swallowed.

Point taken that the branch is already big enough. I'll merge it and open a follow up issue.

PR #1513, opened 2017-09-14
T18520 - Sharing feature (+ save and print) icons

katelk28:
Adding icons for the sharing feature for our apps.
- A simpler facebook icon
- Save icon
- Printer icon taken from gnome-symbolic-icons

https://phabricator.endlessm.com/T18520


data/icons/scalable/apps/printer-symbolic.svg
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
ptomato: Is this icon the same as the `printer-symbolic.svg` included in the GNOME icon theme? If so, we don't need to include it here.

katelk28:
@ptomato Thanks. Ready again!

ptomato:
I saw your comment on the Phabricator issue, that would have answered my earlier question. Sorry about that.

Two more questions:
- Should the new facebook icon replace the one that Juan Pablo added before? In that case it should have the same filename.
- The save icon is really similar to Adwaita's save icon. Can we either use the Adwaita one, or replace the Adwaita one wholesale? In the latter case, the file would have to be named `actions/document-save-symbolic.svg`.

For reference, the Adwaita symbolic icons are here: https://github.com/GNOME/adwaita-icon-theme/tree/master/Adwaita/scalable

ptomato:
One other thing I forgot: the new files need to be added to https://github.com/endlessm/eos-knowledge-lib/blob/master/eos-knowledge.gresource.xml#L16-L19

katelk28:
@ptomato Done for most of your requests -
For the save icon, I did consider using the Adwaita one, but I'm not a big fan of it (it's a bit busy). But interesting idea about adding it to Adwaita; I'll contact the GNOME designers about it. If it doesn't happen, I think it's fine that we use this custom one.

ptomato:
By "replace the Adwaita icon" I meant in our icon theme used in Endless apps. Agreed, looking into modifying the original Adwaita could be nice, but that's up to you and can definitely be separate from this pull request. But I meant that we shouldn't have two save icons that apps can choose from, so if an app uses "the save icon" they should get yours instead of Adwaita's. For that, it needs to be named `document-save-symbolic.svg`.

katelk28:
@ptomato Ah! Totally misunderstood that. That's a good idea. 
Do I need to make a PR for that somewhere else? Or will it be replaced automatically if I use that name and keep it here in e-k-l?

ptomato:
It will work fine if you keep it here and just rename the file. (Although note that if we had everything set up correctly, the PR would have to go somewhere else. See https://phabricator.endlessm.com/T18342)

PR #1517, opened 2017-09-25
T19304

tchx84:
https://phabricator.endlessm.com/T19304


js/app/modules/arrangement/sideBySide.js
```diff
@@ -59,6 +59,7 @@ var SideBySide = new Module.Class({
             hscrollbar_policy: Gtk.PolicyType.NEVER,
             propagate_natural_width: true,
             propagate_natural_height: true,
+            max_content_height: 480,
```
ptomato: If we want this to fit in 720x480 composite resolution, does it have to be slightly less to account for the borders on the popover?

tchx84: true, ill change it!

PR #1518, opened 2017-09-27
T18489

tchx84:
https://phabricator.endlessm.com/T18489

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -370,6 +370,39 @@ eknc_domain_process_subscription (EkncDomain *self,
 }
 
 static gboolean
+eknc_domain_import_subscriptions (EkncDomain *self,
+                                  GFile *subscriptions_dir,
+                                  JsonArray *json_subscriptions,
+                                  const gchar *relative_path,
+                                  GCancellable *cancellable,
+                                  GError **error)
+{
+  g_autoptr(GFileEnumerator) subscriptions_iter = g_file_enumerate_children (subscriptions_dir, "",
+                                                                             G_FILE_QUERY_INFO_NONE,
+                                                                             cancellable, error);
+  while (TRUE)
+    {
+      GFile *dir = NULL;
+      GFileInfo *info = NULL;
+      if (!g_file_enumerator_iterate (subscriptions_iter, &info, &dir, cancellable, error))
+        return FALSE;
+      if (dir == NULL)
+        break;
+      if (g_file_info_get_file_type (info) != G_FILE_TYPE_DIRECTORY)
+        continue;
+      if (!eknc_domain_process_subscription (self,
+                                             dir,
+                                             json_subscriptions,
+                                             relative_path,
+                                             cancellable,
+                                             error))
+          return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
 eknc_domain_initable_init (GInitable *initable,
                            GCancellable *cancellable,
                            GError **error)
@@ -413,12 +446,7 @@ eknc_domain_initable_init (GInitable *initable,
       if (!eknc_parse_subscriptions (self, cancellable, error))
         return FALSE;
 
-      g_autoptr(GFile) subscriptions_dir = NULL;
       JsonArray *json_subscriptions = json_array_new ();
-
-      self->content_dir = eknc_get_data_dir (self->app_id);
-      subscriptions_dir = g_file_get_child (self->content_dir,
-                                            "com.endlessm.subscriptions");
 
       /* Find out root relative path from manifest file */
       g_autoptr(GFile) p = g_file_get_parent (self->manifest_file);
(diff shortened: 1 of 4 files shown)


ekncontent/ekncontent/eknc-domain.c
```diff
@@ -439,6 +468,20 @@ eknc_domain_initable_init (GInitable *initable,
               return FALSE;
         }
 
+      /* get subscriptions from extensions */
+      g_autofree gchar *extensions_path = g_build_filename ("/app", "share", "ekn", "subscriptions", NULL);
```
tchx84: hmm, I think I need to consider the case where this is ran from outside the app itself, like from EknServices.

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -370,6 +379,96 @@ eknc_get_data_dir (const gchar *app_id)
...
+GList *
+eknc_get_extensions_dirs (const gchar *app_id)
+{
+  GList *list = NULL;
+  g_autofree gchar *system_path = NULL;
+  g_autofree gchar *split_path = NULL;
+
+  system_path = g_build_filename ("/var", "lib",
+                                  NULL);
+  list = databases_dirs_from_metadata (system_path, app_id);
```
xjuan: Why dont you pass "/var/lib" directly to databases_dirs_from_metadata()?

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -370,6 +379,96 @@ eknc_get_data_dir (const gchar *app_id)
...
+  if (list != NULL)
+    return list;
+
+  split_path = g_build_filename ("/var", "endless-extra",
+                                 NULL);
+  list = databases_dirs_from_metadata (split_path, app_id);
```
xjuan: same here

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -370,6 +379,96 @@ eknc_get_data_dir (const gchar *app_id)
...
+  if (list != NULL)
+    return list;
```
xjuan: No need for this check, you can simply
`return databases_dirs_from_metadata ("/var/endless-extra", app_id);`

Also, dont you need to return both list concatenated ?

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -423,21 +450,39 @@ eknc_domain_initable_init (GInitable *initable,
...
-                                                 error))
-              return FALSE;
-        }
+      /* Import subscriptions from data directory */
+      g_autoptr(GFile) subscriptions_dir = NULL;
+      self->content_dir = eknc_get_data_dir (self->app_id);
+      subscriptions_dir = g_file_get_child (self->content_dir,
+                                            "com.endlessm.subscriptions");
+      if (g_file_query_exists (subscriptions_dir, NULL))
+      {
```
xjuan: We indent brackets too

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -423,21 +450,39 @@ eknc_domain_initable_init (GInitable *initable,
...
+                                               relative_path->str,
+                                               cancellable,
+                                               error))
+            return FALSE;
+      }
+
+      /* Import subscriptions from extensions directories */
+      g_autoptr(GList) extensions_dirs = eknc_get_extensions_dirs (self->app_id);
+      while (extensions_dirs != NULL)
+      {
```
xjuan: same here

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -332,12 +334,19 @@ eknc_get_data_dir (const gchar *app_id)
   g_autofree gchar *user_path = NULL;
   g_autofree gchar *system_path = NULL;
   g_autofree gchar *split_path = NULL;
+  g_autofree gchar *mount_point = NULL;
+
+  // Try the extensions mount point first
+  mount_point = g_build_filename ("/app", "share",
+                                  NULL);
+  ret = database_dir_from_data_dir (mount_point, app_id);
```
xjuan: Cant you use "/app/share" directly?

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -370,6 +379,96 @@ eknc_get_data_dir (const gchar *app_id)
...
+      g_autoptr(GFile) extension_dir = NULL;
+      gchar *extension_name = NULL;
+      g_autofree gchar *extension_version = NULL;
+      g_autofree gchar *extension_path = NULL;
+
+      if (!g_str_has_prefix (groups[i], FLATPAK_EXTENSION_GROUP))
+        continue;
+
+      extension_name = (extension_name = (groups[i] + strlen (FLATPAK_EXTENSION_GROUP)));
+      if (!g_str_has_prefix (extension_name, app_id))
```
xjuan: You could use g_strstr_len() directly in the previous check instead of this strlen magic :)

ekncontent/ekncontent/eknc-utils.c
tchx84: These are exclusive conditions, the same logic is implemented in `eknc_get_data_dir` and I checked in my mini, the apps and runtimes are installed in `/var/endless-extra`.

If it wasn't the case then it would mean that our search-provider were historically looking at the wrong version of the content.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -370,6 +370,38 @@ eknc_domain_process_subscription (EkncDomain *self,
...
+{
+  g_autoptr(GFileEnumerator) subscriptions_iter = g_file_enumerate_children (subscriptions_dir, "",
+                                                                             G_FILE_QUERY_INFO_NONE,
+                                                                             cancellable, error);
+  while (subscriptions_iter != NULL)
+    {
+      GFile *dir = NULL;
+      GFileInfo *info = NULL;
+      g_file_enumerator_iterate (subscriptions_iter, &info, &dir, cancellable, error);
+      if (dir == NULL)
```
xjuan: nit: if g_file_enumerator_iterate() returns false then there was an error
```
if(!g_file_enumerator_iterate (subscriptions_iter, &info, &dir, cancellable, error))
    break;
```

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -432,22 +459,40 @@ eknc_domain_initable_init (GInitable *initable,
...
+            {
+              if (!eknc_domain_import_subscriptions (self,
+                                                     extension_dir,
+                                                     json_subscriptions,
+                                                     relative_path->str,
+                                                     cancellable,
+                                                     error))
+                  return FALSE;
+            }
+          extensions_dirs = g_list_next(extensions_dirs);
```
xjuan: Sorry I missed this yesterday; you are leaking the list here because once extensions_dirs goes out of scope it will be pointing to NULL so there wont be anything to free.
You have to use another variable to iterate over it, this is why I prefer to use a for statement

`for (l = extensions_dirs; l; l = g_list_next(l))`

ekncontent/ekncontent/eknc-utils.c
```diff
@@ -370,6 +374,93 @@ eknc_get_data_dir (const gchar *app_id)
...
+      extension_path = g_build_filename (flatpak_path, "flatpak",
+                                         "runtime", extension_name,
+                                         arch, extension_version,
+                                         "active", "files",
+                                          NULL);
+      extension_dir = g_file_new_for_path (extension_path);
+      if (!g_file_query_exists (extension_dir, NULL))
+        continue;
+
+      list = g_list_prepend (list, g_file_dup(extension_dir));
```
xjuan: nit: you could increment the reference count with g_object_ref() instead of creating a new object

ekncontent/ekncontent/eknc-domain.c
tchx84: good catch, I think it should `return FALSE;` in that case, to propagate the error properly.

xjuan:
CR Pass
homer_bird

PR #1519, opened 2017-09-27
EknMediaBin: Remove revealer timeout on dispose

ptomato:
This was causing use-after-free when switching from one video to another.

https://phabricator.endlessm.com/T19444

### DIFF
--- a/lib/eosknowledgeprivate/ekn-media-bin.c
+++ b/lib/eosknowledgeprivate/ekn-media-bin.c
@@ -259,6 +259,16 @@ revealer_timeout (gpointer data)
   return G_SOURCE_REMOVE;
 }
 
+static inline void
+ensure_no_timeout(EknMediaBinPrivate *priv)
+{
+  if (!priv->timeout_id)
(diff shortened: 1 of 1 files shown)


lib/eosknowledgeprivate/ekn-media-bin.c
```diff
@@ -869,6 +869,12 @@ ekn_media_bin_dispose (GObject *object)
   EknMediaBin *self = EKN_MEDIA_BIN (object);
   EknMediaBinPrivate *priv = EMB_PRIVATE (self);
 
+  if (priv->timeout_id)
+    {
+      g_source_remove (priv->timeout_id);
+      priv->timeout_id = 0;
+    }
+
```
xjuan: We need to remove this from finalize() and while we are at it we could move it to a new inline function
ekn_media_bin_revealer_timeout_clear() and reuse it in ekn_media_bin_revealer_timeout()

xjuan:
CR Pass

PR #1522, opened 2017-10-02
Layout.ArticleStack: Stop video before switching

ptomato:
Apparently if the video player is running it will starve out the main
loop so that the load_content_promise() in the next line will resolve
only after a few seconds. I'm not sure why this happens, but we can work
around it by stopping the video before loading a new one.

https://phabricator.endlessm.com/T19384


xjuan:
CR Pass

PR #1526, opened 2017-10-05
Gallery template: Migrate two styles from the template to modules

katelk28:
Integrating 2 styles from the gallery template into the framework:
- Pager.Simple margin-top: 0
- Window.Simple scrollbar styling

https://phabricator.endlessm.com/T18966

### DIFF
--- a/data/css/gallery.scss
+++ b/data/css/gallery.scss
@@ -45,23 +45,3 @@
     padding-left: 50px;
     padding-right: 50px;
 }
-
-// FIXME: These 2 selector styles should be integrated into the framework.
-.PagerSimple {
-    margin-top: 0;
-}
-
-.WindowSimple {
-    scrollbar {
-        padding: 3px;
-        background-color: #c3c4c4;
-        border: 1px solid #b6b6b2;
-        // margin-top is for being covered by Layout.TopMenu
-        margin-top: 50px;
-
-        slider {
-            min-width: 6px;
-            background-color: #787c7d;
-        }
-    }
-}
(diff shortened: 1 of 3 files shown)


data/css/gallery.scss
```diff
@@ -45,23 +45,3 @@
...
-    margin-top: 0;
-}
-
-.WindowSimple {
-    scrollbar {
-        padding: 3px;
-        background-color: #c3c4c4;
-        border: 1px solid #b6b6b2;
-        // margin-top is for being covered by Layout.TopMenu
-        margin-top: 50px;
```
ptomato: Does this rule need to remain in the gallery CSS?

data/css/modules/window/_simple.scss
```diff
@@ -4,11 +4,12 @@
 
     scrollbar {
         padding: 3px;
-        background-color: transparentize(black, 1 - 0.20);
+        background-color: #c3c4c4;
```
ptomato: I tried this out on history.en and animals.en as representative preset A and B apps. In preset B it looked out of place in my opinion. The background color of the scrollbar was darker, and that makes the border stand out too much. However, I can't find where the background color was overridden (not in preset_b.scss, nor animals.en's overrides.scss) so maybe I'm testing wrong and there's nothing to do here?

In preset A it would normally look fine, I think, except when there's a darker-colored background such as in history.en; in that case the scrollbar stands out because it's pretty light-colored. Consider moving the previous color into some of the app overrides for preset A apps with darker-colored backgrounds?

katelk28:
@ptomato I decided that this should be closed. The issue arising from this is that our templates should get a scrollbar audit to make sure this is implemented correctly, and I don't have bandwidth for that at the moment. I'll re-name the ticket.;

PR #1528, opened 2017-10-05
Add share actions to article template

xjuan:
https://phabricator.endlessm.com/T18523


xjuan:
Move back to dev after review, still need add the share buttons in the lightbox

js/app/widgets/eknWebview.js
```diff
@@ -8,10 +8,13 @@ const GObject = imports.gi.GObject;
 const Gtk = imports.gi.Gtk;
 const WebKit2 = imports.gi.WebKit2;
 
+const Actions = imports.app.actions;
```
tchx84: I think would be better to move this out of the widget and put it in the Document module.

xjuan: yes! Done!

PR #1530, opened 2017-10-09
T19428

tchx84:
https://phabricator.endlessm.com/T19428


tchx84:
This can be merged after https://github.com/endlessm/eos-knowledge-lib/pull/1518

tchx84:
@erikos @smspillaz does the discovery feed stuff depend on subscriptions.json?

erikos:
Hi Martin, we only add a view fields to the manifest-override.json (like discovery-feed interface information, language...). No dependency on subscriptions.json

smspillaz:
@tchx84 Yup, can confirm, no dependency on `subscriptions.json`. Thanks for checking!

xjuan:
CR Pass

PR #1531, opened 2017-10-09
MediaLightBox: add share actions under close button

xjuan:
https://phabricator.endlessm.com/T18523

### DIFF
--- a/js/app/historyStore.js
+++ b/js/app/historyStore.js
@@ -389,8 +389,9 @@ var HistoryStore = new Lang.Class({
         if (!item || !item.can_share)
             return;
 
+        let model = item.media_model || item.model;
         let app = Gio.Application.get_default();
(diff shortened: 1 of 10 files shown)


js/app/historyStore.js
```diff
@@ -389,8 +389,9 @@ var HistoryStore = new Lang.Class({
         if (!item || !item.can_share)
             return;
 
+        let model = item.media_model || item.model;
```
tchx84: later in this method, the metric system is still referring to item.model, instead of model, I guess we want to track the real thing we shared.

xjuan: Good catch, I missed that one

tchx84:
Nice work ! looks and works nicely.

screenshot from 2017-10-10 12-05-46

PR #1534, opened 2017-10-17
Manually parse some GVariant properties

ebassi:
This mitigates a memory corruption issue that causes critical errors to
be printed in the console when using JSON-GLib to convert JSON into
GVariant for array-based properties that GJS does not support.

The memory corruption is still there, and will present itself given
enough random application launches of some of the knowledge apps.


ebassi:
I ran the tests and ran the apps, indeed.

PR #1536, opened 2017-10-23
Layout.DynamicBackground: break recursive style-updated signal emission

xjuan:
Calling get_background_color() on a style context makes the context emit
the style-updated signal which was making the handler be called continuously.

https://phabricator.endlessm.com/T19667


tchx84:
This certainly would consume a lot of CPU, and after applying this patch the difference is quite noticeable. I guess is another component of the whole problem, plus the linear-gradient workaround (and others we noticed during ours tests).

Merged.

PR #1539, opened 2017-11-06
T19561

tchx84:

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -43,6 +43,7 @@ struct _EkncDomain
   char *language;
 
   EkncDatabaseManager *db_manager;
+  GMutex db_lock;
 
   GFile *content_dir;
   GFile *manifest_file;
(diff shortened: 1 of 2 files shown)


ekncontent/ekncontent/eknc-domain.c
```diff
@@ -136,6 +137,7 @@ eknc_domain_finalize (GObject *object)
   g_free (self->language);
 
   g_clear_object (&self->db_manager);
+  g_mutex_clear  (&self->db_lock);
```
ptomato: nitpick: one space

ptomato:
This time testing the `waatt` branch I also got the JSON conversion error one out of 10 times. So, while this definitely makes the problem much less likely, it does seem to still be missing a lock somewhere.

ptomato:
For the record, here are the occasional messages that I get when testing this branch:
```
Gjs-Message: JS WARNING: [resource:///com/endlessm/knowledge/js/app/utils.js 50]: reference to undefined property "GdkX11Window"
```
```
(com.endlessm.history.en:5): EosKnowledgeContent-CRITICAL **: Unable to convert field 'discovery-feed-content' from JSON to a 'a{sv}' variant: Invalid string value converting to GVariant
```

ptomato:
Just occurred to me that you can probably also revert Emmanuele's previous change for this issue, since that was only meant to arbitrarily shuffle some bits around in the hope that it would crash less.

PR #1540, opened 2017-11-08
Improve queries with spelling fixes

ebassi:
If a query gets fixes for spelling errors we want to include the original terms in the query string, in order to have a more accurate set of results.

https://phabricator.endlessm.com/T19582


ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -581,6 +581,40 @@ get_title_clause2 (EkncQueryObject *self, gchar **terms)
 }
 
 static gchar *
+get_title_clause3 (EkncQueryObject *self,
+                   char **terms,
+                   char **corrected_terms)
+{
+  guint terms_length = g_strv_length (terms);
```
ptomato: nitpick: you've changed `g`types to plain old types elsewhere, how about using `unsigned` or `size_t` here? (Same remark for `g_new0 (char *, ...)` a few lines below.)

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -581,6 +581,40 @@ get_title_clause2 (EkncQueryObject *self, gchar **terms)
 }
 
 static gchar *
+get_title_clause3 (EkncQueryObject *self,
```
ptomato: Maybe out of scope for this, but it might be good to clean up these `...2` and `...3` variants - I think they were only needed for backwards compatibility in xapian-bridge. At least only `get_query_clause2` and not `get_query_clause` is used as far as I can tell.

Should this really be named `get_title_clause3` since it's not used by a corresponding `get_query_clause3`?

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -940,11 +937,13 @@ query_task (GTask *task,
...
     return;
 
-  g_autoptr(GMutexLocker) db_lock = g_mutex_locker_new (&state->domain->db_lock);
+  g_autofree char *query_str = eknc_query_object_get_corrected_query_string (state->query);
+
+  g_debug (G_STRLOC ": Query: '%s'", query_str);
 
-  g_autoptr(XapianMSet) results =
-    eknc_database_manager_query (state->db_manager, state->params, &error);
+  g_autoptr(XapianMSet) results = NULL;
```
ptomato: Not sure why the split here, but OK

ptomato:
Also, the gtk-doc stuff in ekncontent has to be updated - use `flapjack test eos-knowledge-lib` with `extra_config_opts = --enable-gtk-doc` in your eos-knowledge-lib section to verify if it's working.

ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -669,14 +669,14 @@ parse_query_flags (const gchar *str,
 
 static gboolean
 eknc_database_manager_fix_query_internal (EkncDatabaseManager *self,
+                                          const char *query_str,
```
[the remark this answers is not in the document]
ebassi: This is an artefact caused by rebasing on master; it doesn't make any difference, because the quest string is retrieved exactly the same way on both ends.

Originally, I moved the query string out of the hash table and into an argument because I was passing different queries with the same parameters; but now that the query string is one, it's less useful, though a fairly logical clean up.

For SDK3 I want to nuke the parameters hash table, and just pass the QueryObject instance around.

ekncontent/ekncontent/eknc-domain.c
```diff
@@ -940,10 +938,11 @@ query_task (GTask *task,
   if (g_task_return_error_if_cancelled (task))
     return;
 
-  g_autoptr(GMutexLocker) db_lock = g_mutex_locker_new (&state->domain->db_lock);
```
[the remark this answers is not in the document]
ebassi: Another rebase artefact; I thought all locking had been moved from the DatabaseManager to the Domain.

ekncontent/ekncontent/eknc-domain.c
ebassi: Intermediate state between the "run two queries" and the "run a better query" approaches.

I think I should just squash most of the PR.

ekncontent/ekncontent/eknc-query-object.c
ebassi: There are three `get_title_clause()` implementations, but only one is correct — and effectively called. I need to mop up this mess.

ekncontent/ekncontent/eknc-query-object.c
ebassi: `gchar` is a ridiculous type, but `guint` saves characters. :-)

To be fair, I'd go over the whole thing and just use `bool` from `stdbool.h`, and the integer types from `stdint.h`, since we're using C99.

ekncontent/docs/reference/ekncontent/eos-knowledge-content-sections.txt
```diff
@@ -137,16 +137,18 @@ EkncQueryObjectMode
...
 EkncQueryObjectSort
 EkncQueryObjectOrder
+eknc_query_object_get_corrected_query_string
 eknc_query_object_get_cutoff
 eknc_query_object_get_excluded_ids
 eknc_query_object_get_excluded_tags
+eknc_query_object_get_filter_string
+eknc_query_object_get_filter_out_string
 eknc_query_object_get_ids
 eknc_query_object_get_query_string
```
ptomato: `eknc_query_object_get_query` is missing from this list, the gtk-doc test fails.

ebassi: Fixed, and `make check` now passes inside ekncontent.

ebassi:
gtk-doc fixes and PR updated, merging to master.

PR #1541, opened 2017-11-09
Jhbuild moduleset

manuq:


manuq:
@ptomato I totally agree. I will send another request for changes once I get familiar with flapjack.

PR #1542, opened 2017-11-10
T19561 (part2)

tchx84:


tchx84:
moved to dev for obvious omissions...

ebassi:
Looks like a good workaround, until we nuke all the GVariant properties in the model objects.

PR #1548, opened 2017-11-22
T20002

tchx84:
https://phabricator.endlessm.com/T20002


js/app/modules/layout/articleStack.js
```diff
@@ -213,7 +213,8 @@ var ArticleStack = new Module.Class({
         if (GLib.uri_parse_scheme(uri) === 'ekn') {
             Eknc.Engine.get_default().get_object_promise(uri)
             .then((article_model) => {
-                this._webview_tooltip_presenter.show_default_tooltip(tooltip, article_model.title);
+                if (article_model.title)
```
xjuan: Would not be better to add this check in the API itself?

xjuan:
CR Pass

PR #1549, opened 2017-11-24
T20077

tchx84:
https://phabricator.endlessm.com/T20077


data/css/modules/view/_base.scss
```diff
@@ -11,6 +11,11 @@
             font-size: 55px;
             padding: 1.0em 1.0em 0.5em 1.0em;
         }
+
+        /* XXX re-consider how to split Views and Cards better */
```
ptomato: Please open an issue describing what should be done here instead of leaving a comment :-)

PR #1552, opened 2017-12-10
Remove unused eknc_database_manager_get_instance_private() calls

ojwb:
I was freeing up disk space and found this unpushed patch in my EKL tree - `priv` isn't actually used in either case.

I seem to not have all the currently required ekl dependencies installed, so haven't tested this change as updated to current master, but it seems obvious and useful to push.


ptomato:
Indeed, that seems an obvious improvement. Thanks, Olly.

PR #1556, opened 2017-12-12
T20337

tchx84:
https://phabricator.endlessm.com/T20337


xjuan:
CR Pass

Do not forget to open a new one about the button allocation so I can take a look at it

PR #1561, opened 2017-12-15
T20343

tchx84:
https://phabricator.endlessm.com/T20343

### DIFF
--- a/ekncontent/ekncontent/eknc-domain.c
+++ b/ekncontent/ekncontent/eknc-domain.c
@@ -457,6 +457,13 @@ eknc_domain_initable_init (GInitable *initable,
       return FALSE;
     }
 
+  if (self->shards == NULL)
+    {
+      g_set_error (error, EKNC_DOMAIN_ERROR, EKNC_DOMAIN_ERROR_EMPTY,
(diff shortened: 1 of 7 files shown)


ekncontent/ekncontent/eknc-domain.c
```diff
@@ -457,6 +457,13 @@ eknc_domain_initable_init (GInitable *initable,
       return FALSE;
     }
 
+  if (g_slist_length(self->shards) == 0)
```
xjuan: It would be faster to check if its NULL

js/app/utils.js
```diff
@@ -584,3 +584,11 @@ function parse_uri (uri_str, parse_query) {
 
     return {scheme, user, password, host, port, path, query, fragment};
 }
+
+function open_app_center_details (app_id) {
```
xjuan: Do you think this will be used in other place?

tchx84: nope, as we discussed in slack, I will move it back to the Dialog.

tchx84:
@xjuan thanks, should be ready now.

xjuan:
CR Pass

PR #1562, opened 2017-12-19
T14287

tchx84:
https://phabricator.endlessm.com/T14287

### DIFF
--- a/data/preset/A.yaml
+++ b/data/preset/A.yaml
@@ -1,18 +1,53 @@
 ---
-defines:
-- &card
-  Card.Default
+vars:
+  app-banner:
+    shortdef: 'Banner.Dynamic(mode: full, halign: center)'
 
-- &main-arrangement-slot
-  arrangement:
+  home-sets-arrangement:
+    shortdef: 'Arrangement.Overflow(orientation: horizontal)'
+  home-sets-card:
+    type: &card Card.Default
+    properties:
+      excluded_types:
+        - 0
+        - 2
+  home-allsets-arrangement:
     type: Arrangement.SquareGuys
     properties:
       halign: center
       valign: start
       expand: true
-    slots:
-      card: *card
+  home-allsets-card:
+    type: *card
+    properties:
+      excluded_types:
+        - 0
+        - 2
 
+  set-articles-arrangement:
+    type: Arrangement.SquareGuys
+    properties:
+      halign: center
+      valign: start
+      expand: true
+  set-articles-card:
+    type: *card
+  set-articles-order:
+    type: Order.Alphabetical
+
+  search-articles-arrangement:
+    type: Arrangement.SquareGuys
+    properties:
+      halign: center
+      valign: start
+      expand: true
+  search-articles-card:
+      type: *card
+
+  noresults-message-text:
+      shortdef: 'ContentGroup.NoResultsMessage(valign: center, halign: center, justify: center)'
+
+defines:
 - &home-page
   type: Layout.HamburgerBasement
   properties:
@@ -21,7 +56,7 @@ defines:
         source-id: sets-group
         property: has-more-content
   slots:
-    top: 'Banner.Dynamic(mode: full, halign: center)'
+    top: $app-banner
     middle: 'Navigation.SearchBox(halign: center, focus-on-map: true)'
     bottom:
       type: ContentGroup.ContentGroup
@@ -32,14 +67,9 @@ defines:
         valign: center
       slots:
         arrangement:
-          shortdef: 'Arrangement.Overflow(orientation: horizontal)'
+          refvar: $home-sets-arrangement
           slots:
-            card: 
-              type: *card
-              properties:
-                excluded_types:
-                  - 0
-                  - 2
+            card: $home-sets-card
         selection:
           type: Selection.AllSets
     basement:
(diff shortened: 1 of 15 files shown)


ptomato:
Oh, one more thing, I would suggest adding an integration test for variables to the autobahn test suite.

data/preset/A.yaml
```diff
@@ -1,18 +1,53 @@
...
+  set-articles-arrangement:
     type: Arrangement.SquareGuys
     properties:
       halign: center
       valign: start
       expand: true
-    slots:
-      card: *card
+  set-articles-card:
+    type: *card
```
ptomato: It seems to me that customizing this and `search-articles-card` separately would lead to apps with visual inconsistency. Same remark for `set-articles-arrangement` and `search-articles-arrangement`. (In general I'm in favour of using the API to enforce API users to do the right thing, unless there's a good reason why these need to be separated.)

data/preset/escola.yaml
```diff
@@ -1,19 +1,33 @@
 ---
-defines:
+vars:
+  app-banner:
+    shortdef: 'Banner.Dynamic(mode: text, expand: true)'
+  app-brand:
+    shortdef: 'ContentGroup.StaticText(label: "TV ESCOLA")'
```
ptomato: Oh, I had forgotten about this. Perfect use case for variables! It would be even better to make this empty by default, and put "TV ESCOLA" in the app.yaml files.

data/preset/A.yaml
tchx84: Let me double check if I understand properly, because I don't think I agree:

> It seems to me that customizing this and search-articles-card separately would lead to apps with visual inconsistency. Same remark for set-articles-arrangement and search-articles-arrangement.

If an app developer/designer decided that for its content it would be more appropriate to use Arrangement.List/Card.List for search results here (like we do our self in other presets), I don't see why we should prevent that possibility.

> (In general I'm in favour of using the API to enforce API users to do the right thing, unless there's a good reason why these need to be separated.)

As I see it, this feature is about providing more flexibility and visual diversity to our current and (more importantly) future presets. I don't see how preventing them for doing something like I mentioned above would be the right thing to do.

This pattern allows them to do that, and still allows them to use the same card arrangement everywhere if that is what they design.

Both @katelk28 and I reached this same pattern, so I think we need her thoughts on this as well.

(IMPORTANT) On the technical side, just to double check this is clear for others (if reading this), the example you provided:
```
vars:
  something-something-card: &main-card
    shortdef: 'Card.Default(...)'
  something-else-card: *main-card
```

This does not mean that `something-else-card` will get overridden when `something-something-card`. The `main-card` YAML anchor is simply there to save copy pasting `Card.Default` definitions all over the preset, but in the end, each of these variables are completely independent (just happen to have the same default value).

data/preset/escola.yaml
tchx84: Sure, I will empty the string.

data/preset/news.yaml
```diff
@@ -1,12 +1,58 @@
 ---
+vars:
+  app-banner:
+    shortdef: 'Banner.App(show-subtitle: false)'
+
+  home-highlighted-articles-arrangement:
+    shortdef: 'Arrangement.Piano(expand: true)'
+  home-highlighted-articles-card: &main-card
```
tchx84: Note: add order for this one.

data/preset/news.yaml
```diff
@@ -1,12 +1,58 @@
...
+  home-preview-setcontent-card: *main-card
+
+
+  set-articles-arrangement:
+    type: Arrangement.List
+    properties:
+      row-spacing: 20
+  set-articles-card:
+    shortdef: 'Card.List(hexpand: true)'
+  set-preview-subsetcontent-arrangement:
```
tchx84: NOTE: unified with preview-setcontent.

data/preset/buffet.yaml
```diff
@@ -1,15 +1,77 @@
 ---
+vars:
+  home-highlighted-articles-arrangement:
+    type: Arrangement.Windshield
+  home-highlighted-articles-card:
```
tchx84: NOTE: add order

data/preset/escola.yaml
katelk28: Totally agree! But, I recommend that maybe we make a separate ticket to do a larger audit of what the "default template" settings should be; I feel like there will be more things that we'd like to change as default.

But I guess this is such a small & obvious one, that changing it now wouldn't hurt.

data/preset/buffet.yaml
```diff
@@ -1,15 +1,77 @@
...
+      valign: end
+  home-preview-allsets-card: &allsets-card
+    type: Card.Deck
+  home-preview-allsets-text: &allsets-text
+    type: ContentGroup.StaticText
+    properties:
+      label: !translate "Browse by topic"
+      halign: start
+
+  set-preview-subsetcontent-arrangement:
```
katelk28: @tchx84 Unify with `preview-setcontent` variable, instead of `subsetcontent`

data/preset/buffet.yaml
```diff
@@ -1,15 +1,77 @@
...
+    type: Arrangement.Half
+  set-preview-subsetcontent-card: *main-card
```
katelk28: @tchx84 Unify with `setcontent` instead of `subsetcontent`

PR #1564, opened 2018-01-04
Replace all "as" GVariant properties

ptomato:
Replaces most of the GVariant properties with `G_TYPE_STRV` boxed properties.

https://phabricator.endlessm.com/T19625


xjuan:
CR Pass

PR #1565, opened 2018-01-06
Create XapianQuery from EkncQueryObject

ptomato:
https://phabricator.endlessm.com/T20373


ekncontent/ekncontent/eknc-query-object-private.h
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2018 Endless Mobile, Inc. */
+
+#pragma once
+
+#include <glib.h>
+#include <xapian-glib.h>
+
```
ebassi: You should `#include "eknc-query-object.h"` here, so you can drop the include of the public header

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -1,6 +1,7 @@
 /* Copyright 2016 Endless Mobile, Inc. */
 
 #include "eknc-query-object.h"
+#include "eknc-query-object-private.h"
```
ebassi: The usual pattern is to include the private header, which will then include the public one. This avoids having a header that requires a precise ordering of inclusion.

ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -15,27 +15,11 @@
...
-#define QUERY_PARAM_FILTER_OUT "filterOut"
-#define QUERY_PARAM_FLAGS "flags"
-#define QUERY_PARAM_LANG "lang"
-#define QUERY_PARAM_LIMIT "limit"
-#define QUERY_PARAM_MATCH_ALL "matchAll"
-#define QUERY_PARAM_OFFSET "offset"
-#define QUERY_PARAM_ORDER "order"
-#define QUERY_PARAM_QUERYSTR "q"
-#define QUERY_PARAM_SORT_BY "sortBy"
+#include "eknc-query-object.h"
```
ebassi: You should only include the private header — see below…

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -1072,3 +1006,160 @@ G_GNUC_END_IGNORE_DEPRECATIONS
...
+void
+eknc_query_object_configure_enquire (EkncQueryObject *self,
+                                     XapianEnquire *enquire)
+{
+  int sort_value = eknc_query_object_get_sort_value (self);
+
+  if (sort_value > -1)
+    {
+      EkncQueryObjectOrder order;
+      g_object_get (self, "order", &order, NULL);
```
ebassi: Why are you using `g_object_get()`, here?

tools/kermit.js
```diff
@@ -457,10 +457,10 @@ function _pretty_print_table (elements) {
     rows.forEach(row => print(row));
 }
 
-function query (app_id, query_string) {
+function query (app_id, search_terms) {
     let engine = Eknc.Engine.get_default();
     let query_obj = new Eknc.QueryObject({
-        query: query_string,
+        search_terms,
```
ebassi: This is missing the property name.

ptomato: This is new shiny Javascript for `search_terms: search_terms`, so it should work.

ekncontent/ekncontent/eknc-query-object.c
ptomato: There wasn't an accessor method yet, though I could just as well make one.

tools/kermit.js
ebassi: Oooh, nice

ekncontent/ekncontent/eknc-query-object.c
ebassi: You can access `self->order` directly without the GValue boxing/unboxing around it

ekncontent/ekncontent/eknc-query-object-private.h
ptomato: Fixed.

ptomato:
The PR failed because it needs the newer version of xapian-glib, so testing it in flapjack would be more accurate.

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -541,81 +554,86 @@ get_title_clause (EkncQueryObject *self,
...
       const gchar *hash = eknc_utils_id_get_hash (ids[i]);
       if (hash == NULL)
         {
           g_critical ("Unexpected id structure in query object: %s", ids[i]);
-          prefixed_ids[i] = NULL;
-        }
-      else
-        {
-          prefixed_ids[i] = g_strdup_printf ("%s%s", XAPIAN_PREFIX_ID, hash);
+          continue;
```
ojwb: It looks to me like if we hit a NULL hash we'll enter an infinite loop.

I think you need a separate index for `ids` and `prefixed_ids`.

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -510,24 +507,40 @@ get_terms (const gchar *query)
...
+  g_autofree gchar *prefixed = g_strconcat (QUERY_PARSER_PREFIX_EXACT_TITLE, joined, NULL);
 
   /* Combine the term with a wild-carded version if search mode is incremental */
-  return g_strdup_printf ("(%s%s%s*)", prefixed, XAPIAN_OP_OR, prefixed);
+  XapianQueryParserFeature flags = XAPIAN_QUERY_PARSER_FEATURE_DEFAULT;
+  if (self->mode == EKNC_QUERY_OBJECT_MODE_INCREMENTAL)
+    flags |= XAPIAN_QUERY_PARSER_FEATURE_PARTIAL;
+
+  g_autoptr(XapianQuery) retval =
+    xapian_query_parser_parse_query_full (qp, prefixed, flags, "", &error);
```
ojwb: If I follow the code, you could avoid having to create `prefixed` if you specify a default prefix like this:

```
  xapian_query_parser_parse_query_full (qp, joined, flags, XAPIAN_PREFIX_EXACT_TITLE, &error);
```

Note this would need `XAPIAN_PREFIX_EXACT_TITLE` not `QUERY_PARSER_PREFIX_EXACT_TITLE`.

ptomato:
Thanks for the extra review! I opened https://github.com/endlessm/eos-knowledge-lib/pull/1573 to address your comments.

PR #1566, opened 2018-01-10
Various EknContent improvements

ptomato:
Here are a bunch of fixes that can land independently of the main EkncQueryObject issue. This is mostly the removal of dead code plus a big renaming of user-entered search terms from "query" to "search terms" in the code and API.

https://phabricator.endlessm.com/T20373


ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -461,47 +459,12 @@ static XapianMSet *
 eknc_database_manager_fetch_results (EkncDatabaseManager *self,
                                      XapianEnquire *enquire,
                                      XapianQuery *query,
-                                     GHashTable *query_options,
+                                     unsigned offset,
```
ebassi: Can you use `guint`? I don't particularly like `unsigned` by itself.

ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -951,15 +910,50 @@ eknc_database_manager_query_internal (EkncDatabaseManager *self,
...
+                           EKNC_DATABASE_MANAGER_ERROR_INVALID_PARAMS,
+                           "Limit parameter is required for the query");
+      return NULL;
+    }
+
+  /* str may contain a negative value to mean "all matching results"; since
+   * casting a negative floating point value into an unsigned integer is
+   * undefined behavior, we need to perform some level of validation first
+   */
+  {
```
ebassi: It's not really necessary to use a block, here: we're C99-only, we can place declarations after statements.

ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -951,15 +910,50 @@ eknc_database_manager_query_internal (EkncDatabaseManager *self,
         xapian_enquire_set_cutoff (enquire, CLAMP (g_ascii_strtod (cutoff_str, NULL), 0, G_MAXUINT));
     }
 
+  unsigned limit, offset;
```
ebassi: These should be `guint`.

PR #1568, opened 2018-01-12
Remove nav-content from news, blog and buffet presets

xjuan:
https://phabricator.endlessm.com/T20371


data/preset/buffet.yaml
```diff
@@ -307,22 +307,6 @@ defines:
             show-titles: false
             show-toc: false
             custom-css: buffet-custom.css
-        nav-content:
```
ptomato: I don't think we necessarily need to remove this one. It's a lot smaller than the 'see more' popup and doesn't share many of the same problems.

PR #1572, opened 2018-01-16
Add profiling probes to e-k-l

ebassi:
https://phabricator.endlessm.com/T20695

### DIFF
--- a/ekncontent/configure.ac
+++ b/ekncontent/configure.ac
@@ -120,7 +120,7 @@ AC_SUBST([JASMINE_REPORT_ARGUMENT])
 # Package dependencies
 # Update these whenever you use a function that requires a certain API version
 AX_PKG_CHECK_MODULES([EKN_CONTENT], [glib-2.0 gio-2.0 gobject-2.0 json-glib-1.0 libsoup-2.4],
-    [eos-shard-0 >= 0.1.2 xapian-glib-1.0 >= 1.7])
+    [eos-shard-0 >= 0.1.2 xapian-glib-1.0 >= 1.7 endless-0 >= 0.5])
 AX_PKG_CHECK_MODULES([EKN_VFS], [], [eos-shard-0 gio-2.0 gmodule-2.0])
 
 # Code Coverage
--- a/ekncontent/ekncontent/eknc-database-manager.c
+++ b/ekncontent/ekncontent/eknc-database-manager.c
@@ -17,6 +17,8 @@
 #include "eknc-database-manager-private.h"
 #include "eknc-query-object-private.h"
 
+#include <endless/endless.h>
+
 #define PREFIX_METADATA_KEY "XbPrefixes"
 #define STOPWORDS_METADATA_KEY "XbStopwords"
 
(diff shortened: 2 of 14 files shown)


ekncontent/configure.ac
```diff
@@ -120,7 +120,7 @@ AC_SUBST([JASMINE_REPORT_ARGUMENT])
 # Package dependencies
 # Update these whenever you use a function that requires a certain API version
 AX_PKG_CHECK_MODULES([EKN_CONTENT], [glib-2.0 gio-2.0 gobject-2.0 json-glib-1.0 libsoup-2.4],
-    [eos-shard-0 >= 0.1.2 xapian-glib-1.0 >= 1.7])
+    [eos-shard-0 >= 0.1.2 xapian-glib-1.0 >= 1.7 endless-0 >= 0.5])
```
tchx84: did we bump the eos-sdk version? what happened to me is that e-k-l was configured "properly" (with the old eos-sdk I had installed, without profile probes)  but then failed while building ekncontent (because of the missing probe API).

ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -314,6 +316,8 @@ create_database_from_manifest (const char  *manifest_path,
 {
   GError *error = NULL;
 
+  g_autoptr(EosProfileProbe) probe = EOS_PROFILE_PROBE ("/ekncontent/database/create");
+
```
tchx84: Like we mentioned on slack, this would only happens once in a regular apps so we a) allow profile probe to display the one sample time or b) removed it (since it doesn't provide any info).

ekncontent/ekncontent/eknc-database-manager.c
```diff
@@ -338,6 +342,9 @@ create_database_from_manifest (const char  *manifest_path,
 
   for (GList *l = dbs; l != NULL; l = l->next)
     {
+      g_autoptr(EosProfileProbe) db_probe =
+        EOS_PROFILE_PROBE ("/ekncontent/database/create/internal_db");
```
tchx84: Same thing here, allow one sample time or remove this probe.

ekncontent/ekncontent/eknc-object-model.c
```diff
@@ -24,6 +25,8 @@ EkncContentObjectModel *
 eknc_object_model_from_json_node (JsonNode *node,
                                   GError **error)
 {
+  g_autoptr(EosProfileProbe) probe = EOS_PROFILE_PROBE ("/ekncontent/object_model");
```
tchx84: nit: maybe there is a better name for probe? To me, `object_model` didn't suggested that it was this helper.. if the idea is to probe the common point for all model objects then maybe something like `/ekncontent/object`, or we want to make it very explicit then just `/ekncontent/object/from_json_node` or something.

js/app/modules/contentGroup/dynamicTitle.js
```diff
@@ -25,6 +26,8 @@ var DynamicTitle = new Module.Class({
     },
 
     make_ready: function (cb = function () {}) {
+        let probe = Endless.ProfileProbe('/ekn/dynamicTitle/make_ready');
```
tchx84: Missing `start`.

ekncontent/configure.ac
ebassi: Ah, that's a good point; I don't think we did.

ekncontent/ekncontent/eknc-database-manager.c
ebassi: eos-sdk#4177 should fix this.

ekncontent/ekncontent/eknc-object-model.c
ebassi: I went with `object/from_json_node`, as it makes it clear where the probe is applied (outside of the location).

PR #1573, opened 2018-01-16
Suggestions from Olly

ptomato:
https://phabricator.endlessm.com/T20757

### DIFF
--- a/ekncontent/ekncontent/eknc-query-object.c
+++ b/ekncontent/ekncontent/eknc-query-object.c
@@ -18,7 +18,6 @@
 #define XAPIAN_PREFIX_ID "Q"
 #define XAPIAN_PREFIX_TAG "K"
 
-#define QUERY_PARSER_PREFIX_EXACT_TITLE "exact_title:"
 #define QUERY_PARSER_PREFIX_TITLE "title:"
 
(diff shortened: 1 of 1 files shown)


ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -619,17 +619,17 @@ get_ids_clause (char **ids,
     return NULL;
 
   g_auto(GStrv) prefixed_ids = g_new0 (gchar *, length + 1);
-  for (size_t i = 0; i < length;)
+  for (size_t ix = 0, prefixed_ix = 0; ix < length; ix++)
     {
-      const gchar *hash = eknc_utils_id_get_hash (ids[i]);
+      const char *hash = eknc_utils_id_get_hash (ids[ix]);
```
ojwb: Is this `gchar` to `char` change intended?

Looks OK to me otherwise.  I can see CI failed, but I can't seem to load the log (with the VPN up, the browser spinds trying to load until it times out).

ptomato: That's intended, we're trying to move away from `gchar` gradually.

The CI is failing because of an unrelated issue (too old version of xapian-glib in the CI environment); hopefully we'll have that sorted out soon.

PR #1579, opened 2018-01-30
T20761: Add "content-type" filter and "wrapper-uri" prop

smspillaz:
This PR adds a "content-type" filter to `EkncQueryObject` and a "wrapper-uri" property to `EkncVideoObjectModel`.

The "wrapper-uri" property indicates on an `EkncVideoObjectModel` which article ID is "thinly wrapping" it if it was tagged with `EknWrappedContent`. Capturing this relationship might be useful if a client needs to go back from some `EkncVideoObjectModel` to its wrapper.

The "content-type" filter on `EkncQueryObject` makes use of the `XCONTENTTYPE` term prefix added to the shard packer as a means to filter results by content type.

It should be noted that both of these changes aren't strictly speaking necessary in order to enable the feature referred to on the ticket, though they were both requested there, so this PR was created for completeness sake.

https://phabricator.endlessm.com/T20761

### DIFF
--- a/ekncontent/ekncontent/eknc-query-object.c
+++ b/ekncontent/ekncontent/eknc-query-object.c
@@ -16,6 +16,7 @@
 #define MAX_TERM_LENGTH 245
 
 #define XAPIAN_PREFIX_EXACT_TITLE "XEXACTS"
+#define XAPIAN_PREFIX_CONTENT_TYPE "T"
 #define XAPIAN_PREFIX_ID "Q"
 #define XAPIAN_PREFIX_TAG "K"
 
@@ -51,6 +52,7 @@ struct _EkncQueryObject
   gchar *corrected_terms;
   gchar *stopword_free_terms;
   gchar *literal_query;
+  gchar *content_type;
   EkncQueryObjectMode mode;
   EkncQueryObjectMatch match;
   EkncQueryObjectSort sort;
@@ -86,6 +88,7 @@ enum {
   PROP_EXCLUDED_IDS,
   PROP_EXCLUDED_TAGS,
   PROP_CORRECTED_TERMS,
+  PROP_CONTENT_TYPE,
   NPROPS
 };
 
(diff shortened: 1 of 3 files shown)


ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -16,6 +16,7 @@
 #define MAX_TERM_LENGTH 245
 
 #define XAPIAN_PREFIX_EXACT_TITLE "XEXACTS"
+#define XAPIAN_PREFIX_CONTENT_TYPE "XCONTENTTYPE"
```
cosimoc: I believe this will have to be changed to `T` -- see the commend in the other PR

cosimoc:
A couple of other thoughts:
- it would be nice to have some tests for these new properties
- perhaps we should also add a `wraps` property as well

smspillaz:
> it would be nice to have some tests for these new properties

I'll do this soon.

> perhaps we should also add a wraps property as well

IIRC @ptomato didn't want to have such a property, only the property on the child pointing to its parent wrapper (but it is useful to have it in the shard metadata nonetheless and mandatory for libingester, since we need to know if an article was just a thin wrapper for something else).

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -51,6 +52,7 @@ struct _EkncQueryObject
   gchar *corrected_terms;
   gchar *stopword_free_terms;
   gchar *literal_query;
+  gchar *content_type;
```
ptomato: nitpick: Prefer `char` in new code.

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -827,6 +866,27 @@ get_corrected_query (EkncQueryObject *self,
...
+ */
+static XapianQuery *
+get_content_type_clause (const gchar *content_type)
+{
+  g_autofree gchar *prefixed = NULL;
+
+  if (content_type == NULL)
+    return NULL;
+
+  prefixed = g_strconcat (XAPIAN_PREFIX_CONTENT_TYPE, content_type, NULL);
```
ptomato: nitpick: Move `g_autofree char *` in front of this line.

smspillaz:
Closing for now. I'm going to separate this out into separate PRs to have the `content-type` indexing and the `parent` property on embedded objects.

PR #1582, opened 2018-02-02
T14289

tchx84:
https://phabricator.endlessm.com/T14289

### DIFF
--- a/tests/autobahn/overrides.yaml
+++ b/tests/autobahn/overrides.yaml
@@ -1,0 +1,25 @@
+---
+overrides:
+  search-module:
+    shortdef: 'Navigation.Overridden(fully-overridden: false)'
+  root:
+    shortdef: 'Controller.Overridden(added: true)'
+  root.window.contents.1:
+    type: Layout.Overridden
+  root.window.search:
+    properties:
+      fully-overridden: true
+---
+vars:
+  search-module:
+    type: Navigation.Original
+root:
+  type: Controller.Original
+  slots:
+    window:
+      type: Window.Original
+      slots:
+        contents:
+        - shortdef: Layout.Original
+        - shortdef: Layout.Original
+        search: $search-module
--- a/tools/autobahn
+++ b/tools/autobahn
@@ -70,7 +70,9 @@ def recursive_ordered_dict(regdict):
 
 
 def order_module_keys(module):
-    retval = collections.OrderedDict({'type': module['type']})
+    retval = collections.OrderedDict()
+    if 'type' in module:
+        retval['type'] = module['type']
     if 'id' in module:
         retval['id'] = module['id']
     if 'styles' in module:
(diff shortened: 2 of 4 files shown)


tools/autobahn
```diff
@@ -104,6 +106,34 @@ def expand_refvar(module):
...
 
 
+def expand_overrides(overrides):
+    for key in overrides:
+        overrides[key] = walk(overrides[key], expand_slot_shorthand)
+        overrides[key] = walk(overrides[key], expand_shortdef)
+    return overrides
+
+
+def process_custom_overrides(tree):
```
ptomato: It would be better to pass `overrides` to this function, I think.

tests/autobahn/overrides.yaml
```diff
@@ -0,0 +1,25 @@
+---
+overrides:
+  search-module:
+    shortdef: 'Navigation.Overriden(fully-overriden: false)'
```
ptomato: s/overriden/overridden/g

tests/autobahn/overrides.yaml
```diff
@@ -0,0 +1,25 @@
...
+  root:
+    shortdef: 'Controller.Overriden(added: true)'
+  root.window.contents.1:
+    type: Layout.Overriden
+  root.window.search:
+    properties:
+      fully-overriden: true
```
ptomato: I was a bit confused by this at first, as I thought you were introducing a new property that you had to set if you wanted to fully override a module. Maybe it should be called something like `override-sample-property`?

tools/autobahn
```diff
@@ -104,6 +106,34 @@ def expand_refvar(module):
...
+    components = name.split('.')
+    for key in components[1:]:
+        try:
+            if isinstance(node, list):
+                node = node[int(key)]
+            else:
+                node = node['slots'][key]
+        except:
+            return
+    node.update(overrides[name])
```
ptomato: Isn't this going to keep the existing contents of `node` as well as the overridden contents, if they have different keys?

tchx84: Yes, but that is the idea, if you want to fully replace the node, then you need to fully override, but as it is, this allows you to only update certain parts like, e.g. only the properties.

tools/autobahn
tchx84: overrides is globally accessible, but I think is fine either way.

tests/autobahn/overrides.yaml
tchx84: The meaning behind `fully` name is more subtle than I expected then, notice that :
1) the search node references a variable.
2) but the variable is overridden, in the overrides section.
3) but the overridden version has that property set to false, because
4) we also have a custom override for that property.

the reason for this is to include in the test the order in which these are processed.

1) variables replacement and overridden, and then
2) custom overrides applied on top of the final result.

ptomato: Well, maybe still put `example` or `sample` in there? I was confused whether it was part of the overriding API or not.

PR #1584, opened 2018-02-02
WebShareDialog: handle expired sessions

xjuan:
Open control center if account AttentionNeeded is true

https://phabricator.endlessm.com/T20960

### DIFF
--- a/data/widgets/websharedialog.ui
+++ b/data/widgets/websharedialog.ui
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.20.0 -->
+<!-- Generated with glade 3.21.0 -->
 <interface>
   <requires lib="gtk+" version="3.20"/>
   <requires lib="webkit2gtk" version="2.12"/>
   <template class="Gjs_WebShareDialog" parent="GtkDialog">
+    <property name="width_request">320</property>
+    <property name="height_request">320</property>
     <property name="can_focus">False</property>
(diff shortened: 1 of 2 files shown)


xjuan:
You can unpak seahorse from a debian package and run it locally by compiling the schemas and setting GSETTINGS_SCHEMA_DIR env variable

Here is a tarball with the compiled schemas [seahorse.zip](https://github.com/endlessm/eos-knowledge-lib/files/1725271/seahorse.zip)

GSETTINGS_SCHEMA_DIR=$HOME/seahorse/usr/share/glib-2.0/schemas/ seahorse/usr/bin/seahorse

ptomato:
I still have this happening:
> But what I do see when I run com.endlessm.encyclopedia.en and try to share on Facebook, is that a blank window opens up and nothing happens.

I double-checked on SDK3 and it doesn't happen there.

xjuan:
I think thats probably because I never added error handling, so if the internet connection fails you will get stuck with an blank window

Added https://phabricator.endlessm.com/T21193

ptomato:
Maybe someone else should review this while the load-failed issue is being worked on? FWIW, I don't get the blank window on master.

nedrichards:
Good point! We really need a better way of expressing capabilities into flatpak permissions.

nedrichards:
Does it ever talk more specifically to the ControlCenter? you might also need a .* permission underneath that.

xjuan:
No, it only activates a GtkAction from org.gnome.ControlCenter

PR #1587, opened 2018-02-09
ekncontent: Add "parent-uri" prop to EkncMediaObjectModel

smspillaz:
This is the EKN URI for the article which embeds this Media Object,
if it has an embedder and the embedder is known (otherwise it is
an empty string).

https://phabricator.endlessm.com/T21114

### DIFF
--- a/ekncontent/ekncontent/eknc-media-object-model.c
+++ b/ekncontent/ekncontent/eknc-media-object-model.c
@@ -21,6 +21,7 @@ typedef struct {
   gchar *caption;
   guint width;
   guint height;
+  char *parent_uri;
 } EkncMediaObjectModelPrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (EkncMediaObjectModel,
@@ -32,6 +33,7 @@ enum {
   PROP_CAPTION,
   PROP_WIDTH,
   PROP_HEIGHT,
+  PROP_PARENT_URI,
   NPROPS
 };
 
@@ -58,6 +60,10 @@ eknc_media_object_model_get_property (GObject    *object,
 
     case PROP_HEIGHT:
       g_value_set_uint (value, priv->height);
+      break;
+
+    case PROP_PARENT_URI:
+      g_value_set_string (value, priv->parent_uri);
       break;
 
     default:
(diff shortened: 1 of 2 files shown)


ekncontent/ekncontent/eknc-media-object-model.c
```diff
@@ -146,6 +157,16 @@ eknc_media_object_model_class_init (EkncMediaObjectModelClass *klass)
...
+  /**
+   * EkncMediaObjectModel:parent-uri:
+   *
+   * The EKN URI of an #EkncContentObjectModel that embeds this
+   * #EkncMediaObjectModel.
+   */
+  eknc_media_object_model_props[PROP_PARENT_URI] =
+    g_param_spec_string ("parent-uri", "Parent URI",
+      "EKN URI of article that embeds this media object",
+      "", G_PARAM_READWRITE | G_PARAM_CONSTRUCT_ONLY | G_PARAM_STATIC_STRINGS);
```
ptomato: Default value should be `NULL` instead of an empty string.

ekncontent/tests/ekncontent/testMediaObjectModel.js
```diff
@@ -37,6 +38,7 @@ describe('Media Object Model', function () {
             expect(imageObject.caption).toBe('Great musician, or greatest?');
             expect(imageObject.width).toBe(666);
             expect(imageObject.height).toBe(666);
+            expect(imageObject.parent_uri).toBe(MOCK_MEDIA_DATA.parent)
```
cosimoc: Missing semicolon

smspillaz: *cough* fixed.

cosimoc:
LGTM -- I'll leave it unmerged for @ptomato to have a second look.

ekncontent/ekncontent/eknc-media-object-model.c
```diff
@@ -89,6 +95,10 @@ eknc_media_object_model_set_property (GObject *object,
       priv->height = g_value_get_uint (value);
       break;
 
+    case PROP_PARENT_URI:
+      priv->parent_uri = g_value_dup_string (value);
```
ptomato: Not strictly necessary since it's construct-only, but it would be good practice to have `g_clear_pointer(&priv->parent_uri, g_free);` here.

smspillaz: Good point, lets do that.

PR #1588, opened 2018-02-14
T21128

tchx84:
https://phabricator.endlessm.com/T21128


js/app/modules/filter/type.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Class: Type
+ * Filter that includes only a content type
+ *
+ * This filter includes only object models of the content type
+ * specified in its `type` string property.
+ *
+ * The type property accepts a trailing "*" wildcard.
+ *
+ * Using the "invert" property has no effect on this Filter.
```
smspillaz: I see - I guess this is the reason why we wanted to have a `not_content_type` filter. In that case, I think it'd be worth implementing that as well for the sake of not breaking interface contract - I am happy to volunteer.

tests/js/app/modules/filter/testType.js
```diff
@@ -0,0 +1,51 @@
+// Copyright 2018 Endless Mobile, Inc.
+
+const Eknc = imports.gi.EosKnowledgeContent;
+
+const Type = imports.app.modules.filter.type;
+const MockFactory = imports.tests.mockFactory;
+
+describe('Filter.Type', function () {
+    const MODELS = ['text/html', 	'video/jpeg'].map(type =>
```
smspillaz: Extra tab here

tests/js/app/modules/filter/testType.js
```diff
@@ -0,0 +1,51 @@
...
+            content_type: type,
+        }));
+    let filter;
+
+    describe('normal mode', function () {
+        beforeEach(function () {
+            [filter] = MockFactory.setup_tree({
+                type: Type.Type,
+                properties: {
+                    type: 'video/jpeg',
```
smspillaz: No such thing as a `video/jpeg` ;-)

tests/js/app/modules/filter/testType.js
```diff
@@ -0,0 +1,51 @@
...
+                    invert: true,
+                },
+            });
+        });
+
+        it('filters out a non-video models (behavior unaltered)', function () {
+            expect(filter.include(MODELS[0])).toBeFalsy();
+            expect(filter.include(MODELS[1])).toBeTruthy();
+        });
```
smspillaz: Perhaps it'd be better to have an `xit` test here to test the expected behaviour once the `invert` property works correctly.

ekncontent/ekncontent/eknc-query-object.c
```diff
@@ -879,12 +879,18 @@ static XapianQuery *
...
   g_autofree char *prefixed = NULL;
+  XapianQuery *retval = NULL;
 
   if (content_type == NULL)
     return NULL;
 
   prefixed = g_strconcat (XAPIAN_PREFIX_CONTENT_TYPE, content_type, NULL);
-  return xapian_query_new_for_term (prefixed);
+  if (g_str_has_suffix (content_type, "*"))
+      retval = xapian_query_new_wildcard (g_strstrip (g_strdelimit (prefixed, "*", ' ')));
```
smspillaz: This will convert all `*` in the string into `' '`. Probably an edge case to get something like `foo*/bar*`, but perhaps it'd be better to explicitly chop characters off the end using a separate helper function.

tchx84: IIRC Xapian only supports trailing wildcards, not possible to do something like `foo*/bar*`.

js/app/modules/filter/type.js
tchx84: Yeah, it should be trivial, let me discuss it in SDK standup (in a some minutes), it might just do it myself if that is the case.

tests/js/app/modules/filter/testType.js
tchx84: ops!

tests/js/app/modules/filter/testType.js
tchx84: It could have existed https://www.ietf.org/rfc/rfc3555.txt xD

tests/js/app/modules/filter/testType.js
tchx84: Well, if we end-up adding `excluded_content_type` (like we mentioned above), I will change it anyways.

js/app/modules/filter/type.js
```diff
@@ -0,0 +1,56 @@
...
+    Name: 'Filter.Type',
+    Extends: GObject.Object,
+    Implements: [Filter.Filter],
+
+    Properties: {
+        'type': GObject.ParamSpec.string('type', 'Type', 'Type to include',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY, ''),
+    },
+
+    _init: function (props={}) {
```
ptomato: Coding style nitpick: Use new method notation
```js
_init(props={}) {
    ...
},
```

ptomato:
Also, may I bikeshed the name? I like `Filter.ContentType` being more specific.

PR #1599, opened 2018-03-01
ContentGroup.StaticText: expand and align by default

ptomato:
If you don't do this, the title is off-center in a weird way. You can
still customize these properties, but they should have good default
values.

https://phabricator.endlessm.com/T21365


tchx84:
screenshot from 2018-03-13 10-33-25

PR #1601, opened 2018-03-06
T21276

tchx84:
https://phabricator.endlessm.com/T21276


tchx84:
@ptomato Tested the two things we discussed:

1) Size difference between having and not having this value number. Tested with a 1GB shard, a) without the shard size is 1109209172B, while b) with it the shard size is 1109217364B, almost unnoticeable.

2) The behavior with a shard that doesn't include the number value is just the same as before in this e-k-l branch.

PR #1602, opened 2018-03-06
T21272: Port articleHTMLRenderer to C

smspillaz:
This PR creates a new sub-library called `eknrenderer` which is essentially a straight port of `articleHTMLRenderer` to C with no dependencies on `ekncontent`. `articleHTMLRenderer` now uses `eknrenderer` in order to do it its work.

We use https://github.com/x86-64/mustache-c to render the templates and most of the logic that articleHTMLRenderer had is now reimplemented in `EknrRenderer`.

The fact that it is a straight port makes the API a little awkward, in particular, there are functions have have upwards of ten arguments because we can't depend on the models directly. Ideally we'll want to have this as a separate library that doesn't even exist within `knowledge-lib`, but for now the best way to do it is to make it an internal library, make sure the apps work and then look at making it a separate library that gets included in the runtime.

This was tested against the existing tests for `articleHTMLRenderer` and checked with valgrind - apart from https://github.com/x86-64/mustache-c/pull/12 there are no leaks.

https://phabricator.endlessm.com/T21272


eknrenderer/configure.ac
```diff
@@ -0,0 +1,147 @@
...
+    [Use mathjax from a location besides the default /usr/share]),
+    [MATHJAXDIR=$withval],
+    [MATHJAXDIR=/usr/share/javascript/mathjax])
+AC_MSG_CHECKING([for mathjax javsacript library])
+if test -d $MATHJAXDIR; then
+    AC_MSG_RESULT([yes])
+    AC_SUBST([MATHJAXDIR])
+else
+    AC_MSG_ERROR([Cannot find mathjax package])
+fi
```
smspillaz: Having to do this again in the internal lib's `configure.ac` seems really suboptimal to me, but I couldn't figure out a way to get the substitution to propagate from the parent.

eknrenderer/Makefile.am
```diff
@@ -0,0 +1,143 @@
...
+	eknrenderer/eknr-errors.c \
+	eknrenderer/eknr-renderer.c \
+	$(NULL)
+
+glib_enum_headers = \
+	$(eknrenderer_public_installed_headers) \
+	$(NULL)
+glib_enum_h = eknrenderer/eknr-enums.h
+glib_enum_c = eknrenderer/eknr-enums.c
+include $(top_srcdir)/Makefile.am.enums
```
smspillaz: It would also be nice if there was away to include the top-level `Makefile.am.enums` here as opposed to the sub library's one.

eknrenderer/Makefile.am
```diff
@@ -0,0 +1,143 @@
...
+	$(eknrenderer_public_installed_headers) \
+	$(eknrenderer_sources) \
+	$(NULL)
+
+libeos_knowledge_renderer_@EKN_RENDERER_API_VERSION@_la_CPPFLAGS = \
+	@EKN_RENDERER_CFLAGS@ \
+	$(CODE_COVERAGE_CPPFLAGS) \
+	-DG_LOG_DOMAIN=\"EosKnowledgeRenderer\" \
+	-DCOMPILING_EKN_RENDERER \
+	-DMATHJAX_PATH=\"@MATHJAXDIR@\"
```
smspillaz: ekncontent uses a `config.h.in` for this. I couldn't get the variable substitution there to work right, so I've just opted for doing this for now.

eknrenderer/eknrenderer/eknr-boxed-set-info.c
```diff
@@ -0,0 +1,70 @@
+/* Copyright 2018 Endless Mobile, Inc. */
```
smspillaz: Probably needs an actual copyright header, but I'm just following the example in ekncontent for now.

eknrenderer/eknrenderer/eknr-boxed-set-info.c
```diff
@@ -0,0 +1,70 @@
...
+
+/* Boxed type registration for EknrBoxedSetInfo */
+GType
+eknr_boxed_set_info_get_type (void)
+{
+  static GType type_id = 0;
+
+  if (!type_id)
+    type_id = g_boxed_type_register_static (g_intern_static_string ("EknrBoxedSetInfo"),
+                                            (GBoxedCopyFunc) eknr_boxed_set_info_copy,
```
smspillaz: Not sure if we actually want this to be `eknr_boxed_set_info_ref` - I am not sure if the semantics of `GBoxedCopyFunc` require that a deep copy be made or if a shallow copy is permissible.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+#include <mustache.h>
+#include <libintl.h>
+
+#include "eknr-boxed-set-info.h"
+#include "eknr-errors.h"
+#include "eknr-renderer.h"
+
+#define _(String) gettext (String)
+#define gettext_noop(String) String
+#define N_(String) gettext_noop (String)
```
smspillaz: There has to be a better way to do this, but I didn't immediately see one.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+{
+  mustache_api_t api = {
+    .freedata = NULL
+  };
+
+  mustache_free (&api, template);
+}
+
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (mustache_template_t, free_mustache_template);
```
smspillaz: This extra newline here can probably be removed.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+
+static RendererMustacheData *
+renderer_mustache_data_new (GVariantDict  *variables,
+                            GError       **error,
+                            const gchar   *input)
+{
+  RendererMustacheData *data = g_new0 (RendererMustacheData, 1);
+  data->variables = variables;
+  data->error = error;
+  data->input.string = (gchar *) input;
```
smspillaz: I'm not sure if `RendererMustacheData` should take a copy of the input string here - I suppose it could, though the way that we use this at the moment means does not cause any lifetime issues.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+  RendererMustacheData *data = userdata;
+  return mustache_std_strwrite (api, (void *) &data->output, (char *) buffer, buffer_size);
+}
+
+static gchar *
+_lookup_in_gvariant_dict (RendererMustacheData *data,
+                          const gchar          *text,
+                          mustache_api_t       *api)
+{
+  g_autoptr(GVariant) value_v = g_variant_dict_lookup_value (data->variables, text, G_VARIANT_TYPE ("s"));
```
smspillaz: We require that all variables right now be strings. This doesn't seem to cause any problems for our templates, but still a rather odd limitation.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+    case '"':
+      return "&quot;";
+    case '\'':
+      return "&#39;";
+    case '/':
+      return "&#x2F;";
+    default:
+      return NULL;
+  }
+}
```
smspillaz: There's probably a g* function for this somewhere that I missed.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+
+static gboolean
+escape_value_eval_callback (const GMatchInfo *match_info,
+                            GString          *result,
+                            gpointer          user_data)
+{
+  /* Faster to just use if statements here */
```
smspillaz: This comment is now out of date - I'll remove it.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+  g_autofree gchar *matched = g_match_info_fetch (match_info, 0);
+  char matched_ent = matched[0];
+  const gchar *html_entity = html_entity_for_char (matched_ent);
+
+  if (html_entity == NULL)
+    return TRUE;
+
+  g_string_append (result, html_entity);
+
+  /* From the documentation, FALSE to continue to the replacement process,
```
smspillaz: *we should return FALSE

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+    {
+      (*api->error) (api, api_closure, __LINE__, error->message);
+      return NULL;
+    }
+
+  return g_steal_pointer (&escaped);
+}
+
+static uintmax_t
+_renderer_var_from_ht (mustache_api_t            *api,
```
smspillaz: These `from_ht` functions should probably be renamed to `from_vardict`, since we're not using hash table anymore.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+  RendererMustacheData *data = userdata;
+  g_autofree const gchar **strv = g_variant_get_strv (variant, NULL);
+  const gchar **iter = strv;
+
+  /* Keep track of the current section variable, then pop it once we're
+   * done */
+  const gchar *last_section_variable = data->section_variable;
+
+  /* Need to render each sub-template from here */
+  while (*iter != NULL)
```
smspillaz: Maybe we want to promote the `for (iter = strv; *iter != NULL; ++iter)` style here...

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+               "Failed to perform template substitution: %s",
+               msg);
+}
+
+gchar *
+_renderer_render_mustache_document_internal (mustache_template_t  *tmpl,
+                                             GVariant             *variables,
+                                             GError              **error)
+{
+  mustache_api_t api = {
```
smspillaz: We could make this static

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+  EknrRendererPrivate *priv = eknr_renderer_get_instance_private (renderer);
+  mustache_template_t *tmpl = g_hash_table_lookup (priv->cache, path);
+  g_autofree gchar *contents = NULL;
+  mustache_api_t api = {
+    .read = &_renderer_read_from_closure,
+    .write = &_renderer_write_to_closure,
+    .varget = &_renderer_var_from_ht,
+    .sectget = &_renderer_sect_from_ht,
+    .error = &_renderer_set_error
+  };
```
smspillaz: I note that this is the same vtable as above, so it could even be a compilation-unit-global...

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+eknr_mustache_template_compiles (const gchar  *tmpl_text,
+                                 GError      **error)
+{
+  mustache_api_t api = {
+    .read = &_renderer_read_from_closure,
+    .write = &_renderer_write_to_closure,
+    .varget = &_renderer_var_from_ht,
+    .sectget = &_renderer_sect_from_ht,
+    .error = &_renderer_set_error
+  };
```
smspillaz: .... yeah this should definitely be a compilation unit global...

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+
+  return TRUE;
+}
+
+static gchar *
+format_a_href_link (const gchar  *uri,
+                    const gchar  *text,
+                    GError      **error)
+{
+  g_autofree gchar *escaped = eknr_escape_html(text, error);
```
smspillaz: Space

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+        return NULL;
+
+      wikihow_link = format_a_href_link (_("http://wikihow.com"),
+                                         source_name,
+                                         error);
+
+      if (wikihow_link == NULL)
+        return NULL;
+
+      disclaimer = g_strdup_printf(_("See %s for more details, videos, pictures and attribution. Courtesy of %s, where anyone can easily learn how to do anything."),
```
smspillaz: Frustratingly, these will require re-translation

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+
+  /* NULL-terminate */
+  g_ptr_array_add (javascript_files, NULL);
+
+  strv = (GStrv) g_ptr_array_free (javascript_files, FALSE);
+  return g_variant_new_strv ((const gchar * const *) strv, -1);
+}
+
+static GVariant *
+get_legacy_should_include_mathjax (const gchar *source)
```
smspillaz: I am a little torn as to whether these helper functions should be returning a native value or `GVariant *` directly. It is slightly more convenient if it is the latter, even if it couples it to the caller somewhat.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1408 @@
...
+                              const gchar                        *source,
+                              const gchar                        *source_name,
+                              const gchar                        *original_uri,
+                              GStrv                               tags,
+                              GHashTable                         *set_map,
+                              GStrv                               outgoing_links,
+                              EknrOutgoingLinkDeterminationFunc   determination_func,
+                              gpointer                            user_data,
+                              GStrv                               custom_css_files,
+                              GStrv                               override_css_files,
```
smspillaz: Having to have `GStrv` here in the arguments and then cast it away later on in the function is *really* annoying, but I couldn't get gi-scanner to co-operate with me otherwise (its was not treating `const char * const *` as an array, but it works fine with `GStrv`).

smspillaz:
To clarify a little - since the overall goal here encompasses a bit more scope, here's what I consider to be not-in-scope for this PR:
 - Separating out eknrenderer into a separate library (we can do this in a follow-up change once we have decided how eknrenderer is going to look)
 - Anything that is mobile-device friendly: I want to make sure this port works first, so that can be done in a follow-up
 - Large adjustments to the interface to suit the fact that this code is now C

smspillaz:
> It makes no difference for testing the code and deciding how the library is going to look — if you want to iterate on the API, you'll have to make the same amount of changes in articleHTMLRenderer, regardless of whether the library is embedded or not.

That's fair enough. Good news is that now the internal library is already written, its actually not a whole lot of work to split it out into an external one.

Of course, it would increase review complexity somewhat, since there would be two interdependent PRs (one to remove the old rendering code and another one to add it somewhere else). But as long as you're okay with that, we can go down that road.

My idea in keeping it internal was to avoid the churn associated with having two concurrent PRs - it was definitely my intention to split it out ASAP. But again, if this is something we're okay living with, I'm more than happy to split it out now.

smspillaz:
> (That's what I'm actually planning to do this week for EknContent, split it out and port it to Meson...)

I was actually thinking just now - if you were planning to do this, I might just wait for you to do that and then follow your example. I'm not overly familiar with Meson, so it'd be good to have a precedent for what the preferred setup is.

eknrenderer/eknrenderer/eknr-boxed-set-info.c
cosimoc: It does not require that a deep copy is made; feel free to change it to that.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+  NPROPS
+};
+
+static GParamSpec *eknr_renderer_props [NPROPS] = { NULL, };
+
+static mustache_str_ctx *
+_renderer_mustache_str_ctx_new (char *string)
+{
+  mustache_str_ctx *ctx = g_new0 (mustache_str_ctx, 1);
+  ctx->string = string;
```
cosimoc: Don't you want to copy the string here?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+  const char *section_variable;
+} RendererMustacheData;
+
+static RendererMustacheData *
+renderer_mustache_data_new (GVariantDict  *variables,
+                            GError       **error,
+                            const char    *input)
+{
+  RendererMustacheData *data = g_new0 (RendererMustacheData, 1);
+  data->variables = variables;
```
cosimoc: Should probably ref `variables`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+  data->error = error;
+  data->input.string = (char *) input;
```
cosimoc: Should copy `input`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+  data->output.offset = 0;
+
+  return data;
+}
+
+static void
+renderer_mustache_data_free (RendererMustacheData *data)
+{
+  g_clear_pointer (&data->variables, g_variant_dict_unref);
+  g_clear_pointer (&data->output.string, free);
```
cosimoc: What about `data->input`?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+free_mustache_template(mustache_template_t *template)
+{
+  mustache_api_t api = {
+    .freedata = NULL
+  };
+
+  mustache_free (&api, template);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (mustache_template_t, free_mustache_template);
```
cosimoc: Remove the terminal semicolon

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+  g_clear_pointer (&data->variables, g_variant_dict_unref);
+  g_clear_pointer (&data->output.string, free);
+  g_free (data);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (RendererMustacheData,
+                               renderer_mustache_data_free);
+
+static void
+free_mustache_template(mustache_template_t *template)
```
cosimoc: Add space before paren

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+                                 const char   *source,
+                                 const char   *source_name,
+                                 const char   *original_uri,
+                                 const char   *license,
+                                 const char   *title,
+                                 gboolean      show_title,
+                                 gboolean      use_scroll_manager,
+                                 GError      **error)
+{
+  g_autoptr(GFile) file = template_file("legacy-article.mst");
```
cosimoc: Missing space before paren

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+                                                      error);
+
+  if (disclaimer == NULL)
+    return NULL;
+
+  g_variant_dict_init (&vardict, NULL);
+
+  g_variant_dict_insert_value (&vardict,
+                               "title",
+                               show_title ? g_variant_new_string (title) : g_variant_new_boolean (FALSE));
```
cosimoc: You probably want an empty string in the false case, no?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+                                                           variant,
+                                                           error);
+}
+
+char *
+_renderer_render_prensa_libre_content (EknrRenderer *renderer,
+                                       const char   *body_html,
+                                       GError      **error)
+{
+  g_autoptr(GFile) file = template_file("news-article.mst");
```
cosimoc: Missing space before paren

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+
+  switch (prop_id)
+    {
+    default:
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
+    }
+}
+
+static void
+eknr_renderer_set_property (GObject *object,
```
cosimoc: Remove empty setter

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+  variant = g_variant_dict_end (&vardict);
+
+  return eknr_renderer_render_mustache_document_from_file (renderer,
+                                                           file,
+                                                           variant,
+                                                           error);
+}
+
+static void
+eknr_renderer_get_property (GObject    *object,
```
cosimoc: Remove empty getter

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+                              GStrv                               tags,
+                              GHashTable                         *set_map,
+                              GStrv                               outgoing_links,
+                              EknrOutgoingLinkDeterminationFunc   determination_func,
+                              gpointer                            user_data,
+                              GStrv                               custom_css_files,
+                              GStrv                               override_css_files,
+                              GError                            **error)
+{
+  g_autoptr(GFile) file = template_file("article-wrapper.mst");
```
cosimoc: Missing space before paren

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+static GVariant *
+get_wrapper_share_actions_markup (const char *original_uri,
+                                  GError    **error)
+{
+  g_autofree char *facebook = NULL;
+  g_autofree char *twitter = NULL;
+  g_autofree char *whatsapp = NULL;
+  g_autofree char *markup = NULL;
+
+  if (original_uri == NULL || g_strcmp0 (original_uri, "") == 0)
```
cosimoc: Instead, do `(original_uri == NULL || original_uri[0] == '\0')`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+}
+
+static GVariant *
+get_wrapper_css_files (const char * const *custom_css_files)
+{
+  GPtrArray *wrapper_css_files = g_ptr_array_new_with_free_func (g_free);
+  g_auto(GStrv) strv = NULL;
+
+  g_ptr_array_add (wrapper_css_files, g_strdup ("clipboard.css"));
+  g_ptr_array_add (wrapper_css_files, g_strdup ("share-actions.css"));
```
cosimoc: You don't really need to copy all these strings, do you? `g_variant_new_strv()` will make a copy.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+                              GError       **error)
+{
+  if (server_templated)
+    return g_strdup (body_html);
+
+  if (g_strcmp0 (source, "wikipedia") == 0 ||
+      g_strcmp0 (source, "wikihow") == 0 ||
+      g_strcmp0 (source, "wikisource") == 0 ||
+      g_strcmp0 (source, "wikibooks") == 0)
+    return _renderer_render_legacy_content(renderer,
```
cosimoc: Missing space before paren (also below)

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+      return g_variant_new_strv (css_files, -1);
+    }
+
+  return g_variant_new_strv (empty_css_files, -1);
+}
+
+static GVariant *
+get_legacy_javascript_files (gboolean use_scroll_manager)
+{
+  GPtrArray *javascript_files = g_ptr_array_new_with_free_func (g_free);
```
cosimoc: I don't think that strings need to be copied in this function

eknrenderer/eknrenderer/eknr-renderer.c
cosimoc: Missing space before paren

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+        return NULL;
+
+      disclaimer = g_strdup_printf(_("This page conatins content from %s, available under a %s license."),
+                                   source, license_link);
+      return g_variant_new_string (disclaimer);
+    }
+  else if (g_strcmp0 (source, "wikihow") == 0)
+    {
+      g_autofree char *wikihow_article_link = format_a_href_link (original_uri,
+                                                                   title,
```
cosimoc: Indentation

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+                                       const char   *license,
+                                       const char   *title,
+                                       GError      **error)
+{
+  if (g_strcmp0 (source, "wikisource") == 0 ||
+      g_strcmp0 (source, "wikibooks") == 0 ||
+      g_strcmp0 (source, "wikipedia") == 0)
+    {
+      g_autofree char *original_link = format_a_href_link (original_uri,
+                                                            source_name,
```
cosimoc: Indentation

js/app/articleHTMLRenderer.js
```diff
@@ -55,13 +47,10 @@ var ArticleHTMLRenderer = new Knowledge.Class({
 
     _init: function (props={}) {
         this.parent(props);
+        this._renderer = new Eknr.Renderer({});
```
cosimoc: No need to pass an `{}`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1377 @@
...
+}
+
+static void
+_renderer_mustache_str_ctx_free (mustache_str_ctx *ctx)
+{
+  g_clear_pointer (&ctx->string, g_free);
+  g_free (ctx);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (mustache_str_ctx, _renderer_mustache_str_ctx_free);
```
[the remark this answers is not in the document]
smspillaz: I don't indeed. Must have been a leftover.

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: I suppose I could make a copy. In our present usage, `input` is basically treated as not mutable and the closure only stays alive for as long as the caller does. But I guess making a copy makes it easier to avoid mistakes...

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: Not really - the old code inserted a boolean with `false`. This actually translates into "skip this substitution entirely" in mustache-land.

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: Good idea!

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: Handy, okay, I can avoid the copy then.

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: I decided to copy `data->input` so it should be freed below now.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+G_DEFINE_TYPE_WITH_PRIVATE (EknrRenderer,
+                            eknr_renderer,
+                            G_TYPE_OBJECT)
+
+enum {
+  PROP_0,
+  NPROPS
+};
+
+static GParamSpec *eknr_renderer_props [NPROPS] = { NULL, };
```
cosimoc: This and the above enum are now unused

eknrenderer/eknrenderer/eknr-renderer.c
cosimoc: Use `g_free()` here too for consistency

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+  RendererMustacheData *data = userdata;
+  return mustache_std_strwrite (api, (void *) &data->output, (char *) buffer, buffer_size);
+}
+
+static char *
+_lookup_in_gvariant_dict (RendererMustacheData *data,
+                          const char           *text,
+                          mustache_api_t       *api)
+{
+  g_autoptr(GVariant) value_v = g_variant_dict_lookup_value (data->variables, text, G_VARIANT_TYPE ("s"));
```
cosimoc: Nitpick: use `G_VARIANT_TYPE_STRING`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+  const char *value = NULL;
```
cosimoc: Unused?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+
+  if (value_v == NULL)
+    {
+      g_autofree char *msg = g_strdup_printf ("No such variable %s", text);
+      (*api->error) (api, data, __LINE__, msg);
+      return NULL;
+    }
+
+  return g_variant_dup_string (value_v, NULL);
```
cosimoc: This function does not need to return a copy of the string, and the caller should be changed accordingly.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+static uintmax_t
+_renderer_sect_from_ht (mustache_api_t           *api,
+                        void                     *userdata,
+                        mustache_token_section_t *token)
+{
+  RendererMustacheData *data = userdata;
+  g_autoptr(GVariant) value = g_variant_dict_lookup_value (data->variables,
+                                                           token->name,
+                                                           NULL);
+  g_autofree const char **strv = NULL;
```
cosimoc: Unused?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+  if (whatsapp == NULL)
+    return NULL;
+
+  markup = g_strdup_printf ("<div id=\"default-share-actions\" style=\"visibility: hidden;\">%s%s%s</div>",
+                            facebook,
+                            twitter,
+                            whatsapp);
+  return g_variant_new_string (markup);
+}
+
```
cosimoc: Remove empty line

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+  whatsapp = get_wrapper_share_button_markup ("whatsapp", error);
+
+  if (whatsapp == NULL)
+    return NULL;
+
+  markup = g_strdup_printf ("<div id=\"default-share-actions\" style=\"visibility: hidden;\">%s%s%s</div>",
+                            facebook,
+                            twitter,
+                            whatsapp);
+  return g_variant_new_string (markup);
```
cosimoc: Use `g_variant_new_take_string (g_steal_pointer (&markup));` to avoid an extra copy.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+  json_builder_end_object (metadata_builder);
+
+  /* Stringify */
+  node = json_builder_get_root (metadata_builder);
+  generator = json_generator_new ();
+
+  json_generator_set_root (generator, node);
+  json_string = json_generator_to_data (generator, NULL);
+
+  return g_variant_new_string (json_string);
```
cosimoc: Use `g_variant_new_take_string()` to avoid another string copy

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+
+      if (strncmp (tag, "Ekn", 3) == 0)
+        continue;
+
+      set = g_hash_table_lookup (set_map, tag);
+
+      if (set == NULL)
+        continue;
+
+      g_ptr_array_add (sets, eknr_boxed_set_info_ref (set));
```
cosimoc: There's no need to ref/unref the sets when adding them to the array; `set_map` will stay alive and won't change during the execution of this function.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+      set = g_hash_table_lookup (set_map, tag);
+
+      if (set == NULL)
+        continue;
+
+      if (!set->featured)
+        continue;
+
+      g_ptr_array_add (parent_featured_sets,
+                       eknr_boxed_set_info_ref (set));
```
cosimoc: As mentioned elsewhere, there is no need to ref/unref these, as `set_map` owns the sets and will always stay alive.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+    }
+
+  json_builder_end_array (builder);
+  node = json_builder_get_root (builder);
+  generator = json_generator_new ();
+
+  json_generator_set_root (generator, node);
+  json_string = json_generator_to_data (generator, NULL);
+
+  return g_variant_new_string (json_string);
```
cosimoc: Use `g_variant_new_take_string()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+               "HTML is not server templated, but no renderer exists for %s",
+               source);
+
+  return NULL;
+}
+
+static GVariant *
+get_wrapper_css_files (const char * const *custom_css_files)
+{
+  GPtrArray *wrapper_css_files = g_ptr_array_new ();
```
cosimoc: Should be `g_autoptr()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+    "prensa-libre.css",
+    NULL
+  };
+
+  if (stripped_body == NULL)
+    return NULL;
+
+  g_variant_dict_init (&vardict, NULL);
+
+  g_variant_dict_insert_value (&vardict, "body-html", g_variant_new_string (stripped_body));
```
cosimoc: `g_variant_new_take_string()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+
+  if (disclaimer == NULL)
+    return NULL;
+
+  g_variant_dict_init (&vardict, NULL);
+
+  g_variant_dict_insert_value (&vardict,
+                               "title",
+                               show_title ? g_variant_new_string (title) : g_variant_new_boolean (FALSE));
+  g_variant_dict_insert_value (&vardict, "body-html", g_variant_new_string (stripped_body));
```
cosimoc: Use `g_variant_new_take_string()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+      return g_variant_new_strv (css_files, -1);
+    }
+
+  return g_variant_new_strv (empty_css_files, -1);
+}
+
+static GVariant *
+get_legacy_javascript_files (gboolean use_scroll_manager)
+{
+  GPtrArray *javascript_files = g_ptr_array_new ();
```
cosimoc: Should be `g_autoptr()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+                                         error);
+
+      if (wikihow_link == NULL)
+        return NULL;
+
+      disclaimer = g_strdup_printf (_("See %s for more details, videos, pictures and attribution. Courtesy of %s, where anyone can easily learn how to do anything."),
+                                   wikihow_article_link,
+                                   wikihow_link);
+
+      return g_variant_new_string (disclaimer);
```
cosimoc: Use `g_variant_new_take_string()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+        return NULL;
+
+      license_link = format_license_link (license, error);
+
+      if (license_link == NULL)
+        return NULL;
+
+      disclaimer = g_strdup_printf(_("This page conatins content from %s, available under a %s license."),
+                                   source, license_link);
+      return g_variant_new_string (disclaimer);
```
cosimoc: Use `g_variant_new_take_string()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1327 @@
...
+ *
+ * Returns: The renderered document on success, %NULL on error.
+ */
+char *
+eknr_renderer_render_mustache_document_from_file (EknrRenderer *renderer,
+                                                  GFile        *file,
+                                                  GVariant     *variables,
+                                                  GError      **error)
+{
+  g_autofree char *path = g_file_get_uri (file);
```
cosimoc: Maybe call it `uri`?

js/app/articleHTMLRenderer.js
```diff
@@ -7,23 +8,13 @@ const GObject = imports.gi.GObject;
...
-    let uri = 'resource:///com/endlessm/knowledge/data/templates/' + template_filename;
-    if (_template_cache[uri] === undefined) {
-        let file = Gio.file_new_for_uri(uri);
-        let [success, string] = file.load_contents(null);
-        let template = string.toString();
-        Mustache.parse(template);
-        _template_cache[uri] = template;
-    }
-    return _template_cache[uri];
+function template_file(template_filename) {
```
cosimoc: This may be unused now?

eknrenderer/eknrenderer/eknr-boxed-set-info.h
```diff
@@ -0,0 +1,32 @@
...
+G_BEGIN_DECLS
+
+typedef struct _EknrBoxedSetInfo {
+  guint    ref_count;
+  gboolean featured;
+  char    *ekn_id;
+  char    *title;
+  GStrv    child_tags;
+  GStrv    tags;
+} EknrBoxedSetInfo;
```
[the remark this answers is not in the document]
smspillaz: I wouldn't want to use the actual content object model here since part of the point of this PR was to break the dependency on that.

That said, would it be better if this was an actual object? I decided to use a boxed type initially because it was less boilerplate to write for something that was basically there for marshalling data between the caller and the renderer.

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: That's what I wanted, thanks :)

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: Ah, I'm not sure if we want to do that. `data->output.string` is actually allocated by `mustache_c` itself using `malloc`.

eknrenderer/eknrenderer/eknr-boxed-set-info.h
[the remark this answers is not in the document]
ptomato: Hmm, it makes no sense from the point of view of the SDK as a platform if this library can't take EkncContentObjectModels, because then we have to marshal between EkncContentObjectModel and EknrBoxedSetInfo in our code. Maybe in that case we're better off duplicating the code...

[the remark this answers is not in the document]
smspillaz: The issue is that I need to use this library as a separately included module in the companion app and I can't have an ekncontent dependency there. It makes more sense to just pass the primitives around, no?

eknrenderer/eknrenderer/eknr-errors.h
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * EknrError:
+ * @EKNR_ERROR_SUBSTITUTION_FAILED: Template substitution failed
+ * @EKNR_ERROR_HTML_ESCAPE_FAILED: HTML escaping failed
+ *
+ * Error codes for the %EKNR_ERROR error domain
+ */
+typedef enum {
+  EKNR_ERROR_SUBSTITUTION_FAILED,
+  EKNR_ERROR_HTML_ESCAPE_FAILED,
```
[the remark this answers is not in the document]
smspillaz: Yeah, I think I had it there for when the regex compilation failed, but I don't think we need a separate error for that. Will remove.

eknrenderer/eknrenderer/eos-knowledge-renderer.h
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2016 Endless Mobile, Inc. */
```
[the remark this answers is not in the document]
smspillaz: I feel old already :crying_cat_face:

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: Side note: Now that we don't have properties, I was thinking that we could make the renderer a pure function, though it is kinda handy to have a cache of compiled templates around...

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: It does not indeed - thanks for spotting, saves a copy

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: All three of them were in fact, removed.

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: Handy, thanks!

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: True that - thanks!

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: It should be indeed! Thanks for spotting!

(I am not sure how my leak-testing earlier did not catch this ... yikes).

js/app/articleHTMLRenderer.js
smspillaz: It is! thanks!

eknrenderer/eknrenderer/eknr-boxed-set-info.h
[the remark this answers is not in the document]
ptomato: Well, I still have to review the whole thing to get the full context, but if we have to resort to acrobatics to avoid duplicating the code... maybe it would just be better to duplicate the code, and share only the Mustache templates or something like that?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
+/* Copyright 2018 Endless Mobile, Inc. */
+
+#include <string.h>
+#include <stdlib.h>
+
+#include <endless/endless.h>
```
cosimoc: This is the old eos-sdk. Why do you need it?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+#include <glib/gi18n.h>
+#include <json-glib/json-glib.h>
+#include <mustache.h>
+#include <libintl.h>
```
cosimoc: Alphabetical order. But not sure this include is actually needed.

eknrenderer/eknrenderer/eknr-renderer.c
cosimoc: I see, that makes sense.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+
+static mustache_api_t _renderer_mustache_data_vfuncs = {
+  .read = &_renderer_read_from_closure,
+  .write = &_renderer_write_to_closure,
+  .varget = &_renderer_var_from_ht,
+  .sectget = &_renderer_sect_from_ht,
+  .error = &_renderer_set_error
+};
+
+char *
```
cosimoc: This should probably be `static`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+{
+  RendererMustacheData *data = userdata;
+  g_autofree char *value = NULL;
+
+  /* First, if we're in a section in the value is ".", then we
+   * need to replace it with the section name. */
+  if (data->section_variable != NULL && token->text[0] == '.')
+    value = g_strdup (data->section_variable);
+  else
+    value = g_strdup (_lookup_in_gvariant_dict (data, token->text, api));
```
cosimoc: There should be no need to duplicate these strings here

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+}
+
+static uintmax_t
+_renderer_section_from_ht_variant (GVariant                 *variant,
+                                   mustache_api_t           *api,
+                                   void                     *userdata,
+                                   mustache_token_section_t *token)
+{
+  g_autofree char *msg = NULL;
+  const char *vt = g_variant_get_type_string (variant);
```
cosimoc: Use `g_variant_is_of_type()` in this function.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+  g_autofree char *contents = NULL;
+  g_autoptr(RendererMustacheData) data = NULL;
+
+  if (tmpl != NULL)
+    return _renderer_render_mustache_document_internal (tmpl, variables, error);
+
+  if (!g_file_load_contents (file, NULL, &contents, NULL, NULL, error))
+    return NULL;
+
+  data = renderer_mustache_data_new (NULL,
```
cosimoc: Any reason not to pass the variables here? If you did, you could reuse `data` when rendering as well?

cosimoc: Or even, is it common that we call back into this many times during the lifecycle of an EknrRenderer? Why do we not cache things in the non-file case?

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+                                       const char   *original_uri,
+                                       const char   *license,
+                                       const char   *title,
+                                       GError      **error)
+{
+  if (g_strcmp0 (source, "wikisource") == 0 ||
+      g_strcmp0 (source, "wikibooks") == 0 ||
+      g_strcmp0 (source, "wikipedia") == 0)
+    {
+      g_autofree char *original_link = format_a_href_link (original_uri,
```
cosimoc: It does not seem like this is used.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+
+      if (original_link == NULL)
+        return NULL;
+
+      license_link = format_license_link (license, error);
+
+      if (license_link == NULL)
+        return NULL;
+
+      disclaimer = g_strdup_printf(_("This page conatins content from %s, available under a %s license."),
```
cosimoc: Missing space before paren.
Also `conatins`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1311 @@
...
+                                       g_free,
+                                       (GDestroyNotify) free_mustache_template);
+}
+
+EknrRenderer *
+eknr_renderer_new (void)
+{
+  return EKNR_RENDERER (g_object_new (EKNR_TYPE_RENDERER, NULL));
+}
+
```
cosimoc: Remove empty line

eknrenderer/eknrenderer/eknr-boxed-set-info.h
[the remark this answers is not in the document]
cosimoc: @smspillaz My opinion here is that we should try to minimize as much as we can the type proliferation. Perhaps instead of passing the whole table around, we can have API as part of the `render()` call that replaces what we'd need the table for.
Without having given it *too* much thought, for instance, we could have a "tag determination" function that replaces the hash table lookup, and a to-JSON serialization API.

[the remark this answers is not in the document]
smspillaz: > Without having given it too much thought, for instance, we could have a "tag determination" function that replaces the hash table lookup, and a to-JSON serialization API.

I mean, sure, but when we have to go reimplement the exact same logic in the companion app again. I'd prefer to avoid having to do that....

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: `eos_get_license_display_name`

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: It probably is not since we include `gi18n` now

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: :+1:

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: The variables aren't needed when compiling the templates, so there's no need to pass them as its an optional param. In the non-file case we don't have a key to cache by (also, this is a straight port of what the JS code did - the JS code did not cache in the non-file case).

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: It is not ... let me check the original code to see if it is needed.

smspillaz: Okay, good catch - I was using `source` instead of `original_link`!

eknrenderer/eknrenderer/eknr-boxed-set-info.h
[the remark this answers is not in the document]
smspillaz: >  but if we have to resort to acrobatics to avoid duplicating the code... 

I'd really prefer to avoid duplicating the code. The reality is that you'll still need to populate the now-copied Ekncontent structs from something else (though perhaps if ekncontent is a separate library that means that at least from e-k-l we have ABI compatibiltiy across compiled libraries and can pass those structs directly from there ...).

But in any event, if the choice is between duplicating the code and having to copy data around or something else, I'd prefer a "something else", with perhaps the compromise that we define some structs to make passing around what we need a little nicer.

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+_renderer_section_from_ht_variant (GVariant                 *variant,
+                                   mustache_api_t           *api,
+                                   void                     *userdata,
+                                   mustache_token_section_t *token)
+{
+  g_autofree char *msg = NULL;
+
+  if (g_variant_is_of_type (variant, G_VARIANT_TYPE ("as")))
+    return _renderer_strv_sect_from_ht (api, userdata, token, variant);
+  else if (g_variant_is_of_type (variant, G_VARIANT_TYPE ("b")))
```
cosimoc: Nit: `G_VARIANT_TYPE_BOOLEAN`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+    return _renderer_bool_sect_from_ht (api, userdata, token, variant);
+  else if (g_variant_is_of_type (variant, G_VARIANT_TYPE ("s")))
```
cosimoc: Nit: `G_VARIANT_TYPE_STRING`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+
+static uintmax_t
+_renderer_section_from_ht_variant (GVariant                 *variant,
+                                   mustache_api_t           *api,
+                                   void                     *userdata,
+                                   mustache_token_section_t *token)
+{
+  g_autofree char *msg = NULL;
+
+  if (g_variant_is_of_type (variant, G_VARIANT_TYPE ("as")))
```
cosimoc: Nit: `G_VARIANT_TYPE_STRING_ARRAY`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+}
+
+static GFile *
+template_file (const char *filename)
+{
+  g_autofree char *uri = g_strdup_printf ("resource:///com/endlessm/knowledge/data/templates/%s", filename);
+  return g_file_new_for_uri (uri);
+}
+
+char *
```
cosimoc: Should be `static`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+  g_autoptr(GPtrArray) javascript_files = g_ptr_array_new ();
+  const char *content_fixes_js = "content-fixes.js";
+  const char *hide_broken_images_js = "hide-broken-images.js";
+  const char *scroll_manager_js = "scroll-manager.js";
+
+  g_ptr_array_add (javascript_files, (gpointer) content_fixes_js);
+  g_ptr_array_add (javascript_files, (gpointer) hide_broken_images_js);
+
+  if (use_scroll_manager)
+    g_ptr_array_add (javascript_files, (gpointer) scroll_manager_js);
```
cosimoc: I doubt you have to cast to `gpointer` explicitly all the strings in this function; you most likely can simply pass the strings directly to `g_ptr_array_add()`

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+                                        const char   *tmpl_text,
+                                        GVariant     *variables,
+                                        GError      **error)
+{
+  g_autoptr(GVariantDict) variables_dict = g_variant_dict_new (variables);
+  g_autoptr(RendererMustacheData) data = renderer_mustache_data_new (variables_dict,
+                                                                     error,
+                                                                     tmpl_text);
+  g_autoptr(mustache_template_t) tmpl = mustache_compile (&_renderer_mustache_data_vfuncs,
+                                                          (char *) data);
```
cosimoc: Pretty sure that this cast is wrong; either you need to cast to `(void *)` or if that's the case, most likely omit the cast altogether since it wouldn't be necessary.

eknrenderer/eknrenderer/eknr-boxed-set-info.h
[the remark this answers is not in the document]
ptomato: This way just requires duplication of different code, though — the code to marshal the data into the `EknrBoxedSetInfo`.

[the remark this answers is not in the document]
smspillaz: On the plus side of `EknrBoxedSetInfo` - you're only exposing things to the renderer library that it actually needs. However, I can see your point. What would your thoughts be on linking to ekncontent-as-a-library (eg, the companion app links to both as libraries included in its own manifest and not as part of the SDK) and re-using the code that way?

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: whoops, it is the wrong cast! I must have thought that `data` was a string when I was writing this, not a closure! I'll omit the cast.

eknrenderer/eknrenderer/eknr-renderer.c
smspillaz: I think the issue here was that I'd get a warning about an implicit cast away of `const`. But I'll check without it.

smspillaz: Yeah, I get a warning if I do that:

```
usr/include/glib-2.0/glib/garray.h:168:12: note: expected ‘gpointer {aka void *}’ but argument is of type ‘const char *’
 void       g_ptr_array_add                (GPtrArray        *array,
```

eknrenderer/eknrenderer/eknr-renderer.c
```diff
@@ -0,0 +1,1312 @@
...
+  g_autofree char *matched = g_match_info_fetch (match_info, 0);
+  char matched_ent = matched[0];
+  const char *html_entity = html_entity_for_char (matched_ent);
+
+  if (html_entity == NULL)
+    return TRUE;
+
+  g_string_append (result, html_entity);
+
+  /* From the documentation, returning FALSE to continues to the replacement
```
cosimoc: `to continues...to stops`

eknrenderer/eknrenderer/eknr-renderer.c
cosimoc: `g_ptr_array_add (javascript_files, (gpointer) "the-file.js");` should be the warning-less way to do it I'm looking for.

smspillaz: Hum, that does work - TIL.

smspillaz:
Jenkins failure appears to be unrelated, seems like the same thing fails on master:

> # tests/js/app/modules/arrangement/testTiledGrid.js: Suite finished with 0 failures: Arrangement.TiledGrid
ERROR: tests/js/app/modules/arrangement/testTiledGrid.js - exited with status 127 (command not found?)

ptomato:
Splitting out EknContent (https://phabricator.endlessm.com/T21655) is almost done, that should serve as a model for splitting out this library.

I haven't looked at the PR yet. For me the main questions to answer are "does it make sense as an API in the context of the app developer platform" and "are the rendering needs on mobile and desktop likely to diverge"

cosimoc:
>I haven't looked at the PR yet. For me the main questions to answer are "does it make sense as an API in the context of the app developer platform" and "are the rendering needs on mobile and desktop likely to diverge"

My 2c:
- No, it does not make sense in the context of the app developer platform, and it is not really important that it does. This is an internal library for the purpose of supporting the legacy mustache-based renderer across different devices and application developers should never have a need to use this.
- Rendering is most likely going to diverge between mobile and desktop, but I don't think that the needs will diverge. I expect that eventually we're going to use e.g. different CSS stylesheets when rendering on mobile.

smspillaz:
> Splitting out EknContent (https://phabricator.endlessm.com/T21655) is almost done, that should serve as a model for splitting out this library.

In light of this, @cosimoc can you create a new repo for `eos-knowledge-content-renderer` ?

cosimoc:
Done: https://github.com/endlessm/eos-knowledge-content-renderer

smspillaz:
Thanks!

nedrichards:
Some CSS rendering context - we've had a long term desire to use print stylesheets to enable high quality printing output from Endless apps, so the principle that content adapts to the vessel it's presented in is embedded, even if the execution isn't there yet.

js/app/articleHTMLRenderer.js
```diff
@@ -179,134 +69,51 @@ var ArticleHTMLRenderer = new Knowledge.Class({
...
-            case 'wikihow':
-                return this._render_legacy_content(model);
-            case 'prensa-libre':
-                return this._render_prensa_libre_content(model);
-            default:
-                return null;
-            }
-        }
+        return this._renderer.render_content(this._get_html(model),
+                                             model.is_server_templated,
```
ptomato: Coding style nitpick: indent only to 4 spaces and keep arguments on the same line where possible.

js/app/articleHTMLRenderer.js
```diff
@@ -179,134 +69,51 @@ var ArticleHTMLRenderer = new Knowledge.Class({
...
-            'custom-css-files': this._get_app_override_css_files(),
-            'javascript-files': js_files,
-            'copy-button-text': _("Copy"),
-            'share-actions': this._get_share_actions_markup(model),
-            'content': content,
-            'crosslink-data': this._get_crosslink_data(model),
-            'chunk-data': this._get_chunk_data(model),
-            'content-metadata': this._get_metadata(model),
-        });
+        return this._renderer.render_wrapper(content,
```
ptomato: Same comment about indentation

js/app/articleHTMLRenderer.js
```diff
@@ -179,134 +69,51 @@ var ArticleHTMLRenderer = new Knowledge.Class({
...
-
-        function get_button_markup (network) {
-            let file = Gio.file_new_for_uri(`resource:///com/endlessm/knowledge/data/icons/scalable/apps/${network}-symbolic.svg`);
-            let [success, svg] = file.load_contents(null);
-            return `
-            <a class="share-action"
-               onclick="window.webkit.messageHandlers.share_on_${network}.postMessage(0)">
-               ${svg.toString()}
-            </a>`;
+        if (tagMap) {
```
ptomato: I'm thinking move this into the previous commit and add it as a `get_legacy_boxed_set_info()` method of `setMap.js`. That way it can be more easily unit tested.

cosimoc:
@smspillaz I think this is waiting on you agreeing with @ptomato about how the templates are split between e-k-l and eknr and where they live, and implementing whatever we agree upon.

smspillaz:
> @smspillaz I think this is waiting on you agreeing with @ptomato about how the templates are split between e-k-l and eknr and where they live, and implementing whatever we agree upon.

Yup, that's right.

To clarify, I'm not super tied to the idea that the non-legacy code has to live here. But I think if the non-legacy code lives in e-k-l, then we shouldn't have any non-legacy rendering done on the Companion App. To me it seems like those crosslinks and chunks have some sort of importance that we want to preserve, but if they really don't matter on mobile then I'm happy to drop this entire path and put it back in e-k-l.

For now, I'll add another commit that effectively re-implements the non-legacy rendering code here using `eknr_renderer_render_mustache_document_from_file`

smspillaz:
Okay, for a bit of context, since I'm still not fully convinced yet that we should be keeping the non-legacy stuff in e-k-l only, what I have done is added the commits which would put the non-legacy stuff back into e-k-l "on top", so at least we know what it looks like both ways to make an informed decision (also to save work a bit later since making a decision either way just involves squashing a few commits together or dropping them).

js/app/articleHTMLRenderer.js
ptomato: In JS we don't indent to the opening bracket, only 4 spaces. That allows keeping more arguments on the same line without overrunning a sensible character limit.

js/app/articleHTMLRenderer.js
```diff
@@ -179,21 +68,15 @@ var ArticleHTMLRenderer = new Knowledge.Class({
         // "Legacy" content, including most wiki sources and also 2.6 Prensa Libre content,
         // is templated and styled on the client.
 
-        if (model.is_server_templated) {
```
ptomato: I think it might make sense to move this logic around a little bit. Since you still have `_get_html()` in this code anyway, you could remove the `is_server_template` argument from `eknr_renderer_render_content()`, and rename that method to `eknr_renderer_render_content_legacy()`. Then put the test for `is_server_templated` directly in `render()`.

js/app/setMap.js
```diff
@@ -28,3 +28,7 @@ function get_parent_set (subset) {
         return undefined;
     return parentMap.get(subset.ekn_id);
 }
+
+function get_tag_map () {
```
ptomato: Not used anymore.

smspillaz:
> Okay, for a bit of context, since I'm still not fully convinced yet that we should be keeping the non-legacy stuff in e-k-l only, what I have done is added the commits which would put the non-legacy stuff back into e-k-l "on top", so at least we know what it looks like both ways to make an informed decision (also to save work a bit later since making a decision either way just involves squashing a few commits together or dropping them).

@ptomato and I had a discussion about the non-legacy stuff this morning, here's some quick findings:

1. `crosslinks.js` is important. We want that on the mobile side, but the implementation will need to be a little different (more on that later).
2. `chunks.js` is probably irrelevant. We think this was only ever used in prensa-libre, which has been discontinued. That was one big consumer of the `SetMap` and `EknrBoxedSetInfo` stuff which was a pain point.
3. `content-metadata` is probably important, though less so on the mobile side since we don't render any of our own chrome. Unfortunately, this is another big consumer of the `SetMap` and `EknrBoxedSetInfo` stuff. After having a look at the requirements, we decided that instead of having to do an awkward dance to get the sets marshalled into C without DmSet so that we could share hte rest of the logic in populating the metadata, we'll just duplicate the code. I'd like to put a big warning in both EKL and the Companion App that the code is duplicated and that it should be updated in both places if it needs to be updated.
4. The rest of the properties being injected into `article-wrapper.mst` are "copypaste" properties, eg, things that we just extract directly from the metadata and then put directly into the properties dict. So there's not much difference between just doing that versus calling a function that does the same thing from C.
5. We can drop all the copy button/share actions, etc stuff.
6. Since there's a bunch of stuff in `article-wrapper.mst` that we'll end up not needing, it is probably better to fork it and create a `mobile-article-wrapper.mst` that we use in the companion app service.
7. I would prefer not to push any of the responsibility for doing "final rendering" on to the Companion App itself. The Service would be the place that has the final say in that case and the Companion App is just a "dumb client" that receives HTML that "just works".

So, actions:

1. Squash in the commits that "brought back" non-legacy rendering to articleHTMLRenderer in this PR
2. Remove all the non-legacy stuff from eos-knowledge-content-renderer
3. Implement a our own non-legacy content renderer in the Companion App Service. I have a WIP branch that is doing that here: https://github.com/endlessm/eos-knowledge-lib/pull/1602

Hopefully that resolves the confusion and mixed signals that was happening here :)

js/app/articleHTMLRenderer.js
smspillaz: Yeah, I see your point here. Will do.

configure.ac
```diff
@@ -187,6 +187,7 @@ PKG_CHECK_MODULES([EOS_KNOWLEDGE_PRIVATE], [
     epoxy
 ])
 PKG_CHECK_MODULES([GRESOURCE_PLUGIN], [
+    eknr-0,
```
ptomato: Oh, that was all it needed? Excellent

smspillaz: In theory, that's all that's needed. I wasn't able to end-to-end test this as specified on the phabricator comment, but this will get the resource loaded into the process' memory space and there should not be any shadowing issues.

PR #1603, opened 2018-03-11
use feedURI and fix semicolon

princeofgiri:
* for beginner, using two variable feedURI and FEED_URI will make their code seems not working, better just use feedURI
* found semicolon displacement


ptomato:
Thanks! I'm happy to see you contributing :smile:

PR #1604, opened 2018-03-14
ThemeableImage: ignore PixbufLoader exceptions

xjuan:
On close pixbuf loaders try to finish parsing the image and can throw
exceptions if there is any error. But since we are not trying to
actually load the image, we just wanted to ge the image size,
we need to ignore them.

https://phabricator.endlessm.com/T21534


ptomato:
The tests are still failing, although I think that may be because the PR job is building master and not the actual branch :confounded:

ptomato:
I merged this despite the eos-build fix not getting in, so that we can at least fix our builds.

PR #1605, opened 2018-03-15
T17423

tchx84:
https://phabricator.endlessm.com/T17423


js/app/modules/card/default.js
```diff
@@ -98,6 +112,42 @@ var Default = new Module.Class({
             this.set_media_overlay_from_model(this._thumbnail_overlay);
     },
 
+    _setup_justification: function () {
+        if (this.justify === Gtk.Justification.LEFT) {
+            this._title_label.halign = Gtk.Align.START;
```
ptomato: I think you might be able to do this more elegantly like this:
```js
const property_changes = {
    halign: Gtk.Align.START,
    justify: Gtk.Justification.LEFT,
    xalign: 0,
};
Object.assign(this._title_label, property_changes);
Object.assign(this._synopsis_label, property_changes);
```
though I'm not 100% sure it works with GObject properties.

PR #1606, opened 2018-03-16
Next chapter of walkthrough

ptomato:
https://phabricator.endlessm.com/T21578


manuq:
At Cleaning up the blog posts, Cheerio is used but there is no `npm install cheerio` and no require line before this point of the turorial. But the final code has it.

ptomato:
Cheerio is already included from libingester, but since we use it directly we should probably include it explicitly in the dependencies! I'll change that.

docs/framework/tutorial/install.md
```diff
@@ -5,6 +5,40 @@ short-description: Installing the tools needed for the walkthrough
 
 This section will walk through installing the tools needed for this walkthrough on Endless OS.
 
+## Installing Flatpak ##
+
+You will need Flatpak on your system to complete this walkthrough.
+If you're using Endless OS, it is already included.
+For other Linux distributions, check [Flatpak's installation instructions][flatpak-install].
+
+> **NOTE:** It's possible to complete the walkthrough and make an app without Flatpak, but you're on your own — that's outside the scope.
```
cosimoc: I'd replace `that's outside the scope` with `packaging and installing an application using a different file format is outside the scope of this tutorial`.

docs/framework/tutorial/install.md
```diff
@@ -25,3 +59,32 @@ export PATH="$HOME/.local/bin:$PATH"
...
+## Installing the Source Sans Pro fonts ##
+
+We'll be using the Source Sans Pro font in our app.
+You can find the latest release on its [GitHub page][source-sans-pro-releases].
+Download the `.otf` versions of the fonts.
+
+We'll build these fonts into our apps, but in order for Hatch Previewer to find them during development, you should also copy them into your local fonts directory:
+
+```bash
+mkdir -p ~/.fonts
```
cosimoc: I think it's `~/.local/share/fonts` on modern distros (`~/.fonts` is still supported but legacy/deprecated at this point)

docs/framework/tutorial/shard.md
```diff
@@ -0,0 +1,353 @@
...
+
+Running the ingester in the last section should have created a directory with a name like `hatch_cc-blog_20180215_104946` (where the numbers are the date and time that the ingester was run.)
+Start Hatch Previewer, either by clicking its icon on the desktop, or from the command line:
+```bash
+flatpak run com.endlessm.HatchPreviewer
+```
+You'll be shown a dialog box where you select a directory.
+Navigate to the hatch directory and select it.
+
+> **NOTE:** You can also pass the hatch directory on the command line, but you have to give an absolute path due to a limitation in Flatpak.
```
cosimoc: Really? Why is that out of curiosity?

docs/framework/tutorial/shard.md
```diff
@@ -0,0 +1,353 @@
...
+You can see the page as it looks when rendered in the article format, and press the "Flip" button to view the raw HTML.
+You can also see the metadata on the right-hand side.
+
+Browse through the posts, and take note of anything that looks odd.
+You can also compare them to the posts on their original website in your browser, to make sure nothing is missing.
+
+The first thing you might notice is that there are some blank spaces in the article format we used.
+These will display metadata, but haven't been filled in yet.
+There's a space after the dash at the top right, for the date that the post was published.
+This isn't pre-filled because we want to display it in whatever date format the user has selected in their settings, and that's impossible to know at this point.
```
cosimoc: I know that this is a good rationale, but in this case it feels a bit like a poor excuse. We could just as well show it in the date format for the locale Hatch Previewer is running on (but not sure how easy that would be)

docs/framework/tutorial/shard.md
```diff
@@ -0,0 +1,353 @@
...
+There's also a space at the bottom of the post where the tags will go.
+We won't determine what tags will be used in the app until later, so there's nothing to display here yet.
+
+## Customizing the fonts and colors ##
+
+Perhaps the most obvious thing to refine is the use of default fonts and colors in the rendered posts.
+We want to change these; in our case to better reflect the style of the original site.
+
+The way we do this is with the [`set_custom_scss`][libingester-set-custom-scss] method of the `BlogArticle` format.
+Each article format has its own default CSS theme.
```
cosimoc: It feels like here we would need a link to a (possibly not existing) documentation of the themes/variables available for each format.

docs/framework/tutorial/shard.md
```diff
@@ -0,0 +1,353 @@
...
+
+## Downloading linked media ##
+
+As a last tweak to our ingester, it seems that one of the articles we are ingesting has an embedded video: ["Art and the Every Day with Mike Winkelmann (AKA beeple)"][embedded-video].
+(Again, if you are doing this walkthrough later, that article might not be in the batch you ingested.)
+We want to download this video and save it as an asset as well.
+
+> **NOTE:** This works a bit differently from saving images as assets.
+> The video doesn't get put into the hatch directly; instead, a placeholder is created, and the video is downloaded later in the next processing step.
+> This is in order to allow transcoding the video in order to use a free codec or a different resolution.
```
cosimoc: Another reason is to keep the hatch creation time quick.

docs/framework/tutorial/shard.md
ptomato: For file-type command-line arguments, flatpak opens a portal for the file if the app doesn't have filesystem access, and the portal code is limited to absolute paths for some reason.

docs/framework/tutorial/install.md
ptomato: Hatch previewer doesn't find them there. (I think it's probably due to Electron using outdated versions of the stack.)

docs/framework/tutorial/shard.md
ptomato: Yeah, we could inject a JS script into Hatch Previewer similar to the one we inject from eos-knowledge-lib. (I'll open an issue for that.)

ptomato: https://phabricator.endlessm.com/T21646

docs/framework/tutorial/shard.md
ptomato: Indeed, this documentation doesn't exist.

docs/framework/tutorial/install.md
cosimoc: That's quite odd. It does not find them when running as a flatpak, or otherwise?

docs/framework/tutorial/shard.md
cosimoc: Can you please open a ticket for that?

docs/framework/tutorial/install.md
ptomato: Flatpak.

docs/framework/tutorial/shard.md
ptomato: https://phabricator.endlessm.com/T21658

tchx84:
@ptomato something I found while reviewing the basin PR, you need to add some classes to the video elements in your ingester, otherwise the "play" button/overlay won't be displayed.

Take a look at the gallery article format for examples.

ptomato:
There must be something else going on; those classes are added automatically by libingester (https://github.com/endlessm/libingester/blob/master/lib/util.js#L136), but indeed they are not in the final product in hatch-previewer.

sgnn7:
@ptomato / @tchx84 previewer [fakes the video styling](https://github.com/endlessm/hatch-previewer/blob/master/preview_renderer.js#L72-L83) because the play button is styled in the app iirc

ptomato:
But the classes added in the line that I linked, are not present in the HTML that I can see in hatch-previewer with the Flip view.

sgnn7:
@ptomato Oh.. that's rather interesting. Got nothing for you on that one.

PR #1609, opened 2018-03-20
T21528

tchx84:
https://phabricator.endlessm.com/T21528

### DIFF
--- a/js/app/modules/layout/topMenu.js
+++ b/js/app/modules/layout/topMenu.js
@@ -8,6 +8,16 @@ const Gtk = imports.gi.Gtk;
 const Module = imports.app.interfaces.module;
 const SlidingPanelOverlay = imports.app.widgets.slidingPanelOverlay;
 
+const Direction = {
+    UP: 0,
+    DOWN: 1,
+};
+
+const MIN_PIXELS_TO_ACTIVATE = {
+    [Direction.UP]: 20,
+    [Direction.DOWN]: 275,
+};
+
 /**
  * Class: TopMenu
  * A layout template with a top menu bar that slides out over the content
@@ -66,6 +76,10 @@ var TopMenu = new Module.Class({
         this._menu_panel = this.add_panel_widget(top_menu_container, Gtk.PositionType.TOP);
         this._menu_panel.expand = true;
 
+        this._previous_position = 0;
+        this._previous_start = 0;
+        this._previous_direction = Direction.DOWN;
+
         this._open_menu();
         let content = this.create_submodule('content');
         this._grid.add(content);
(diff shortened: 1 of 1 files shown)


js/app/modules/layout/topMenu.js
```diff
@@ -8,6 +8,14 @@ const Gtk = imports.gi.Gtk;
 const Module = imports.app.interfaces.module;
 const SlidingPanelOverlay = imports.app.widgets.slidingPanelOverlay;
 
+const DIRECTION = {
```
ptomato: Coding style nitpick: Use `Direction.DOWN` and `Direction.UP`.

js/app/modules/layout/topMenu.js
```diff
@@ -8,6 +8,14 @@ const Gtk = imports.gi.Gtk;
...
+    UP: 0,
+    DOWN: 1,
+};
+
+const DISTANCE_UP = 20;
```
ptomato: Clarify these with a comment, or give them names that indicate what the distance represents, or both. Maybe you could make something that's indexable by a value of `Direction`? e.g.
```js
const APPEAR_THRESHOLD_PX = {
    [Direction.UP]: 20,
    [Direction.DOWN]: 200,
};
```

js/app/modules/layout/topMenu.js
```diff
@@ -66,6 +74,10 @@ var TopMenu = new Module.Class({
         this._menu_panel = this.add_panel_widget(top_menu_container, Gtk.PositionType.TOP);
         this._menu_panel.expand = true;
 
+        this._previous_value = 0;
```
ptomato: I'd name this `_previous_position` to avoid the generic `value`

js/app/modules/layout/topMenu.js
```diff
@@ -82,7 +91,7 @@ var TopMenu = new Module.Class({
         let content = this.create_submodule('content');
         this._grid.add(content);
 
-        if (content instanceof Gtk.ScrolledWindow) {
+        if (this.hide_on_scroll && content instanceof Gtk.ScrolledWindow) {
```
ptomato: I don't think this check is correct, it's cutting out other logic if `hide_on_scroll` is turned off.

js/app/modules/layout/topMenu.js
```diff
@@ -14,7 +14,7 @@ const DIRECTION = {
 };
 
 const DISTANCE_UP = 20;
-const DISTANCE_DOWN = 200;
+const DISTANCE_DOWN = 275;
```
ptomato: The code says 275, but the commit message 270. Either seems fine, but they should agree.

tchx84: Will check with @katelk28 , but for now I will assume the code is correct.

PR #1611, opened 2018-03-22
T21684

tchx84:
https://phabricator.endlessm.com/T21684

### DIFF
--- a/data/preset/escola.yaml
+++ b/data/preset/escola.yaml
@@ -2,11 +2,18 @@
 vars:
   app-banner:
     shortdef: 'Banner.Dynamic(mode: text, expand: true)'
-  app-brand:
+  app-banner-text:
     type: ContentGroup.StaticText
     properties:
       halign: center
       label: ''
+  app-brand:
+    type: Decoration.ThemeableImage
+    properties:
+      halign: end
+      expand: true
+      valign: center
+      image-uri: resource:///app/assets/watermark
 
   home-sets-arrangement:
     type: Arrangement.Grid
@@ -40,7 +47,7 @@ defines:
     - footer
   slots:
     contents:
-    - shortdef: 'Decoration.ThemeableImage(halign: end, expand: true, valign: center, image-uri: "resource:///app/assets/watermark")'
+    - $app-brand
 
 - &background
   shortdef: 'Layout.Box(orientation: vertical)'
@@ -67,7 +74,7 @@ defines:
               slots:
                 contents:
                 - refvar: $app-banner
-                - refvar: $app-brand
+                - refvar: $app-banner-text
         - shortdef: 'Layout.Box(orientation: vertical)'
           slots:
             contents:
--- a/data/preset/library-list.yaml
+++ b/data/preset/library-list.yaml
@@ -2,6 +2,13 @@
 vars:
   app-banner:
     shortdef: 'Banner.Dynamic(mode: full)'
+  app-brand:
+    type: Decoration.ThemeableImage
+    properties:
+      halign: end
+      valign: end
+      expand: true
+      image-uri: 'resource:///app/assets/watermark'
 
   home-sets-arrangement:
     shortdef: 'Arrangement.Overflow(orientation: horizontal)'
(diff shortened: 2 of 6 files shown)


tchx84:
NOT FOR MERGING YET.

tchx84:
Moving this to Dev, to fix a typo...

data/preset/escola.yaml
```diff
@@ -2,11 +2,13 @@
...
   app-banner:
     shortdef: 'Banner.Dynamic(mode: text, expand: true)'
-  app-brand:
+  app-banner-text:
     type: ContentGroup.StaticText
     properties:
       halign: center
       label: ''
+  app-brand:
+    shortdef: 'Decoration.ThemeableImage(halign: end, expand: true, valign: center, image-uri: "resource:///app/assets/watermark")'
```
ptomato: Prefer the long form here, since this makes for an overly long line.

data/preset/library-list.yaml
```diff
@@ -2,6 +2,13 @@
 vars:
   app-banner:
     shortdef: 'Banner.Dynamic(mode: full)'
+  app-brand:
+    type: Decoration.ThemeableImage
+    properties:
+      halign: end
+      valign: end
+      expand: true
+      image-uri: 'resource:///app/assets/watermark'
```
ptomato: Is there any way to make the align/expand properties of this configurable but not the `image-uri`? Otherwise, we effectively have two ways to customize it: including an asset at `resource:///app/assets/watermark` or not, and overriding this variable or not.

data/preset/escola.yaml
```diff
@@ -2,11 +2,13 @@
 vars:
   app-banner:
     shortdef: 'Banner.Dynamic(mode: text, expand: true)'
-  app-brand:
+  app-banner-text:
     type: ContentGroup.StaticText
     properties:
       halign: center
       label: ''
+  app-brand:
```
ptomato: Are you sure this `app-brand` should be named the same as the other `app-brand`s? This is overlaid in the corner like a watermark, and would be omitted on most uses of the template except for FNDE, and the other ones are a more integral part of the design.

data/preset/library-list.yaml
tchx84: You might want to use something completely different here, like a dynamic logo. I think the fact that we can also customize this default by providing a different watermark is just circumstantial.

data/preset/escola.yaml
tchx84: > would be omitted on most uses of the template except for FNDE

In general, variables are optional to use (if a variable exist, it doesn't mean the user will be forced to use the var). In particular, if no watermark is provided, the behavior of this default (Banner.App) is to hide it. Taking these two things into account, I don't see a problem if this gets ignored by whoever else uses this preset in the future.

Also, the `app-brand` var is (as of this PR) used in `video_list`, `library-list`and `escola` presets. (trying to think as someone using this) If I see this watermark in one preset and can be customized with a `app-brand` var then, I would expect that if I see this same watermark in another preset, then I would also expect a similar var in that preset.

I think consistency is a good reason for keeping the name.

data/preset/library-list.yaml
ptomato: Looking for an asset at `resource:///app/assets/watermark` was intentional, just like looking for one at `resource:///app/assets/logo` for the dynamic logo.

data/preset/escola.yaml
ptomato: But in `video_list` it's not a watermark. The default value there is equal to `*app-banner`, so it's not consistent. (I guess I should have commented on the `video_list` one, instead of this one.)

data/preset/library-list.yaml
ptomato: I'm not sure I agree with the point about putting something completely different here. Sure, you could put a ContentGroup, but I'm not super excited about allowing that.

data/preset/escola.yaml
tchx84: I still believe we shouldn't focus too much on what the current default value is, I am not suggesting we should encourage them to put a ContentGroup, but say someone wants to put a StaticText with "www.mycompany.com", it would be a valid override in all of these 3 places, because conceptually that is what that place is for (regardless of whether one uses a themeableImage or a dynamiclogo as default).

PR #1614, opened 2018-03-23
T21701

tchx84:
https://phabricator.endlessm.com/T21701

### DIFF
--- a/data/css/modules/window/_simple.scss
+++ b/data/css/modules/window/_simple.scss
@@ -7,7 +7,7 @@
         background-color: transparentize(black, 1 - 0.20);
 
         slider {
-            min-width: 8px;
+            min-width: 10px;
             background-color: transparentize(white, 1 - 0.50);
             border-radius: 20px;
         }
--- a/data/css/news.scss
+++ b/data/css/news.scss
@@ -118,3 +118,10 @@ $support-font: 'Fira Sans' !default;
         }
     }
 }
+
+.WindowSimple {
+    scrollbar {
+        // margin-top is for being covered by Layout.TopMenu
+        margin-top: 50px;
+    }
+}
--- a/data/css/gallery.scss
+++ b/data/css/gallery.scss
@@ -54,8 +54,7 @@
 .WindowSimple {
     scrollbar {
         padding: 3px;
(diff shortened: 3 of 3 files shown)


xjuan:
screenshot from 2018-03-27 15-45-56
Gallery looks fine, but news looks broken
should not the background on news also be transparent?

tchx84:
@xjuan yeah, I think it's missing that in the news overries... I will fix it.

data/css/modules/window/_simple.scss
```diff
@@ -4,12 +4,12 @@
 
     scrollbar {
         padding: 3px;
-        background-color: transparentize(black, 1 - 0.20);
+        background-color: #d6d7d8;
```
tchx84: I think this needs more work, since it doesn't look too good on other presets/themes:

screenshot from 2018-03-27 15-27-59
screenshot from 2018-03-27 15-27-38

I am going to try to revert these changes and try to extract the core improvements without affecting other presets/themes.

data/css/news.scss
```diff
@@ -118,3 +118,8 @@ $support-font: 'Fira Sans' !default;
         }
     }
 }
+
+.WindowSimple scrollbar {
+    // margin-top is for being covered by Layout.TopMenu
+    margin-top: 70px;
```
tchx84: I think this wasn't tested against the default news, I will tweak it as part of the rework.

screenshot from 2018-03-27 15-32-35

data/css/news.scss
```diff
@@ -118,3 +118,10 @@ $support-font: 'Fira Sans' !default;
         }
     }
 }
+
+.WindowSimple {
+    scrollbar {
+        // margin-top is for being covered by Layout.TopMenu
+        margin-top: 50px;
```
xjuan: We need to open a new task to handle this properly.
Ideally this should be the responsibility of the container that packs the topbar over the webview

xjuan:
CR pass

data/css/news.scss
ptomato: If I understand correctly what you mean, then I agree! Can you make sure to open this issue?

PR #1615, opened 2018-03-23
Depend on xapian-glib-2.0

ebassi:
We're switching to a new version of Xapian, which means we're going to
be using a new version of the Xapian-GLib bindings.

https://phabricator.endlessm.com/T21284


ekncontent/tests/ekncontent/testQueryObject.js
```diff
@@ -1,4 +1,6 @@
 const Eknc = imports.gi.EosKnowledgeContent;
+
+imports.gi.Xapian.version = '2.0'
```
ptomato: missing semi

PR #1616, opened 2018-03-23
T21717

tchx84:
https://phabricator.endlessm.com/T21717

### DIFF
--- a/js/app/modules/banner/dynamic.js
+++ b/js/app/modules/banner/dynamic.js
@@ -44,6 +44,13 @@ var Dynamic = new Module.Class({
          */
         'mode': GObject.ParamSpec.string('mode', 'mode', 'mode',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT, 'text'),
+         /**
+         * Property: Layout
+         *
+         * The layout of the logo in full mode, it can be auto, horizontal or vertical.
+         */
+        'layout': GObject.ParamSpec.string('layout', 'layout', 'layout',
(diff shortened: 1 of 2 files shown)


js/app/modules/banner/dynamic.js
```diff
@@ -44,6 +44,13 @@ var Dynamic = new Module.Class({
          */
         'mode': GObject.ParamSpec.string('mode', 'mode', 'mode',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT, 'text'),
+         /**
+         * Property: Layout
+         *
+         * The layout of the logo in full mode, it can be auto, horizontal or vertical.
+         */
+        'layout': GObject.ParamSpec.string('layout', 'layout', 'layout',
```
ptomato: Drive-by comment: it would be good to have an enum here; for the YAML file it makes no difference.

tchx84: As we spoke, we don't do this anywhere and seems it can't be done (without creating a C enum I supposed).

js/app/widgets/dynamicLogo.js
```diff
@@ -139,6 +147,17 @@ var DynamicLogo = new Knowledge.Class({
         return this._mode;
     },
 
+    set layout(value) {
+        if (this._layout === value)
+            return;
+        this._layout = value;
+        this.queue_draw();
```
xjuan: This should actually call queue_resize() instead of draw

tchx84: This is a weird widget, it receives an imposed size and it will always make sure that everything fits inside the size. Therefore, changing the layout won't need a resize, because the widget will simply draw in whatever space has available.

xjuan:
CR Pass

PR #1620, opened 2018-03-28
Utils: use proper hand on set_hand_cursor_on_widget()

xjuan:
Use pointer hand instead of grab hand

https://phabricator.endlessm.com/T21802


katelk28:
looks good to me!

PR #1622, opened 2018-03-28
T17144

tchx84:
https://phabricator.endlessm.com/T17144


js/app/modules/filter/currentSet.js
```diff
@@ -51,25 +51,32 @@ var CurrentSet = new Module.Class({
...
+
+        if (this.invert) {
+            if (query.excluded_tags.indexOf(tags[0]) !== -1)
+                return query;
+        } else {
+            if (query.tags_match_any.indexOf(tags[0]) !== -1)
+                return query;
+        }
+
+        let subtags = [];
```
ptomato: `concat()` in a loop is creating a lot of new array objects and discarding them. Better would be `const subtags = SetMap.get_children_sets(this._current_set).map(set => { ... });`

js/app/modules/filter/currentSet.js
```diff
@@ -51,25 +51,32 @@ var CurrentSet = new Module.Class({
...
+        SetMap.get_children_sets(this._current_set).forEach(set => {
+            subtags = subtags.concat(set.child_tags);
+        });
+
         if (this.invert) {
             return Eknc.QueryObject.new_from_object(query, {
-                excluded_tags: Utils.union(query.excluded_tags,
-                    this._current_set.child_tags),
+                tags_match_any: subtags,
```
ptomato: I think this still has to take any existing `excluded_tags` and `tags_match_any` on the query object into account.

tchx84: Not sure what you mean, can you clarify?

ptomato: The current code in this PR overwrites any `excluded_tags` and `tags_match_any` that were already set on the passed-in query object. So it makes this filter incompatible to chain with other filters that modify those properties. That's the reason why `Utils.union()` was originally in this code.

tchx84: ouch, yeah, my very bad.

PR #1631, opened 2018-04-05
T20580

tchx84:
https://phabricator.endlessm.com/T20580

### DIFF
--- a/js/app/interfaces/scrolling.js
+++ b/js/app/interfaces/scrolling.js
@@ -1,0 +1,46 @@
+// Copyright 2018 Endless Mobile, Inc.
+
+const Gtk = imports.gi.Gtk;
+const Lang = imports.lang;
+
+const HistoryStore = imports.app.historyStore;
+const Module = imports.app.interfaces.module;
+
+/**
+ * Interface: Scrolling
+ */
+var Scrolling = new Lang.Interface({
+    Name: 'Scrolling',
+    GTypeName: 'EknScrolling',
+    Requires: [Gtk.ScrolledWindow, Module.Module],
+
+    Slots: {
+        /**
+         * Slot: content
+         * Where to put the content
+         */
+        'content': {},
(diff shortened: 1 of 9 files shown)


js/app/interfaces/scrolling.js
```diff
@@ -0,0 +1,46 @@
...
+
+    Slots: {
+        /**
+         * Slot: content
+         * Where to put the content
+         */
+        'content': {},
+    },
+
+    _interface_init: function () {
```
ptomato: Use the new method syntax in new code. `_interface_init() { ... }`

js/app/interfaces/scrolling.js
```diff
@@ -0,0 +1,46 @@
...
+
+    // return scroll position to the top of the window
+    _return_to_top: function () {
+        let store = HistoryStore.get_default();
+        let current = store.get_current_item();
+        let previous = store.get_direction() === HistoryStore.Direction.FORWARDS ?
+            store.get_previous_item() : store.get_next_item();
+
+        // if switching to or from a lightbox, don't reset
+        if ((current && current.media_model) || (previous && previous.media_model)) {
```
ptomato: I think you want XOR here. This code will reset the scroll if you use the arrows to page to another image while in the lightbox.

Unfortunately there is no logical XOR operator in JS but you can fake it with: `if (!(current && current.media_model) !== !(previous && previous.media_model))`

tchx84: > This code will reset the scroll if you use the arrows to page to another image while in the lightbox.

Switching between images in a lightbox does not interact with the historyStore. It won't modify the historyStore, so it can't be navigated with the page buttons (and can't reset the scroll).

However, the change should be fine, and it would make it more robust for possible changes in the future.

tchx84:
Found one more issue, moving it back to Dev.

PR #1633, opened 2018-04-06
Fix tutorial node tar extraction step

Alphabeater:
Error gives:
```tar (child): --strip-components=1: Cannot open: No such file or directory```


ptomato:
Thank you! This is correct.

PR #1635, opened 2018-04-09
Use Maxwell to embed nav-content in articles

xjuan:
https://phabricator.endlessm.com/T21356


xjuan:
NOTE: I still need to test these changes with PDF content

PR #1636, opened 2018-04-11
Layout.HamburgerBasement improvements

ptomato:
https://phabricator.endlessm.com/T22044

### DIFF
--- a/js/app/moltresEngine.js
+++ b/js/app/moltresEngine.js
@@ -74,16 +74,15 @@ var MoltresEngine = new Lang.Class({
     },
 
     _get_sets: function (query) {
-        let set_models = this._set_models.filter((model) => {
-            if (query.tags_match_any.length === 0)
-                return true;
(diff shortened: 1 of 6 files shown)


js/app/moltresEngine.js
```diff
@@ -74,16 +74,15 @@ var MoltresEngine = new Lang.Class({
...
+        let set_models = this._set_models;
+
+        if (query.tags_match_any.length > 0) {
+            set_models = set_models.filter(model =>
+                query.tags_match_any.some(tag => model.tags.includes(tag)));
+        }
 
         return {
-            models: set_models,
+            models: set_models.slice(query.offset, query.limit + query.offset),
```
tchx84: like I mentioned in slack, I still can't see the load more button even with these changes.

ptomato: To recap from Slack, the changes apply only when displaying dummy sets.

PR #1638, opened 2018-04-12
T21938

tchx84:
http://phabricator.endlessm.com/T21938


tchx84:
@ptomato good catch! Another thing we could do is just setting its visibility to false, so the widget/class will still be there. Any preference?

PR #1642, opened 2018-04-18
T22162

tchx84:
https://phabricator.endlessm.com/T22162

### DIFF
--- a/data/css/course.scss
+++ b/data/css/course.scss
@@ -106,9 +106,9 @@ $display-font: 'Raleway Black' !default;
     &__title {
       font-size: 48px;
       font-family: $display-font;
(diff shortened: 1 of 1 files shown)


katelk28:
Everything looks good, except for a few things we should keep in mind.

**Course**
I don't think we fully tested the Course template, because I'm seeing this background bug when loading `com.endlessm.fnde_mec_literature.pt_BR`:
image
image
The background looked fine when I was testing it with you, but this happened. I feel like we keep running into problems like this with background, depending on the size of the screen and the size of the background image. It's almost too modular... it needs to be a little smarter. Probably out of scope for this issue.

PR #1643, opened 2018-04-19
T22166: Fix failing articleHTMLRenderer tests

smspillaz:
https://phabricator.endlessm.com/T22166


ptomato:
Thanks! Good find. And it was my fault! :disappointed_relieved:

PR #1652, opened 2018-05-08
Workaround for CI test failure

xjuan:
Test fail under xvfb because libGLX is not properly initialized by
the linker and for some unknown reason pre loading libGL makes
GLX to work.

https://phabricator.endlessm.com/T21657


ebassi:
Let's merge this, then :+1:

PR #1656, opened 2018-05-24
T22630

tchx84:
https://phabricator.endlessm.com/T22630


js/app/warehouse.js
```diff
@@ -12,6 +12,8 @@ var Warehouse = new Knowledge.Class({
     _init: function (props={}) {
         this.parent(props);
         this._local_modules = new Map();
+        this._import_path = imports.app.modules;
```
ptomato: This is confusingly named since it's not a path.

tests/test-content/js/layout/placeholder.js
```diff
@@ -0,0 +1,36 @@
...
+const Gtk = imports.gi.Gtk;
+
+const Config = imports.app.config;
+const Module = imports.app.interfaces.module;
+
+/**
+ * Class: Placeholder
+ * This is a modified version of the SDK Placeholder version
+ */
+var Placeholder = new Module.Class({
```
ptomato: I think you could make this inherit from Layout.Placeholder and just always set the "Overridden" label in the constructor.

tests/test-content/js/custom/custom.js
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018 Endless Mobile, Inc.
+
+/* exported Custom */
+
+const Gtk = imports.gi.Gtk;
+
+const Module = imports.app.interfaces.module;
```
ptomato: Now that I think about it, adding the ability to ship JS in the app makes `imports.app` a really confusing name. Maybe we can rename it to `imports.framework` or something?

tests/test-content/test-content.gresource.xml
```diff
@@ -5,6 +5,10 @@
         <file>backgroundHome</file>
         <file>logo</file>
     </gresource>
+    <gresource prefix="/app">
```
ptomato: It would be better to refactor the lines above so that the image and logo are in an `assets/` directory and you can put the whole thing under `prefix="/app"`.

js/app/warehouse.js
```diff
@@ -32,7 +34,7 @@ var Warehouse = new Knowledge.Class({
                     return current_dir[file_name][module_path];
                 }
                 return current_dir[file_name];
-            }, imports.app.modules);
+            }, this._import_path);
```
ptomato: Not related to this PR, but why do we have this reduce loop anyway? It implies that there should be more than one dot possible. Do we want that?

tests/test-content/js/layout/placeholder.js
tchx84: Well, the trick that allows the "override" to happen is simply the fact that app-version file takes precedence  over the SDK-version file, so (as of now) it would be like trying to import itself (which won't work).

js/app/warehouse.js
tchx84: I asked the same, but I don't see how it would be a problem either, note that this also support zero dots too (which I think is fine).

tchx84:
@ptomato ready again! 

One note: because of the renames, we would need to re-run transifex manually, so the pot/po files get updated with the new paths as well, after this land in master I suppose.

tests/test-content/js/custom/custom.js
ptomato: I don't foresee that, because the only code that could possibly have used the old name is inside eos-knowledge-lib.

tests/test-content/js/layout/placeholder.js
ptomato: In that case - it doesn't seem all that useful to override a framework module, if it's forced to be a completely unrelated implementation. Are you sure that we want to encourage that?

tests/test-content/js/layout/placeholder.js
```diff
@@ -0,0 +1,36 @@
...
+    Name: 'Layout.Placeholder',
+    Extends: Gtk.Frame,
+
+    Properties: {
+        'text': GObject.ParamSpec.string('text', 'Text', 'Placeholder text',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            'Overridden'),
+    },
+
+    _init: function (props={}) {
```
ptomato: Use the new method syntax: `_init(props={}) {`

ptomato:
Another comment, maybe it's worth making a separate pull request with the rename, since this will lose a rebase war with any other PR :-)

tests/test-content/js/layout/placeholder.js
tchx84: Like I mentioned before, this "feature" is the side effect of merging both framework's and app's modules into the same importer node 'imports.framework.modules.*'.

The only benefit I see from this "feature" is to implement customized modules without the extra step of using custom YAML (e.g, to reference the new custom module).

But, yeah, this could be a double-edge sword as tool, e.g. in case some developer inadvertently clashes with framework's modules. 

It could be that the risk is higher than the benefit, and since we don't really loose any functionalities, we could just prevent this.

tests/test-content/js/layout/placeholder.js
tchx84: Good catch.

tchx84:
> Another comment, maybe it's worth making a separate pull request with the rename, since this will lose a rebase war with any other PR :-)

Yeah, that is why I moved it to a separate/following commit, I knew this would happen. We would need a temporal truce ;) to land this rename.

docs/framework/concepts/custom_modules.md
```diff
@@ -0,0 +1,55 @@
...
+root:
+  type: Custom.Custom
+```
+
+## A Custom Importer
+
+To access a custom module class, from another custom module, an importer named `imports_custom` is provided.
+
+```javascript
+const Custom = imports_custom.custom.custom;
```
ptomato: This seems like too many customs :stuck_out_tongue: Maybe we should consider a different developer-facing name for "custom modules"? How about "app modules"? "local modules"? Hmm, maybe "custom modules" is best after all...

js/app/warehouse.js
```diff
@@ -12,16 +12,17 @@ var Warehouse = new Knowledge.Class({
     _init: function (props={}) {
         this.parent(props);
         this._local_modules = new Map();
+        this._importer = imports.app.modules;
+        this._custom_importer = imports['.'];
+        this._custom_importer.searchPath = [];
```
ptomato: just do `this._custom_importer.searchPath = ['resource:///app/js'];`

js/app/warehouse.js
```diff
@@ -32,10 +33,29 @@ var Warehouse = new Knowledge.Class({
                     return current_dir[file_name][module_path];
                 }
                 return current_dir[file_name];
-            }, imports.app.modules);
+            }, importer);
         } catch (error if (error.message.startsWith('No JS module'))) {
```
ptomato: optional: While you're touching this code, consider getting rid of the `catch (error if condition)` since I recently learned the `if` clause is nonstandard and will be removed from the JS engine

js/app/warehouse.js
```diff
@@ -12,16 +12,17 @@ var Warehouse = new Knowledge.Class({
...
 
     register_class: function (module_name, module_class) {
         this._local_modules.set(module_name, module_class);
     },
 
-    type_to_class: function (module_name) {
-        if (this._local_modules.has(module_name))
-            return this._local_modules.get(module_name);
-
+    _load_module: function (module_name, importer) {
```
ptomato: As long as you're changing these lines, use the method shorthand syntax `_load_module(module_name, importer) {`

tchx84: Ill change the other methods as well, I prefer that the same file is consistent in these things.

docs/framework/concepts/custom_modules.md
tchx84: Sure.

js/app/warehouse.js
```diff
@@ -32,10 +32,29 @@ var Warehouse = new Knowledge.Class({
                     return current_dir[file_name][module_path];
                 }
                 return current_dir[file_name];
-            }, imports.app.modules);
-        } catch (error if (error.message.startsWith('No JS module'))) {
```
ptomato: I would just move the `if` statement inside the catch clause, because otherwise this will swallow other errors:
```js
} catch (error) {
    if (error.message.startsWith('No JS module'))
        return null;
    throw error;
}
```

ptomato:
retest this please

PR #1657, opened 2018-05-26
Navigation.SearchBox: Rewrite popup and retheme

ptomato:
This removes the default GtkEntry autocomplete popup in favour of a more
customizable GtkPopover, since a lot of the requested visuals weren't
possible with the default popup (including rounded corners.)

Separates the theming of the popup coming from a search box in the title
bar, from the popup coming from the main search box. Adds mixins to the
SCSS with the various size parameters to ease tweaking in the future.

In some cases (focus out, click) the popup doesn't have the "popdown"
animation. This is because when the window starts animating, the popup's
animation is slotted in after the window's animation, which makes it
jump to the side of the window as its search box starts animating away.

https://phabricator.endlessm.com/T22184

### DIFF
--- a/data/css/modules/navigation/_searchBox.scss
+++ b/data/css/modules/navigation/_searchBox.scss
@@ -5,7 +5,11 @@ $navsb-titlebar-font-size: 16px;
                                 $vpadding: 12px,
                                 $icon-outside-padding: 5px,
                                 $icon-inside-padding: 7px) {
-    padding: $vpadding $hpadding;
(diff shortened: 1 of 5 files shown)


tchx84:
This is looking VERY nice!

A quick question, when we show the new popup on the title bar it looks a bit crowded, should we expand the title-bar text-entry a bit more, so that the popup has extra vertical space?

ptomato:
As discussed in the standup this morning, we'll go ahead and merge this. The title bar can be changed in the future if we so decide.

PR #1658, opened 2018-05-26
ContentGroup.MediaLightbox: Drop submodule on nav change

ptomato:
When the lightbox is exited, or any other navigational change occurs, we
should stop any media file playing in the lightbox.

https://phabricator.endlessm.com/T20433


xjuan:
CR Pass

PR #1660, opened 2018-05-31
 PDFView: implement new design

xjuan:
https://phabricator.endlessm.com/T22633


tchx84:
Everything seems to be working fine now, I even tested disabling the TOC, one last question, is it part of the design that there should always be a title ?

if View.Document(show-titles: true), then you see the View title on top.
if View.Document(show-titles: false), then you see the PDF widget title on top.

No matter what, you get a title on top.

If this is undefined in the design, could we do something like, if `show-titles: true`, and its a pdf, then just show the pdf top title? and if `show-titles: false` don't show any title at all?

tchx84:
A couple more things I noticed:

1) The highlight color used when selecting text is hard to read..

screenshot from 2018-06-07 11-25-46

2) How can I actually see the share options? Is there any specific app? I think I am using the same one @katelk28 was using.

xjuan:
1. I do not know which color its using I hope its not the foreground color as its used to implement thje margins using the border property, I will double check
2. Use the inspector to show the button, or create a shard with pdf content with original uri :)

PR #1661, opened 2018-06-04
Remove unused icons and add fallback ones

xjuan:
Add fallback icons for named icons not present in hicolor theme
Make sure we use EndlessOS icon theme (needed for others OS)

https://phabricator.endlessm.com/T22795


js/app/articleHTMLRenderer.js
```diff
@@ -154,12 +154,22 @@ var ArticleHTMLRenderer = new Knowledge.Class({
             return '';
 
         function get_button_markup (network) {
-            let file = Gio.file_new_for_uri(`resource:///com/endlessm/knowledge/data/icons/scalable/apps/${network}-symbolic.svg`);
-            let [success, svg] = file.load_contents(null);
+            let file = Gio.file_new_for_path(`/usr/share/runtime/share/icons/EndlessOS/scalable/apps/${network}-symbolic.svg`);
+            let markup = null;
+
+	    try {
```
ptomato: nitpick: Indentation is not lining up in this line

js/app/articleHTMLRenderer.js
```diff
@@ -154,12 +154,22 @@ var ArticleHTMLRenderer = new Knowledge.Class({
...
+                let [success, svg] = file.load_contents(null);
```
ptomato: nitpick: if you don't use the `success` variable, do `let [, svg] = file.load_contents(null)`

js/app/articleHTMLRenderer.js
```diff
@@ -154,12 +154,22 @@ var ArticleHTMLRenderer = new Knowledge.Class({
...
+                markup = svg.toString();
+            } catch(e) {
+                let file = Gio.file_new_for_uri(`resource:///com/endlessm/knowledge/data/icons/scalable/apps/${network}-symbolic.svg`);
```
ptomato: Don't shadow the outer `file`, just reassign to the variable

xjuan: never liked that block of code, reworked it into a new helper function

js/app/articleHTMLRenderer.js
```diff
@@ -153,13 +153,28 @@ var ArticleHTMLRenderer = new Knowledge.Class({
         if (!model.original_uri || model.original_uri === '')
             return '';
 
+        function get_svg (uri) {
+            let file = Gio.file_new_for_uri(uri);
+
+            try {
+                let [success, svg] = file.load_contents(null);
+                return (success) ? svg.toString() : null;
```
ptomato: I think this still needs a small rework - `success` can never be false in these GIO functions since otherwise an exception is thrown. So it's better to always ignore it. (It's an annoying [bug](https://gitlab.gnome.org/GNOME/gjs/issues/68) in GJS but can't be changed unless we do a big API break.)

Here's what I would do:
```js
try {
    let [, svg] = file.load_contents(null);
    return svg.toString();
} catch (e) {
    return null;
}
```

xjuan: ohh right, in my narrow C api view I did not realized that it will always return true :)

PR #1662, opened 2018-06-04
docs: Add Electron tutorials

ptomato:
These have been sitting around for a while, so add a notice that they
are not currently supported.

https://phabricator.endlessm.com/T22876


tchx84:
I built the Sdk + this PR, and the docs looks fine, but I have an honest question... Can't we move this somewhere else?

This project has nothing to do with eos-knowledge-lib anymore, the only intersection between these two projects where the ekn-content library which is not part of e-k-l anymore, so I don't see how this belong in this project.

ptomato:
Closing in favour of endlessm/eos-knowledge-content-node#2.

PR #1668, opened 2018-06-20
tests: Remove tests moved to eos-knowledge-content-renderer

smspillaz:
https://phabricator.endlessm.com/T23017


smspillaz:
> it seems like more of the remaining tests should still belong in EKNR?

I am not sure what other tests belong there. The tests I left in place are the ones which apply to server templated content too.

There was an opportunity to get rid of `wikisource_model` though, so I removed that.

ptomato:
OK then.

PR #1669, opened 2018-06-21
PDFView: improve thumbnail generation performance

xjuan:
The icon view redraws itself everytime the list store changes which makes
updating thumbnails one by one really slow.

To speed up things we only update once when we have 16 thumbs and when we finished
generating all of them.

https://phabricator.endlessm.com/T22974


tchx84:
hmm, noticed something odd:

0. update Sdk//master and build your branch with flapack.
1. run `flatpak run --runtime=com.endlessm.appsdev.Sdk/x86_64/master com.endlessm.library.pt/x86_64/eos3 -r`
2. search for `quixote`.
3. scroll the PDF content 2 pages, and it will segfault.

Sadly, no backtrace...

xjuan:
Looks like the real problem was that we need to call a lock()

PR #1671, opened 2018-07-09
T23215

tchx84:
https://phabricator.endlessm.com/T23215

### DIFF
--- a/data/css/widgets/_pDFView.scss
+++ b/data/css/widgets/_pDFView.scss
@@ -2,6 +2,11 @@
 
     background-color: $background-light-color;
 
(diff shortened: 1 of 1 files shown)


data/css/widgets/_pDFView.scss
```diff
@@ -2,6 +2,11 @@
 
     background-color: $background-light-color;
 
+    /* Both panel and content */
+    scrollbar.vertical slider {
+        min-height: 50px;
```
manuq: Adwaita has 40px as `min-height`, we should do the same. And also do this for horizontal sliders as well.

PR #1675, opened 2018-07-25
T23256 (shard_stats)

tchx84:
https://phabricator.endlessm.com/T23256


manuq:
Works like a charm. Code looks good too. Remaining question is, where are videos counted, besides "objects".

PR #1679, opened 2018-08-30
T23124

cfoch:
https://phabricator.endlessm.com/T23124


Makefile.am
```diff
@@ -144,6 +144,9 @@ scss_targets = $(patsubst $(srcdir)/%.scss,$(builddir)/%.css,$(scss_toplevels))
 .scss.css:
 	$(scss_verbose)$(MKDIR_P) $(@D) && $(SCSS) -a $< $@
 
+.yaml.pot:
```
ptomato: This rule target would be appropriate if it was a general rule to turn a .yaml file into a .pot file, one-to-one. In this case it's not, it should be `data/preset/po/presets.pot:`

scripts/genpotfile
```diff
@@ -0,0 +1,31 @@
...
+    esac
+done;
+shift $((OPTIND-1))
+
+if [ -z "${d}" ]; then
+    usage
+fi
+
+toplevel_dir=$(dirname $(readlink -f "$0"))/..
+tmp_dir="/tmp/eos-knowledge-lib/pot"
```
ptomato: I think it would be better to create a directory with `mktemp`, and use a subdir of the build directory, which is after all a temporary directory. Maybe `mktemp -pd`?

scripts/genpotfile
```diff
@@ -0,0 +1,31 @@
...
+po_dir=${d}
+mkdir -p $tmp_dir
+mkdir -p $po_dir
+
+for yaml in ${@:1}; do
+    root=$(basename $yaml | cut -d'.' -f1);
+    $toplevel_dir/tools/autobahn $yaml -c -o $tmp_dir/$root.pot;
```
ptomato: You can just name the file `$yaml.pot` and not bother removing the extension, because you are going to delete the temporary pot file anyway.

Makefile.am
```diff
@@ -144,6 +144,9 @@ scss_targets = $(patsubst $(srcdir)/%.scss,$(builddir)/%.css,$(scss_toplevels))
 .scss.css:
 	$(scss_verbose)$(MKDIR_P) $(@D) && $(SCSS) -a $< $@
 
+data/preset/po/presets.pot: $(dist_yamlpreset_DATA)
+	$(top_srcdir)/scripts/genpotfile -d $(dir $@) $^
```
ptomato: Instead of `$(dir $@)` you can use `$(@D)`, it's already defined automatically

PR #1683, opened 2018-09-15
Fix source code that's incompatible with mozjs60

ptomato:
This is the result of scanning all the sources in js/ and tests/ with
moz60tool (https://gitlab.gnome.org/ptomato/moz60tool) and fixing the
issues.

Especially the str.replace('foo', 'bar', 'g') which was actually already
invalid since GNOME 3.26, was actually causing some minor bugs.

### DIFF
--- a/tests/js/framework/modules/layout/testArticleStack.js
+++ b/tests/js/framework/modules/layout/testArticleStack.js
@@ -71,7 +71,7 @@ describe('Layout.ArticleStack', function () {
         });
 
         selection = factory.get_last_created('contents.selection');
-        view = factory.get_last_created('contents.article.0');
+        view = factory.get_last_created('contents.article');
     });
 
     it('transitions in new content when state changes to article page', function () {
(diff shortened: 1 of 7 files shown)


tests/js/framework/modules/layout/testArticleStack.js
```diff
@@ -71,7 +71,7 @@ describe('Layout.ArticleStack', function () {
         });
 
         selection = factory.get_last_created('contents.selection');
-        view = factory.get_last_created('contents.article.0');
+        view = factory.get_last_created('contents.article');
```
manuq: I ran the moz60tool and it does not complain about this.

tests/js/framework/modules/layout/testArticleStack.js
```diff
@@ -95,7 +95,7 @@ describe('Layout.ArticleStack', function () {
         selection.get_models.and.returnValue([different_article_model]);
         selection.emit('models-changed');
         Utils.update_gui();
-        let new_view = factory.get_created('contents.article.1')[0];
+        let new_view = factory.get_created('contents.article')[1];
```
manuq: Same as above, the moz60tool does not report this as an error. I will run the test suit to see if this change is needed to adapt the test to the changes.

manuq: Yes, after running `flapjack test eos-knowledge-lib` I can see that's the case. Looks good to merge.

PR #1685, opened 2018-09-17
Restart API

ptomato:
~~This pull request is stacked on top of #1683 and #1684, so review those first.~~

To test: it can be a bit difficult, since it's not possible to pass file descriptors with any of the usual DBus tools like D-Feet or `gdbus call`. Also GJS does not support passing file descriptors either out of the box (https://gitlab.gnome.org/GNOME/gjs/issues/204). I think maybe the best way to test it is to checkout endlessm/hack-toolbox-app branch `controlpanel` and build and run it:

```sh
./build-flatpak.sh
flatpak remote-add --no-gpg-verify hack-toolbox-local repo
flatpak install --assumeyes repo com.endlessm.HackToolbox
flapjack run com.endlessm.dinosaurs.en  # run with flapjack, assuming you have built this branch
flatpak run com.endlessm.HackToolbox & sleep 1 ; gjs tools/trigger.js com.endlessm.dinosaurs.en /com/endlessm/dinosaurs/en/window/1
# ...hack toolbox window should appear...
gjs tools/trigger.js com.endlessm.dinosaurs.en /com/endlessm/dinosaurs/en/window/1 flip-back
# ...hack toolbox window should disappear and app should restart...
```

~~Note that the `splice_async` call that reads the file descriptor, takes two seconds and I'm not sure why. Maybe that should be fixed before merging this, but I'm going to proceed using this branch for the time being.~~

https://phabricator.endlessm.com/T23773

### DIFF
--- a/js/framework/application.js
+++ b/js/framework/application.js
@@ -357,12 +357,10 @@ var Application = new Knowledge.Class({
     },
 
     RestartAsync([file_fds, resource_fds, shard_fds, options], invocation, fdlist) {
-        void options;  // no options at this time
-
         const fds = fdlist.peek_fds();
 
         _determine_restart_args(file_fds, resource_fds, shard_fds, fds)
-        .then(args => this._restart_app(args))
+        .then(args => this._restart_app(args, options))
         .then(() => {
             invocation.return_value(null);
             this.quit();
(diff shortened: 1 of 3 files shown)


ptomato:
I was thinking today that an alternative approach might be to scrap the `paths` argument, and pass only handles to GResource files, shard files, and a manifest.json.
The overrides.css, and app.json can be included directly in the GResource file and thereby override the app's original GResources. The web css can be included in the GResource file and referenced in the app.json.

This would be less convenient for callers of the API, but the API would become simpler.

ptomato:
I believe I have fixed the problems with the interface now, and this can once more be considered for merging.

manuq:
In the notes to reproduce, line 3 should be: 
```
flatpak install --assumeyes hack-toolbox-local com.endlessm.HackToolbox
```

manuq:
AC pass. The disco effect combined with alien sounds is hilarius. Code-wise looks good to push as well.

PR #1687, opened 2018-09-28
Revert "New translations from Transifex: transifex-update-eos-knowled…

cfoch:
https://phabricator.endlessm.com/T23124


cfoch:
Preset translations have been removed by the Jenkins script. I think it is because there were no translations in the `preset` resource from Transifex and they were pulled to eos-knowledge-lib.

I think that the solution is as simple as reverting this commit, then doing a "tx push" to update Transifex with the translated files. Then

PR #1691, opened 2019-02-22
Add support for enable_scroll_manager and show_title in non legacy do…

dnohales:
…cuments

These options were only supported by the legacy rendering (used in
Wiki based apps). Now we are stopping to use the legacy rendering by
porting the Wikipedia based apps to a libingester based ingestion
pipeline, so we need to support these to options in non-legacy
documents.

https://phabricator.endlessm.com/T24162

### DIFF
--- a/js/framework/modules/view/document.js
+++ b/js/framework/modules/view/document.js
@@ -266,7 +266,7 @@ var Document = new Module.Class({
     _get_webview: function () {
         let webview = this._create_webview();
 
-        webview.renderer.enable_scroll_manager = true;
+        webview.renderer.enable_scroll_manager = this.toc.visible;
         webview.renderer.show_title = !this.show_toc;
         if (this.custom_css)
             webview.renderer.set_custom_css_files([this.custom_css]);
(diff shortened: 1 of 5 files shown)


dnohales:
Hmmm, there are some server issues in the Jenkins build.

dnohales:
jenkins test this please

dnohales:
lol, it worked

js/framework/modules/view/document.js
```diff
@@ -266,7 +266,7 @@ var Document = new Module.Class({
     _get_webview: function () {
         let webview = this._create_webview();
 
-        webview.renderer.enable_scroll_manager = true;
+        webview.renderer.enable_scroll_manager = this.toc.visible;
```
ptomato: Out of curiosity, is this a fix for a pre-existing bug or is it made necessary by this change? It seems like it can improve performance on apps that don't use the table of contents widget, since I think that scroll-manager script makes things slow.

dnohales: Exactly, with this change, current legacy apps which doesn't show the side TOC, will disable the scroll manager. This change was actually necessary anyway besides the fine side effect, since without it, absolutely all apps (legacy or not) will enable it.

dnohales:
Jenkins test this please

dnohales:
Jenkins test this please

dnohales:
Jenkins test this please

PR #1693, opened 2019-03-23
MVP copy mitigation for content apps

dylanmccall:
These commits add new functionality to PDFView to respect the can-export and can-print metadata belonging to some content. When these parameters are set to False, the widget will not provide a UI to save or print its contents.

https://phabricator.endlessm.com/T25706

This pull request is friends with endlessm/libdmodel#10.


ptomato:
Oh, nice, I did not know PDFs included this information themselves.

dnohales:
Jenkins test this please

dylanmccall:
Jenkins test this please

dylanmccall:
Okay, there was some code in here to check model metadata for the can_print and can_export fields. _But_ there's a problem with libmodel where the default for both of those is `false`, while it should probably be `true`. Removed that code for now, so this can be merged as-is, but I definitely need to solve that :)

dnohales:
Can we fix the failing test case?

dylanmccall:
Yeah, working on that now. Had a failing test earlier, now it's a mysterious error instead.

ptomato:
My memory is a bit fuzzy on this point, but I think the idea was that we didn't want to automatically opt-in all existing apps to be allowed to print and export, so the default was false and apps would opt-in by changing their metadata and rebuilding.

dylanmccall:
Oh really! That's good to know. In that case I won't mess with defaults, but I'll make sure libingester has the tools to help with this. Would that change be safe for this SDK release, then, or would it be better to make at another point?

ptomato:
libingester isn't part of the flatpak SDK, it's more of a build tool. So that should be fine.

PR #1695, opened 2019-04-08
Check that GdkWindow exists before using it

dylanmccall:
This branch fixes several small problems with the MediaBin widget which may have contributed to instability we observed after playing videos.

https://phabricator.endlessm.com/T26126

### DIFF
--- a/js/framework/utils.js
+++ b/js/framework/utils.js
@@ -41,11 +41,13 @@ function set_hand_cursor_on_widget(widget) {
 
     let enter_id = widget.connect('enter-notify-event', function (widget) {
         let cursor = Gdk.Cursor.new_from_name(widget.get_display(), "pointer");
-        widget.window.set_cursor(cursor);
+        if (widget.window)
+            widget.window.set_cursor(cursor);
         return Gdk.EVENT_PROPAGATE;
     });
     let leave_id = widget.connect('leave-notify-event', function (widget) {
-        widget.window.set_cursor(null);
+        if (widget.window)
+            widget.window.set_cursor(null);
         return Gdk.EVENT_PROPAGATE;
     });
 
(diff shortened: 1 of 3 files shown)


js/framework/utils.js
```diff
@@ -41,11 +41,13 @@ function set_hand_cursor_on_widget(widget) {
 
     let enter_id = widget.connect('enter-notify-event', function (widget) {
         let cursor = Gdk.Cursor.new_from_name(widget.get_display(), "pointer");
-        widget.window.set_cursor(cursor);
+        if (widget.window)
```
ManMower: These kinds of "NULL checks" make me nervous - why did we get here with widget.window == NULL?  Is it legitimate to be in this path in this state, or is this hiding a bug from somewhere else?

Also, it feels like only this hunk is directly described by the oneline commit log, I think maybe this should be split?

dylanmccall: Oh, yeah, this was totally added in a mad rush of adding null checks around get_window. Not needed here. Thanks for noticing! I'll remove this commit when I squash. (I think there was another more legitimate null check in here originally, but that may have found its way to master at some point).

dylanmccall:
I like those suggestions! This is on my list for today or early next week. Will see what I can figure out. Ping me if I forget :)

dylanmccall:
In summary, *picard-facepalm-dot-jpg* :P I feel like I may have stumbled across a working solution for this somewhere, but definitely ended up committing something that just falls back to gtksink here. Could have sworn I'd been paying attention to those details in ekn_media_bin_gl_check, but, alas.

At any rate, looks like gtkglsink itself is doing some weird stuff to its surface, so, that's a start :/

dylanmccall:
Okay, killed those commits (moved them to T26126-gl-issues), so this PR is only about that first commit now.

PR #1700, opened 2019-09-18
GNOME 3.32 compatibility fixes

dylanmccall:
Blocks https://github.com/endlessm/endless-sdk-flatpak/pull/60

https://phabricator.endlessm.com/T28284


js/framework/application.js
```diff
@@ -406,8 +406,8 @@ var Application = new Knowledge.Class({
         // Fixed in org.gnome.Platform//3.30. In that version, just pass strings
         // to SpawnSync() instead of byte arrays.
         function zeroTerminatedByteArray(string) {
```
ptomato: If you're updating to 3.32 then you should be able to delete this function entirely.

js/framework/widgets/eknWebview.js
```diff
@@ -113,8 +112,9 @@ var EknWebview = new Knowledge.Class({
         .then((model) => {
             if (model instanceof DModel.Article) {
                 let html = this.renderer.render(model);
-                let bytes = ByteArray.fromString(html).toGBytes();
-                let stream = Gio.MemoryInputStream.new_from_bytes(bytes);
+                let stream = Gio.MemoryInputStream.new_from_bytes(
```
ptomato: I think (but I'm not 100% sure) that you should be able to pass a Uint8Array to any function that expects a GBytes. So `ByteArray.fromString(html)` might work here.

dylanmccall: Ah! I was thinking that back and forth between types felt pretty awkward. Will try it. Thanks :)

js/framework/articleHTMLRenderer.js
```diff
@@ -53,8 +54,9 @@ var ArticleHTMLRenderer = new Knowledge.Class({
     _get_html: function (model) {
         let engine = DModel.Engine.get_default();
         let domain = engine.get_domain();
-        let [, bytes] = domain.read_uri(model.id);
-        return bytes.get_data().toString();
+        let [, data_gbytes] = domain.read_uri(model.id);
+        let data_uint8array = ByteArray.fromGBytes(data_gbytes);
+        return ByteArray.toString(data_uint8array);
```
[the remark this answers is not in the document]
dylanmccall: Thanks for checking! That one generates a warning in the newer version of gjs here :)

> Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the time being, the old behavior has been preserved, but please fix your code anyway to explicitly call ByteArray.toString(array).

The idea is that warning gets shown when we call toString on a Uint8Array instance, while in previous versions of gjs we were working with a ByteArray instance.

dylanmccall:
Okay, so most of these failing tests are from running in the wrong runtime, although I did notice some issues from `make check` with this inside the new runtime (endlessm/endless-sdk-flatpak#60). Seems to be a lingering issue with autobahn which I'm not sure about (those tests are … a bit weird?).

I changed this PR to point to the sdk-6 branch rather than master. I guess the cleverest thing is to get the new runtime building, get the CI to test the sdk-6 branch against the new runtime, then merge when the tests pass :)

dylanmccall:
I took care of that test failure with autobahn. Also noticed that [tests/js/framework/modules/window/testSimple.js](https://github.com/endlessm/eos-knowledge-lib/blob/dylanmccall-gnome-3-32/tests/js/framework/modules/window/testSimple.js) errors out following a segfault inside GTK, but it seems to do the same for me under SDK 5. Kind of strange :/

dbnicholson:
I'm not really in the know about knowledge lib development, but is there any reason master is behind sdk-6?

dylanmccall:
That was just so we don't break master builds of the flatpak runtime. Should be able to tidy that up shortly.

PR #1701, opened 2019-10-10
T28066 corenav

dnohales:
The idea here is to allow to open audio and video files by opening links to those, without the complexity of adding those files as children of the original article.

https://phabricator.endlessm.com/T28066

### DIFF
--- a/js/framework/historyStore.js
+++ b/js/framework/historyStore.js
@@ -321,10 +321,14 @@ var HistoryStore = new Lang.Class({
                 });
             } else if (model instanceof DModel.Media) {
                 let old_item = this.get_current_item();
+                let context = old_item.model ? old_item.model.resources : [];
+                if (!context.includes(model.id)) {
+                    context.unshift(model.id);
+                }
                 this.set_current_item_from_props({
                     page_type: old_item.page_type,
                     model: old_item.model,
-                    context: old_item.model ? old_item.model.resources : [],
+                    context: context,
                     media_model: model,
                 }, EntryPoints.ARTICLE_LINK_CLICKED);
             }
--- a/js/framework/historyStore.js
+++ b/js/framework/historyStore.js
@@ -319,7 +319,7 @@ var HistoryStore = new Lang.Class({
                     model: model,
                     context_label: model.title,
                 });
-            } else if (model instanceof DModel.Media) {
+            } else if (model instanceof DModel.Media || model instanceof DModel.Audio) {
                 let old_item = this.get_current_item();
                 let context = old_item.model ? old_item.model.resources : [];
                 if (!context.includes(model.id)) {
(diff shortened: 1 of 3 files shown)


js/framework/historyStore.js
```diff
@@ -319,12 +319,16 @@ var HistoryStore = new Lang.Class({
                     model: model,
                     context_label: model.title,
                 });
-            } else if (model instanceof DModel.Media) {
+            } else if (model instanceof DModel.Media || model instanceof DModel.Audio) {
                 let old_item = this.get_current_item();
+                let context = old_item.model ? old_item.model.resources : null;
+                if (!context || context.length === 0) {
+                    context = [model];
```
dylanmccall: What does this case cover? There's a similar method, *load_dbus_item*, implemented here as well as in *courseHistoryStore*. In *courseHistoryStore* it does this same thing, setting *context* to the selected model (since it doesn't have an old model to work with).

dylanmccall: Relatedly, I suspect you'll want to change that other function to include DModel.Audio:

https://github.com/endlessm/eos-knowledge-lib/blob/5743a58f434e0642efc55309d8d10773dcc9826d/js/framework/courseHistoryStore.js#L91

And that's so strange DModel.Audio isn't a DModel.Media subclass :thinking:

js/framework/historyStore.js
```diff
@@ -319,12 +319,16 @@ var HistoryStore = new Lang.Class({
...
+                }
```
dylanmccall: What does this case cover? It reminds me of *load_dbus_item* in *CourseHistoryStore*, which sets the context the to the selected model as in this case. Is it possible that that covers the case you're thinking of here?

Relatedly, you may want to change that function to include DModel.Audio as well:

https://github.com/endlessm/eos-knowledge-lib/blob/5743a58f434e0642efc55309d8d10773dcc9826d/js/framework/courseHistoryStore.js#L91

dylanmccall:
I suspect it's an error that DModel.Audio isn't a subclass of DModel.Media, seeing as DModel.Video and DModel.Image both are, but the approach here seems just fine for the task at hand :) Created a task for libdmodel here: https://phabricator.endlessm.com/T28202.

dnohales:
> I suspect it's an error that DModel.Audio isn't a subclass of DModel.Media, seeing as DModel.Video and DModel.Image both are, but the approach here seems just fine for the task at hand :) Created a task for libdmodel here: https://phabricator.endlessm.com/T28202.

Yes, that's something really crucial that I forgot to mention since it doesn't make any sense to me. Maybe we should make it that way and remove the special condition in this PR.

js/framework/historyStore.js
dnohales: The problem is, I'm not sure if `CourseHistoryStore` is suitable since this is not necessarily a course since it doesn't use that template, it looks more wrong than using this fallback.

dylanmccall:
Looks good to me! Merging now... :)

PR #1706, opened 2020-02-11
T29258 kolibri html5 zip

dnohales:
This depends on https://github.com/endlessm/libdmodel/pull/14

I made this originally for the `sdk-5` branch because I wasn't able to build my SDK with Flapjack against SDK-6, I have to ask @dylanmccall how to do that...

The point is that the port to SDK-6 was untested, so this code is untested. Gonna put it on Draft mode until I figure out how to test it.

https://phabricator.endlessm.com/T29258

### DIFF
--- a/js/framework/articleHTMLRenderer.js
+++ b/js/framework/articleHTMLRenderer.js
@@ -52,9 +52,17 @@ var ArticleHTMLRenderer = new Knowledge.Class({
     },
 
     _get_html: function (model) {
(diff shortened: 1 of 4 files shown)


dnohales:
So, I tested this change in SDK-6 built by Buildstream and looks like it works just fine.

dylanmccall:
Excellent! It's still a WIP so I won't, but feel free to merge it when you're ready :)

PR #1708, opened 2020-03-04
Fix more ByteArray warnings

dylanmccall:


js/framework/application.js
```diff
@@ -475,12 +475,14 @@ var Application = new Knowledge.Class({
...
+        let [, stdout_bytes, stderr_bytes, status] = GLib.spawn_command_line_sync(SCSS_COMMAND + theme_file.get_path());
         if (status !== 0) {
-            printerr(new Error(stderr.toString()));
+            let stderr = ByteArray.toString(stderr_bytes)
+            printerr(new Error(stderr));
             System.exit(1);
         }
-        return stdout.toString();
+        let stdout = ByteArray.toString(stdout_bytes);
+        return stdout;
```
dnohales: I would just do `return ByteArray.toString(stdout_bytes);`, same in other places

tools/kactus
```diff
@@ -96,11 +97,12 @@ function show_message (text) {
     dialog.destroy();
 }
 
-function read_sream (stream) {
+function read_stream (stream) {
```
dnohales: :scream:

tools/kactus
```diff
@@ -96,11 +97,12 @@ function show_message (text) {
...
     let output = '', buffer;
     do {
         buffer = stream.read_bytes(1024 * 8, null);
-        output += buffer.get_data().toString();
+        output_uint8array = buffer.get_data();
+        output += ByteArray.toString(output_uint8array);
```
dnohales: `output += ByteArray.toString(buffer.get_data());` is more readable.

dylanmccall:
Good points :) I changed those to be more straight-forward now.

PR #1711, opened 2020-03-05
Replace deprecated GdkScreen get_width in PDFView

dylanmccall:
Instead of gdk_screen_get_width, we should use gdk_monitor_get_geometry.

https://phabricator.endlessm.com/T28284

### DIFF
--- a/js/framework/widgets/PDFView.js
+++ b/js/framework/widgets/PDFView.js
@@ -391,9 +391,12 @@ var PDFView = new Knowledge.Class({
     },
 
     vfunc_get_preferred_width: function () {
-        let width = this.get_screen().get_width();
         let [minimal, ] = this.parent();
-        return [minimal, width];
+
+        let gdk_window = this.get_window();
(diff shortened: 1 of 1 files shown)


js/framework/widgets/PDFView.js
```diff
@@ -391,9 +391,16 @@ var PDFView = new Knowledge.Class({
...
 
     vfunc_get_preferred_width: function () {
-        let width = this.get_screen().get_width();
         let [minimal, ] = this.parent();
-        return [minimal, width];
+        let natural = null;
+
+        let gdk_window = this.get_window();
+        let gdk_display = gdk_window.get_display();
+        let gdk_monitor = gdk_display.get_monitor_at_window(gdk_window);
```
dnohales: You can remove `gdk_display` and the code is still readable.

dylanmccall: Okay, done :) I realized that gdk_monitor should always be set, as well, so I removed the condition there.

PR #1717, opened 2020-06-24
Move genpotfile to scripts

dylanmccall:
This file was moved to `tools` in an earlier commit. However, [scripts in
eos-build](https://github.com/endlessm/eos-build/blob/master/ci-integration/jenkins-sdk-element-transifex-update#L83) expect it to be located in the `scripts` directory.


dylanmccall:
Good question :) The reason I changed it here is the `transifex-update-{repo}` jobs also run for eos-sdk and eos-knowledge-content-renderer, and it assumes genpotfile is in scripts. So we'd need to change `jenkins-sdk-element-transifex-update` to look for genpotfile in multiple places. But I think there's some logic in these directories being separate, really: scripts is stuff for the build environment, whereas tools is stuff we include in the SDK. (It's just the naming is a bit off since "scripts" and "tools" sound like the same thing, and "scripts" should really be ".jenkins" or something).

dnohales:
@dylanmccall That's not the case for the repos you mention, there's no `scripts` folder, `genpotfile` is a specific case for this repo, see [this line](https://github.com/endlessm/eos-build/blob/master/ci-integration/jenkins-sdk-element-transifex-update#L76).

I think we better change this in `eos-build`.

dylanmccall:
Oh :facepalm: Got it! Thanks :) Over to eos-build, then.

dylanmccall:
I committed the inverse solution over here: https://github.com/endlessm/eos-build/commit/ca0d8cb9c6cb82b805de20cf121b84d328481547

PR #1720, opened 2020-07-20
Add release notes for SDK version 6.0

dnohales:
I skipped the mentions on EGL-GLES improvements as Dylan suggested (not that I agree though, the fixes are there after all, right?)

https://phabricator.endlessm.com/T30539

### DIFF
--- a/releases/sdk6.md
+++ b/releases/sdk6.md
@@ -6,17 +6,41 @@ permalink: /releases/6/
 
 ## Release schedule ##
 
-| Date            | Milestone
-|:----------------|:---------
-| Thu Oct 4 2018  | Feature freeze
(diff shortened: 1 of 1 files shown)


dylanmccall:
To clarify what I was thinking about the EGL-GLES stuff, most of the stuff there was unbreaking things that were already working in (or recently fixed in) SDK 5, or moving fixes upstream (with no real effect downstream). Come to think of it, one related thing worth noting (and I think downstream-specific at the moment):

- Fixed a problem where certain programs would never search for libraries from Flatpak extensions, such as the GL host extension.

dylanmccall:
Thank you! I'm merging this now...

